PR #16, opened 2023-05-02
Create database indices on recently added fields

jgiannuzzi:
I forgot to ensure the new `runs.row_num` and `metrics.iter` fields were indexed. This update should increase query speed in a few operations.

### DIFF
--- a/pkg/database/database.go
+++ b/pkg/database/database.go
@@ -149,7 +149,7 @@ func ConnectDB(dsn string, slowThreshold time.Duration, poolMax int, reset bool,
 		tx.First(&schemaVersion)
 	}
 
-	if alembicVersion.Version != "97727af70f4d" || schemaVersion.Version != "8073e7e037e5" {
+	if alembicVersion.Version != "97727af70f4d" || schemaVersion.Version != "ed364de02645" {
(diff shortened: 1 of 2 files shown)


pkg/database/database.go
```diff
@@ -311,6 +311,23 @@ func ConnectDB(dsn string, slowThreshold time.Duration, poolMax int, reset bool,
...
+				log.Info("Migrating database to fasttrack schema ed364de02645")
+				if err := DB.Transaction(func(tx *gorm.DB) error {
+					if err := tx.Migrator().CreateIndex(&Run{}, "RowNum"); err != nil {
+						return err
+					}
+					if err := tx.Migrator().CreateIndex(&Metric{}, "Iter"); err != nil {
+						return err
+					}
+					return tx.Model(&SchemaVersion{}).
+						Where("1 = 1").
```
dannyfriar, line 324: Why is this needed?

dannyfriar (approved):

pkg/database/database.go
jgiannuzzi: This is a quirky way to tell `gorm` to one row we have in that table. See the [docs](https://gorm.io/docs/update.html#Block-Global-Updates) for more details.

PR #17, opened 2023-05-03
Refactor MLFlow API implementation

DSuhinin:
- split some of the requests/response objects to be independent.
- a bit improve error but not fuly finished.
- add error helper functions for different type of errors.
- re group import to use single formatting pattern.
- start adding comments.
- extend error object with more information to keep original error and properly log it in future.


go.mod
```diff
@@ -1,30 +1,26 @@
...
-go 1.18
-
-replace (
-	github.com/mattn/go-sqlite3 v1.14.16 => github.com/jgiannuzzi/go-sqlite3 v1.14.17-0.20230327164124-765f25ea5431
-	gorm.io/driver/sqlite v1.4.3 => github.com/jgiannuzzi/gorm-sqlite v1.4.4-0.20221215225833-42389ad31305
-)
+go 1.20
 
 require (
+	github.com/G-Research/fasttrack v0.1.2
```
jgiannuzzi, line 6: This doesn't look right 🤔

go.mod
```diff
@@ -1,30 +1,26 @@
 module github.com/G-Research/fasttrack
 
-go 1.18
-
-replace (
-	github.com/mattn/go-sqlite3 v1.14.16 => github.com/jgiannuzzi/go-sqlite3 v1.14.17-0.20230327164124-765f25ea5431
-	gorm.io/driver/sqlite v1.4.3 => github.com/jgiannuzzi/gorm-sqlite v1.4.4-0.20221215225833-42389ad31305
-)
```
jgiannuzzi, line 8: Those are important in order to have SQLite3MultipleCiphers support!

go.mod
jgiannuzzi, line 1: Is any dependency update necessary for this particular PR?
If not, I'd like to split them into another PR — we also probably want to enable Dependabot on the repo so it can manage that for us.

jgiannuzzi (commented):
Thanks for the PR @DSuhinin!

I have a first series of comment below about `go.mod`.

go.mod
DSuhinin: emmm, It seems like my IDE just did it. let me revert.

go.mod
DSuhinin: ook

jgiannuzzi (commented):
I believe `pkg/api/mlflow/api.go` should be removed in this PR as well?

DSuhinin:
> I believe `pkg/api/mlflow/api.go` should be removed in this PR as well?

yep, yep. that's already done.

pkg/api/mlflow/api/error.go
```diff
@@ -0,0 +1,90 @@
...
+func NewInternalServerError(msg string, args ...any) *ErrorResponse {
+	return &ErrorResponse{
+		Message:    fmt.Sprintf(msg, args...),
+		ErrorCode:  ErrorCodeInternalError,
+		StatusCode: http.StatusInternalServerError,
+	}
+}
+
+// NewCodeInvalidParameterValueError creates new Response object with ErrorCodeInternalError.
+func NewCodeInvalidParameterValueError(msg string, args ...any) *ErrorResponse {
```
jgiannuzzi, line 57: This should be `NewInvalidParameterValueError`

pkg/api/mlflow/api/error.go
```diff
@@ -0,0 +1,90 @@
...
+	return &ErrorResponse{
+		Message:    fmt.Sprintf(msg, args...),
+		ErrorCode:  ErrorCodeInvalidParameterValue,
+		StatusCode: http.StatusBadRequest,
+	}
+}
+
+// NewResourceNoExistsError creates new Response object with ErrorCodeResourceDoesNotExist.
+func NewResourceNoExistsError(msg string, args ...any) *ErrorResponse {
```
jgiannuzzi, line 66: This should be `NewResourceDoesNotExistError`

pkg/api/mlflow/api/error.go
```diff
@@ -0,0 +1,90 @@
...
+	return &ErrorResponse{
+		Message:    fmt.Sprintf(msg, args...),
+		ErrorCode:  ErrorCodeResourceDoesNotExist,
+		StatusCode: http.StatusBadRequest,
+	}
+}
+
+// NewResourceAlreadyExistError creates new Response object with ErrorCodeResourceAlreadyExists.
+func NewResourceAlreadyExistError(msg string, args ...any) *ErrorResponse {
```
jgiannuzzi, line 75: This should be `NewResourceAlreadyExistsError`

pkg/api/mlflow/api/error.go
```diff
@@ -0,0 +1,90 @@
...
+func NewBadRequestError(msg string, args ...any) *ErrorResponse {
+	return &ErrorResponse{
+		Message:    fmt.Sprintf(msg, args...),
+		ErrorCode:  ErrorCodeBadRequest,
+		StatusCode: http.StatusBadRequest,
+	}
+}
+
+// NewInternalServerError creates new Response object with ErrorCodeInternalError.
+func NewInternalServerError(msg string, args ...any) *ErrorResponse {
```
jgiannuzzi, line 48: This should be `NewInternalError`

pkg/api/mlflow/api/request/expiriment.go
jgiannuzzi, line 1: File should be named `experiment.go`

pkg/api/mlflow/api/request/expiriment.go
```diff
@@ -0,0 +1,57 @@
...
+	Key   string `json:"key"`
+	Value string `json:"value"`
+}
+
+// ExperimentPartialRequest is a partial request object for different requests.
+type ExperimentPartialRequest struct {
+	ID               string                        `json:"experiment_id"`
+	Name             string                        `json:"name"`
+	ArtifactLocation string                        `json:"artifact_location"`
+	LifecycleStage   string                        `json:"lifecycle_stage"`
```
jgiannuzzi, line 14: The type should be `LifecycleStage`

pkg/api/mlflow/api/request/expiriment.go
```diff
@@ -0,0 +1,57 @@
...
+	LastUpdateTime   int64                         `json:"last_update_time"`
+	CreationTime     int64                         `json:"creation_time"`
+	Tags             []ExperimentTagPartialRequest `json:"tags"`
+}
+
+// CreateExperimentRequest is a request object for `POST mlflow/create` endpoint.
```
jgiannuzzi, line 20:
```suggestion
// CreateExperimentRequest is a request object for `POST mlflow/experiments/create` endpoint.
```

Same remarks for all the comments below ⬇️

pkg/api/mlflow/api/error.go
DSuhinin: done.

pkg/api/mlflow/api/error.go
DSuhinin: done.

pkg/api/mlflow/api/error.go
DSuhinin: done.

pkg/api/mlflow/api/request/expiriment.go
```diff
@@ -0,0 +1,57 @@
...
+	Value string `json:"value"`
+}
+
+// SearchExperimentsRequest is a request object for `POST mlflow/list` or `POST mlflow/search` or `GET mlflow/search` endpoint.
+type SearchExperimentsRequest struct {
+	MaxResults int64    `json:"max_results" query:"max_results"`
+	PageToken  string   `json:"page_token"  query:"page_token"`
+	Filter     string   `json:"filter"      query:"filter"`
+	OrderBy    []string `json:"order_by"    query:"order_by"`
+	ViewType   string   `json:"view_type"   query:"view_type"`
```
jgiannuzzi, line 56: The type should be `ViewType`

pkg/api/mlflow/api/request/expiriment.go
DSuhinin: there is no reason to have a type here. It is just a string. So you can probably put any string and it will be valid. We have to check that during the validation process to guarantee, that  `LifecycleStage` has correct value.

pkg/api/mlflow/api/request/log.go
```diff
@@ -0,0 +1,47 @@
...
+
+// MetricPartialRequest is a partial request object for different requests.
+type MetricPartialRequest struct {
+	Key       string `json:"key"`
+	Value     any    `json:"value"`
+	Timestamp int64  `json:"timestamp"`
+	Step      int64  `json:"step"`
+}
+
+// LogParamRequest is a request object for `POST mlflow/log-parameter` endpoint.
```
jgiannuzzi, line 23:
```suggestion
// LogParamRequest is a request object for `POST mlflow/runs/log-parameter` endpoint.
```

Same remark for all the other comments below ⬇️

jgiannuzzi (commented):
General question: you have removed a lot of struct embedding that was used to avoid repeating lines like these
```go
	Key   string `json:"key"`
	Value string `json:"value"`
```

Could you briefly explain why you think this is better? (I'm only wondering — I don't think I have developed a strong opinion about it)

pkg/api/mlflow/api/request/log.go
DSuhinin: done.

pkg/api/mlflow/api/request/expiriment.go
DSuhinin: done.

pkg/api/mlflow/api/request/run.go
```diff
@@ -0,0 +1,102 @@
...
+// CreateRunRequest is a request object for `POST mlflow/create` endpoint.
+type CreateRunRequest struct {
+	ExperimentID string                 `json:"experiment_id"`
+	UserID       string                 `json:"user_id"`
+	Name         string                 `json:"run_name"`
+	StartTime    int64                  `json:"start_time"`
+	Tags         []RunTagPartialRequest `json:"tags"`
+}
+
+// UpdateRunRequest is a request object for `POST mlflow/update` endpoint.
```
jgiannuzzi, line 59:
```suggestion
// UpdateRunRequest is a request object for `POST mlflow/runs/update` endpoint.
```

Same remark for the other comments below

pkg/api/mlflow/api/response/expiriment.go
jgiannuzzi, line 1: This file should be named `experiment.go`

pkg/api/mlflow/api/request/expiriment.go
jgiannuzzi: Understood!

DSuhinin:
> General question: you have removed a lot of struct embedding that was used to avoid repeating lines like these
> 
> ```go
> 	Key   string `json:"key"`
> 	Value string `json:"value"`
> ```
> 
> Could you briefly explain why you think this is better? (I'm only wondering — I don't think I have developed a strong opinion about it)

Sure! As you can see I've splitted  everything into `resquest` and `response` packages. Also Ive splitted everything to be in each separate file per on entity object. So I removed a lot of embedded structs because of:
- it is much easier in future to work with such structs everywhere: code, integration tests, etc.
- even it is much easier to move them, because they are fully independent.
- usually, request and response objects has to be just independent even if the have similar parts.
- since they are splitted right now they has to be fully independed.
- if response or request object will change, it will be changed in own scope.
- there is no advantage to have so many structs and sub structs to just save some lines.

pkg/api/mlflow/api/response/expiriment.go
DSuhinin: done.

pkg/api/mlflow/api/request/run.go
```diff
@@ -0,0 +1,102 @@
...
+	Name           string `json:"run_name"`
+	ExperimentID   string `json:"experiment_id"`
+	UserID         string `json:"user_id,omitempty"`
+	Status         string `json:"status"`
+	StartTime      int64  `json:"start_time"`
+	EndTime        int64  `json:"end_time,omitempty"`
+	ArtifactURI    string `json:"artifact_uri,omitempty"`
+	LifecycleStage string `json:"lifecycle_stage"`
+	ID             string `json:"run_id"`
+}
```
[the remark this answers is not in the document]
DSuhinin: yep, this stayed like a from one of my rename/change/move integrations. done.

pkg/api/mlflow/runs.go
```diff
@@ -866,21 +874,21 @@ func logParams(id string, params []RunParam) error {
 		}
 	}
 
-	if tx := database.DB.CreateInBatches(&dbParams, 100); tx.Error != nil {
-		return NewError(ErrorCodeInternalError, "Unable to insert params for run '%s': %s", id, tx.Error)
+	if tx := database.DB.Create(&dbParams); tx.Error != nil {
```
jgiannuzzi, line 877: Same remark as above

pkg/api/mlflow/runs.go
```diff
@@ -807,15 +815,15 @@ func logMetrics(id string, metrics []Metric) error {
 		}
 	}
 
-	if tx := database.DB.CreateInBatches(&dbMetrics, 100); tx.Error != nil {
-		return NewError(ErrorCodeInternalError, "Unable to insert metrics for run '%s': %s", id, tx.Error)
+	if tx := database.DB.Create(&dbMetrics); tx.Error != nil {
```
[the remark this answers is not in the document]
DSuhinin: done

pkg/api/mlflow/runs.go
```diff
@@ -905,21 +913,20 @@ func setRunTags(id string, tags []RunTag) error {
 
 	tx.Clauses(clause.OnConflict{
 		UpdateAll: true,
-	}).CreateInBatches(&dbTags, 100)
+	}).Create(&dbTags)
```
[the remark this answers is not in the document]
DSuhinin: done.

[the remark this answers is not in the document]
DSuhinin: done.

pkg/api/mlflow/api/request/common.go
```diff
@@ -0,0 +1,30 @@
...
+	LifecycleStageDeleted LifecycleStage = "deleted"
+)
+
+type ViewType string
+
+const (
+	ViewTypeActiveOnly  ViewType = "ACTIVE_ONLY"
+	ViewTypeDeletedOnly ViewType = "DELETED_ONLY"
+	ViewTypeAll         ViewType = "ALL"
+)
```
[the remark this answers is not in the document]
DSuhinin: I keep them a bit for future. Don't worry about unsed code. We will run at the end linter to remove all `dead` code.

pkg/api/mlflow/api/request/run.go
```diff
@@ -0,0 +1,82 @@
...
+	Timestamp int64  `json:"timestamp"`
+	Step      int64  `json:"step"`
+}
+
+// RunDataPartialRequest is a partial request object for different requests.
+type RunDataPartialRequest struct {
+	Metrics []RunMetricPartialRequest `json:"metrics,omitempty"`
+	Params  []RunParamPartialRequest  `json:"params,omitempty"`
+	Tags    []RunTagPartialRequest    `json:"tags,omitempty"`
+}
```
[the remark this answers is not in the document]
DSuhinin: cleaned everything but don't worry. at some point of time we will run linters and Im sure they give a lot of moments to fix/change/delete.

jgiannuzzi:
> * there is no advantage to have so many structs and sub structs to just save some lines.

I generally agree, though I was thinking about the ease of conversion in e.g. [this line](https://github.com/G-Research/fasttrack/blob/v0.1.3/pkg/api/mlflow/runs.go#L592):
```go
	if err := logMetrics(id, []Metric{req.Metric}); err != nil {
```

vs [your change](https://github.com/DSuhinin/fasttrack/blob/bc528a3a9ef0db03d9acd8e8db280fa6806eb117/pkg/api/mlflow/runs.go#L591):
```go
	if err := logMetrics(id, []request.MetricPartialRequest{{
		Key:       req.Key,
		Step:      req.Step,
		Value:     req.Value,
		Timestamp: req.Timestamp,
	}}); err != nil {
```

What are your thoughts on that?

jgiannuzzi (approved):
Thanks @DSuhinin, LGTM 👍

DSuhinin:
> > * there is no advantage to have so many structs and sub structs to just save some lines.
> 
> I generally agree, though I was thinking about the ease of conversion in e.g. [this line](https://github.com/G-Research/fasttrack/blob/v0.1.3/pkg/api/mlflow/runs.go#L592):
> 
> ```go
> 	if err := logMetrics(id, []Metric{req.Metric}); err != nil {
> ```
> 
> vs [your change](https://github.com/DSuhinin/fasttrack/blob/bc528a3a9ef0db03d9acd8e8db280fa6806eb117/pkg/api/mlflow/runs.go#L591):
> 
> ```go
> 	if err := logMetrics(id, []request.MetricPartialRequest{{
> 		Key:       req.Key,
> 		Step:      req.Step,
> 		Value:     req.Value,
> 		Timestamp: req.Timestamp,
> 	}}); err != nil {
> ```
> 
> What are your thoughts on that?

Im ready for this question too =). Yes my solution looks a bit longer and function also looks a bit ugly but:
- we have to keep the things independend, so request object has to be used in most upper levels and do not go to other levels or functions. maybe validator could use it - yes but this is probably it. we have to transform `request` object as soon as possible to `dto` or even actual `db` model and use it.
- function is not final - I know. We are moving from bigger things to smaller. Im sure that this functionality will be perfect when Ill start to work with improvements and movements related to functions and internal logic itself.

So during the refactoring some usual moments may start looking not well unfortunately, but I can't refactor everything at the same time and I even worry to do not break anything. To tell the truth, everything is sooooo coupled so it is a bit hard to split all the thing to own levels but Im sure we can do that.

PR #22, opened 2023-05-12
Geoffw/developer help

suprjinx:


pkg/ui/aim/embed/build.sh
```diff
@@ -1,25 +1,24 @@
...
-# Save current directory
+# current directory and checkout location
 current=$(dirname $(realpath $0))
+repo="${current}/repo"
 
-# Create temporary directory
-repo=$(mktemp -d)
-trap "rm -rf ${repo}" EXIT
+# Checkout source and build if necessary
+if [ ! -d "${repo}" ]; then
```
suprjinx, line 8: This is the meat of this change -- don't rebuild the UI if the repo dir is already cloned. Makes the dev loop faster.

pkg/ui/mlflow/embed/build.sh
```diff
@@ -1,25 +1,24 @@
...
-# Save current directory
+# current directory and checkout location
 current=$(dirname $(realpath $0))
+repo="${current}/repo"
 
-# Create temporary directory
-repo=$(mktemp -d)
-trap "rm -rf ${repo}" EXIT
+# Checkout source and build if necessary
+if [ ! -d "${repo}" ]; then
```
suprjinx, line 8: This is the meat of this change -- don't rebuild the UI if the repo dir is already cloned. Makes the dev loop faster.

Makefile
```diff
@@ -11,6 +14,10 @@ tests-run-integration: ## run integration tests.
 	@echo ">>> Running integration tests."
 	go test -v -p 1 -tags="integration" ./tests/integration/...
 
+PHONY: tests-mlflow
```
dsuhinin, line 17: can you please rename `tests-mlflow` -> `paython-tests-run-integration` and please rename existing targets: `tests-run-unit` and `tests-run-integration` to be `go-tests-run-unit` and `go-tests-run-integration` to be consistent.

also please add a comment like `tests-mlflow: build` -> `tests-mlflow: build ## run python integration tests.`. In that case when you call `make help` you will see useful help message for each target.

suprjinx: I'd suggest "test" should be the first the token of these targets -- so like "test-go-unit", "test-python-integration"? Also as a verb, "test", like "build" or "clean," instead of "tests".

Makefile
```diff
@@ -30,3 +37,24 @@ help: ## display this help
 	@ grep -E '^[a-zA-Z_-]+:.*?## .*$$' $(MAKEFILE_LIST) | awk 'BEGIN {FS = ":.*?## "}; {printf "    \033[36m%-10s\033[0m - %s\n", $$1, $$2}'
 	@ echo
 
+#
+# Build targets
+# 
+PHONY: clean
+clean:
```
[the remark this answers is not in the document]
suprjinx: will do

dsuhinin (approved):
looks good within the context of what I can understand.

PR #23, opened 2023-05-17
improvements and code refactoring.

dsuhinin:
- introduce controller/transport layer.
- introduce dao layer.
- a bit refactor application initialization.
- introduce some sort of DI container.
- other changes and improvements.


suprjinx (commented):
Hey Dmitry -- I work with Jonathan at GR-OSS (look forward to meeting you!). This PR is very clean and well written, nicely done. I have a couple of general comments/questions: 

1) does the scope of this application really require package separation of models from the types used for api request and response? it seems like we could just have api types and scan directly to those from the db, without need for models as intermediary -- but I could easily be wrong. 

2) I'm hoping we can start adding go unit tests to this project and that's easier to do now than later. Even adding some minimal tests to this PR would help. Is it easy to mock the dependent layers?

dsuhinin:
> Hey Dmitry -- I work with Jonathan at GR-OSS (look forward to meeting you!). This PR is very clean and well written, nicely done. I have a couple of general comments/questions:
> 
> 1. does the scope of this application really require package separation of models from the types used for api request and response? it seems like we could just have api types and scan directly to those from the db, without need for models as intermediary -- but I could easily be wrong.
> 2. I'm hoping we can start adding go unit tests to this project and that's easier to do now than later. Even adding some minimal tests to this PR would help. Is it easy to mock the dependent layers?

Hey! Nice to meet you! So:
1. Yep. That's the good practise which you can find in other languages as well. I don't know why in Go it is rarely used. The main main purpose to separate -> because it was already separated by Jonathan  =). Very often parameters of request, response could be changed. Some of them could be just renamed, some of them deleted but not deleted from db. So from my practise it is very, very often request/response object uses as a db model but it gives so much inconvenience. Also in that case our database deeply connected to protocol(in our case HTTP) so if we want to jump to another protocol or use another serialisation mechanism JSON it is almost impossible. Also here we can semi apply SRP principle - each object should has only one responsibility. In case if we want to use one type of everything it makes testing difficult, binds strongly different layers and add more and more problems. Also remember that database could have custom filed types and other additional logic which is hard to keep just in one object. Sooooooo, it will be a lot of problems in future, when application will grow up or when other developers will join. I want to standardise approach, so anyone new in the team has to just follow this rules.
2. Yes, you are totally right and I've already integrated `unit` tests into build pipeline and created some of them for `validation` logic. This is PR is not like a final/latest It is a part of big changes. So yep, Im going to introduce `unit` tests definitely but a bit later.

suprjinx:
I could throw in some elementary unit tests if you want -- I think this is easiest if the various layers are implementing interfaces, so they can be mocked.

dsuhinin:
> I could throw in some elementary unit tests if you want -- I think this is easiest if the various layers are implementing interfaces, so they can be mocked.

let me finish what I planned so then we can discuss.

suprjinx (approved):

PR #28, opened 2023-05-22
Move runs to deleted stage when containing experiment is deleted

suprjinx:
┆Issue is synchronized with this [Jira Task](https://gr-oss.atlassian.net/browse/FML-10) by [Unito](https://www.unito.io)


pkg/api/mlflow/service/experiment/service.go
```diff
@@ -219,6 +221,16 @@ func DeleteExperiment(c *fiber.Ctx) error {
 		return api.NewInternalError("Unable to delete experiment '%d': %s", *exp.ID, tx.Error)
 	}
 
+	if tx := database.DB.Model(&run).Where("experiment_id = ?", exp.ID).Updates(&database.Run{
+		LifecycleStage: database.LifecycleStageDeleted,
+		DeletedTime: sql.NullInt64{
+			Int64: time.Now().UTC().UnixMilli(),
+			Valid: true,
```
dsuhinin, line 228: Probably this changes now a bit outdated, because all the services were changed. Can you please update this branch from main?

dsuhinin (approved):

dsuhinin (approved):

pkg/api/mlflow/dao/repositories/run.go
```diff
@@ -22,6 +22,8 @@ type RunRepositoryProvider interface {
 	Create(ctx context.Context, run *models.Run) error
 	// Delete marks existing models.Run entity as deleted.
 	Delete(ctx context.Context, run *models.Run) error
+	// DeleteByExperimentID marks existing models.Run having the given expirimentID as deleted.
+	DeleteByExperimentID(ctx context.Context, experiementID int32) error
```
dsuhinin, line 26: Im not sure that we need a new method. It seems like we already have `Delete(ctx context.Context, run *models.Run) error` and it does it seems like the same what you did. Can you please take a look?

pkg/api/mlflow/controller/experiment.go
```diff
@@ -93,6 +93,10 @@ func (c Controller) DeleteExperiment(ctx *fiber.Ctx) error {
 		return err
 	}
 
+	if err := c.runService.DeleteRunsByExperiment(ctx.Context(), &req); err != nil {
```
dsuhinin, line 96: just an open question - do we need to use transaction here? @jgiannuzzi what do you think?

dsuhinin (changes requested):

pkg/api/mlflow/dao/repositories/run.go
dsuhinin: hm, what about to have then something like `DeleteByOptions` ?

pkg/api/mlflow/controller/experiment.go
dsuhinin: I don't remember, does sqlite support this?

pkg/api/mlflow/dao/repositories/run.go
suprjinx: i'd support that when we add another case :-)

pkg/api/mlflow/controller/experiment.go
suprjinx: Tx wrapper added

dsuhinin (changes requested):

pkg/api/mlflow/dao/repositories/experiment.go
```diff
@@ -79,8 +79,35 @@ func (r ExperimentRepository) GetByName(ctx context.Context, name string) (*mode
 
 // Update updates existing models.Experiment entity.
 func (r ExperimentRepository) Update(ctx context.Context, experiment *models.Experiment) error {
+	tx := r.db.Begin()
```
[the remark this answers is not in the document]
dsuhinin: yep yep, it will be ok =).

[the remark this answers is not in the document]
suprjinx: okay, see that it does, i'll make this change https://github.com/go-gorm/gorm/blob/7dd702d379bca5e28a2b65278b5a373f326c72ca/finisher_api.go#L616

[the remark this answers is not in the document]
suprjinx: though I don't see this as simpler :-)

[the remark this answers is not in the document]
dsuhinin: yep =). it is just a nice shortcut.

PR #35, opened 2023-05-25
Delete run functionality in aim ui

suprjinx:
┆Issue is synchronized with this [Jira Task](https://gr-oss.atlassian.net/browse/FML-17) by [Unito](https://www.unito.io)


pkg/api/aim/runs.go
```diff
@@ -888,6 +890,31 @@ func SearchAlignedMetrics(c *fiber.Ctx) error {
 	return nil
 }
 
+// DeleteRun will update the run to the deleted lifecycle stage
+func DeleteRun(c *fiber.Ctx) error {
```
dave-gantenbein, line 894: Testme?

suprjinx: there's a big refactor coming that will make this testable

suprjinx:
> LGTM — could you also implement [batch deletion](https://github.com/aimhubio/aim/blob/v3.17.5/aim/web/api/runs/views.py#L245)?

done

pkg/api/aim/runs.go
```diff
@@ -915,6 +916,111 @@ func DeleteRun(c *fiber.Ctx) error {
...
+				err = runRepo.Restore(c.Context(), &run)
+			}
+			if err != nil {
+				return err
+			}
+		}
+		return nil
+	}); err != nil {
+		return err
+	}
```
jgiannuzzi, line 993: I think that `RunRepository` should be the one processing a batch of archive/restore requests. This would avoid making n DB calls for n runs. Also I don't think that this transaction actually works, as `tx` is not passed to the `Archive`/`Restore` functions.

pkg/api/aim/runs.go
```diff
@@ -915,6 +916,111 @@ func DeleteRun(c *fiber.Ctx) error {
...
+			run := models.Run{ID: id}
+			err := runRepo.Delete(c.Context(), &run)
+			if err != nil {
+				return err
+			}
+		}
+		return nil
+	}); err != nil {
+		return err
+	}
```
jgiannuzzi, line 1018: Same remark as above for `ArchiveBatch`

jgiannuzzi (changes requested):
Great stuff, thanks @suprjinx!

I have just a couple of comments (the same one really) about batching calls to the database.

pkg/api/aim/runs.go
suprjinx: ah yes, I see -- so I will add a batch interface for these and then we won't need transactions as it will be a singular

suprjinx: Updated with batch sql statements in respository

jgiannuzzi (approved):

PR #63, opened 2023-06-07
[FML-35] - partially finish implementation.

dsuhinin:
┆Issue is synchronized with this [Jira Task](https://gr-oss.atlassian.net/browse/FML-45) by [Unito](https://www.unito.io)


suprjinx (dismissed):
Looking good!

.github/workflows/ci.yml
```diff
@@ -37,6 +37,12 @@ jobs:
           mkdir pkg/ui/{aim,mlflow}/embed/build
           touch pkg/ui/{aim,mlflow}/embed/build/index.html
 
+      - name: Download mockery
+        run: go install github.com/vektra/mockery/v2@v2.20.0
+
+      - name: Generate mocks
+        run: make mocks-generate; sleep 5;
```
jgiannuzzi, line 44: Why is there a `sleep 5` in there?

dsuhinin: it could be a small lag, so mocks appear with a small delay. 5 seconds enough for all files to appear.

dsuhinin: so just to make sure.

jgiannuzzi: do you mean that when `make mocks-generate` is done, it's actually not fully done? 🤔

dsuhinin: it seems like it works well without sleep but I had such a problem on my local. ok, then lets see.

fabiovincenzi (approved):

suprjinx (approved):
LGTM

pkg/api/mlflow/dao/convertors/log.go
```diff
@@ -37,5 +41,33 @@ func ConvertLogBatchRequestToDBModel(
...
+	for n, metric := range req.Metrics {
+		m := models.Metric{
+			Key:       metric.Key,
+			Timestamp: metric.Timestamp,
+			Step:      metric.Step,
+			RunID:     runID,
+		}
+		if v, ok := metric.Value.(float64); ok {
+			m.Value = v
+		} else if v, ok := metric.Value.(string); ok {
```
dave-gantenbein, line 55: It would be great to have tests asserting these corner cases, and the invalid default case.

PR #68, opened 2023-06-08
allow exact duplicate params to be created without error

suprjinx:
…iling

┆Issue is synchronized with this [Jira Task](https://gr-oss.atlassian.net/browse/FML-50) by [Unito](https://www.unito.io)

### DIFF
--- a/pkg/api/mlflow/dao/repositories/param.go
+++ b/pkg/api/mlflow/dao/repositories/param.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/rotisserie/eris"
 	"gorm.io/gorm"
+	"gorm.io/gorm/clause"
 
 	"github.com/G-Research/fasttrackml/pkg/api/mlflow/dao/models"
 )
@@ -31,7 +32,10 @@ func NewParamRepository(db *gorm.DB) *ParamRepository {
 
 // CreateBatch creates []models.Param entities in batch.
 func (r ParamRepository) CreateBatch(ctx context.Context, batchSize int, params []models.Param) error {
-	if err := r.db.CreateInBatches(params, batchSize).Error; err != nil {
+	if err := r.db.Clauses(clause.OnConflict{
+		Columns:   []clause.Column{{Name: "key"}, {Name: "run_uuid"}},
+		DoUpdates: clause.AssignmentColumns([]string{"value"}),
+	}).CreateInBatches(params, batchSize).Error; err != nil {
 		return eris.Wrap(err, "error creating params in batch")
 	}
 	return nil
--- a/pkg/api/mlflow/dao/repositories/param.go
+++ b/pkg/api/mlflow/dao/repositories/param.go
@@ -2,6 +2,7 @@ package repositories
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/rotisserie/eris"
 	"gorm.io/gorm"
@@ -31,8 +32,44 @@ func NewParamRepository(db *gorm.DB) *ParamRepository {
 
 // CreateBatch creates []models.Param entities in batch.
 func (r ParamRepository) CreateBatch(ctx context.Context, batchSize int, params []models.Param) error {
+	// skip inserting params which exactly match existing rows
+	params, err := r.removeExactMatches(ctx, params)
+	if err != nil {
+		return err
+	}
 	if err := r.db.CreateInBatches(params, batchSize).Error; err != nil {
 		return eris.Wrap(err, "error creating params in batch")
 	}
 	return nil
+}
+
+// removeExactMatches will return a new slice of params which excludes exact matches
+func (r ParamRepository) removeExactMatches(ctx context.Context, params []models.Param) ([]models.Param, error) {
+	var keys, foundKeys []string
+	paramMap := map[string]models.Param{}
+	paramsToReturn := []models.Param{}
+	for _, param := range params {
+		key := fmt.Sprintf("%v-%v-%v", param.RunID, param.Key, param.Value)
+		keys = append(keys, key)
+		paramMap[key] = param
+	}
+
+	tx := r.db.Raw(`
+           select run_uuid || '-' || key || '-' || value
+           from params
+           where run_uuid || '-' || key || '-' || value in ?`, keys).
+		Find(&foundKeys)
+	if tx.Error != nil {
+		return paramsToReturn, eris.Wrap(tx.Error, "problem selecting existing params")
+	}
+
+	for _, foundKey := range foundKeys {
+		delete(paramMap, foundKey)
+	}
+
+	for _, v := range paramMap {
+		paramsToReturn = append(paramsToReturn, v)
+	}
+
+	return paramsToReturn, nil
 }
--- a/pkg/api/mlflow/dao/repositories/param.go
+++ b/pkg/api/mlflow/dao/repositories/param.go
@@ -5,7 +5,6 @@ import (
 
 	"github.com/rotisserie/eris"
 	"gorm.io/gorm"
-	"gorm.io/gorm/clause"
 
 	"github.com/G-Research/fasttrackml/pkg/api/mlflow/dao/models"
 )
@@ -32,10 +31,7 @@ func NewParamRepository(db *gorm.DB) *ParamRepository {
 
 // CreateBatch creates []models.Param entities in batch.
 func (r ParamRepository) CreateBatch(ctx context.Context, batchSize int, params []models.Param) error {
-	if err := r.db.Clauses(clause.OnConflict{
-		Columns:   []clause.Column{{Name: "key"}, {Name: "run_uuid"}},
-		DoUpdates: clause.AssignmentColumns([]string{"value"}),
-	}).CreateInBatches(params, batchSize).Error; err != nil {
+	if err := r.db.CreateInBatches(params, batchSize).Error; err != nil {
 		return eris.Wrap(err, "error creating params in batch")
 	}
 	return nil
--- a/pkg/api/mlflow/dao/repositories/param.go
+++ b/pkg/api/mlflow/dao/repositories/param.go
@@ -32,47 +32,45 @@ func NewParamRepository(db *gorm.DB) *ParamRepository {
 
 // CreateBatch creates []models.Param entities in batch.
 func (r ParamRepository) CreateBatch(ctx context.Context, batchSize int, params []models.Param) error {
-	var err error
-	err = r.db.CreateInBatches(params, batchSize).Error
-	if err != nil {
+	// try to create params in batch; error condition requires special handling
+	// to allow certain duplicates
+	if err := r.db.CreateInBatches(params, batchSize).Error; err != nil {
 		// remove duplicate rows and try again
-		params, err1 := r.removeExactMatches(ctx, params)
-		if err1 != nil {
-			return eris.Wrap(err1, "error removing exact matches after: " + err.Error())
+		dedupedParams, errRemovingMatches := r.removeExactMatches(ctx, params)
+		if errRemovingMatches != nil {
+			return eris.Wrap(errRemovingMatches, "error removing duplicates in batch")
 		}
-		err = r.db.CreateInBatches(params, batchSize).Error
-	}
-
-	if err != nil {
-		return eris.Wrap(err, "error creating params in batch")
+		if err := r.db.CreateInBatches(dedupedParams, batchSize).Error; err != nil {
+			return eris.Wrap(err, "error creating params in batch after removing duplicates")
+		}
 	}
 	return nil
 }
 
 // removeExactMatches will return a new slice of params which excludes exact matches
 func (r ParamRepository) removeExactMatches(ctx context.Context, params []models.Param) ([]models.Param, error) {
-	var keys, foundKeys []string
+	var keys []string
 	paramMap := map[string]models.Param{}
-	paramsToReturn := []models.Param{}
 	for _, param := range params {
 		key := fmt.Sprintf("%v-%v-%v", param.RunID, param.Key, param.Value)
 		keys = append(keys, key)
 		paramMap[key] = param
 	}
 
-	tx := r.db.Raw(`
-           select run_uuid || '-' || key || '-' || value
-           from params
-           where run_uuid || '-' || key || '-' || value in ?`, keys).
-		Find(&foundKeys)
-	if tx.Error != nil {
-		return paramsToReturn, eris.Wrap(tx.Error, "problem selecting existing params")
+	var foundKeys []string
+	if err := r.db.Raw(`
+		select run_uuid || '-' || key || '-' || value
+		from params
+		where run_uuid || '-' || key || '-' || value in ?`, keys).
+		Find(&foundKeys).Error; err != nil {
+		return []models.Param{}, eris.Wrap(err, "problem selecting existing params")
 	}
 
 	for _, foundKey := range foundKeys {
 		delete(paramMap, foundKey)
 	}
 
+	paramsToReturn := []models.Param{}
 	for _, v := range paramMap {
 		paramsToReturn = append(paramsToReturn, v)
 	}
--- a/pkg/api/mlflow/dao/repositories/param.go
+++ b/pkg/api/mlflow/dao/repositories/param.go
@@ -33,8 +33,8 @@ func NewParamRepository(db *gorm.DB) *ParamRepository {
 // CreateBatch creates []models.Param entities in batch.
 func (r ParamRepository) CreateBatch(ctx context.Context, batchSize int, params []models.Param) error {
 	if err := r.db.Clauses(clause.OnConflict{
-		Columns:   []clause.Column{{Name: "key"}, {Name: "run_uuid"}},
-		DoUpdates: clause.AssignmentColumns([]string{"value"}),
+		Columns:   []clause.Column{{Name: "key"}, {Name: "value"}, {Name: "run_uuid"}},
+		DoNothing: true,
 	}).CreateInBatches(params, batchSize).Error; err != nil {
 		return eris.Wrap(err, "error creating params in batch")
 	}
--- a/pkg/api/mlflow/dao/repositories/param.go
+++ b/pkg/api/mlflow/dao/repositories/param.go
@@ -32,12 +32,18 @@ func NewParamRepository(db *gorm.DB) *ParamRepository {
 
 // CreateBatch creates []models.Param entities in batch.
 func (r ParamRepository) CreateBatch(ctx context.Context, batchSize int, params []models.Param) error {
-	// skip inserting params which exactly match existing rows
-	params, err := r.removeExactMatches(ctx, params)
+	var err error
+	err = r.db.CreateInBatches(params, batchSize).Error
 	if err != nil {
-		return err
+		// remove duplicate rows and try again
+		params, err1 := r.removeExactMatches(ctx, params)
+		if err1 != nil {
+			return eris.Wrap(err1, "error removing exact matches after: " + err.Error())
+		}
+		err = r.db.CreateInBatches(params, batchSize).Error
 	}
-	if err := r.db.CreateInBatches(params, batchSize).Error; err != nil {
+
+	if err != nil {
 		return eris.Wrap(err, "error creating params in batch")
 	}
 	return nil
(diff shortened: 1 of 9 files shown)


pkg/api/mlflow/dao/repositories/param.go
```diff
@@ -31,8 +32,50 @@ func NewParamRepository(db *gorm.DB) *ParamRepository {
 
 // CreateBatch creates []models.Param entities in batch.
 func (r ParamRepository) CreateBatch(ctx context.Context, batchSize int, params []models.Param) error {
-	if err := r.db.CreateInBatches(params, batchSize).Error; err != nil {
+	var err error
+	err = r.db.CreateInBatches(params, batchSize).Error
+	if err != nil {
```
dsuhinin, line 37: this block looks not well understandable.

pkg/api/mlflow/dao/repositories/param.go
```diff
@@ -31,8 +32,50 @@ func NewParamRepository(db *gorm.DB) *ParamRepository {
...
+		// remove duplicate rows and try again
+		params, err1 := r.removeExactMatches(ctx, params)
```
dsuhinin, line 39: this is not good to use such names like `err2` or `err2` or any other.

pkg/api/mlflow/dao/repositories/param.go
```diff
@@ -31,8 +32,50 @@ func NewParamRepository(db *gorm.DB) *ParamRepository {
...
+		if err1 != nil {
+			return eris.Wrap(err1, "error removing exact matches after: " + err.Error())
+		}
+		err = r.db.CreateInBatches(params, batchSize).Error
```
dsuhinin, line 43: it is also not a good practise to shadow upper level `err`. when we integrate linter, linter gives and error about that.

pkg/api/mlflow/dao/repositories/param.go
```diff
@@ -31,8 +32,50 @@ func NewParamRepository(db *gorm.DB) *ParamRepository {
...
 // CreateBatch creates []models.Param entities in batch.
 func (r ParamRepository) CreateBatch(ctx context.Context, batchSize int, params []models.Param) error {
-	if err := r.db.CreateInBatches(params, batchSize).Error; err != nil {
+	var err error
+	err = r.db.CreateInBatches(params, batchSize).Error
+	if err != nil {
+		// remove duplicate rows and try again
+		params, err1 := r.removeExactMatches(ctx, params)
+		if err1 != nil {
+			return eris.Wrap(err1, "error removing exact matches after: " + err.Error())
```
dsuhinin, line 41: please do not use `err.Error()`. We have to bubble up all the errors and stack trace.

pkg/api/mlflow/dao/repositories/param.go
```diff
@@ -31,8 +32,50 @@ func NewParamRepository(db *gorm.DB) *ParamRepository {
...
 		return eris.Wrap(err, "error creating params in batch")
 	}
 	return nil
 }
+
+// removeExactMatches will return a new slice of params which excludes exact matches
+func (r ParamRepository) removeExactMatches(ctx context.Context, params []models.Param) ([]models.Param, error) {
+	var keys, foundKeys []string
+	paramMap := map[string]models.Param{}
+	paramsToReturn := []models.Param{}
```
dsuhinin, line 56: can we please init variable new the place where we use it?

pkg/api/mlflow/dao/repositories/param.go
```diff
@@ -31,8 +32,50 @@ func NewParamRepository(db *gorm.DB) *ParamRepository {
...
+	tx := r.db.Raw(`
+           select run_uuid || '-' || key || '-' || value
+           from params
+           where run_uuid || '-' || key || '-' || value in ?`, keys).
+		Find(&foundKeys)
+	if tx.Error != nil {
+		return paramsToReturn, eris.Wrap(tx.Error, "problem selecting existing params")
+	}
+
+	for _, foundKey := range foundKeys {
```
dsuhinin, line 72: ohh, we have 3 cycles. are we 100% sure about that?

pkg/api/mlflow/dao/repositories/param.go
```diff
@@ -31,8 +32,50 @@ func NewParamRepository(db *gorm.DB) *ParamRepository {
...
+	paramMap := map[string]models.Param{}
+	paramsToReturn := []models.Param{}
+	for _, param := range params {
+		key := fmt.Sprintf("%v-%v-%v", param.RunID, param.Key, param.Value)
+		keys = append(keys, key)
+		paramMap[key] = param
+	}
+
+	tx := r.db.Raw(`
+           select run_uuid || '-' || key || '-' || value
```
dsuhinin, line 64: @jgiannuzzi @suprjinx  what about next solution -> when we create a param, lets calculate a `hash` into the own column and then let's try to use this hash to find duplicated.

pkg/api/mlflow/dao/repositories/param.go
```diff
@@ -31,8 +32,50 @@ func NewParamRepository(db *gorm.DB) *ParamRepository {
...
+	var keys, foundKeys []string
+	paramMap := map[string]models.Param{}
+	paramsToReturn := []models.Param{}
+	for _, param := range params {
+		key := fmt.Sprintf("%v-%v-%v", param.RunID, param.Key, param.Value)
+		keys = append(keys, key)
+		paramMap[key] = param
+	}
+
+	tx := r.db.Raw(`
```
dsuhinin, line 63: you don't need to create a var here. you can put everything into `if` statement and get `r.db.Raw().Find.Error` immediately.

tests/integration/golang/mlflow/run/log_parameter_test.go
```diff
@@ -1,3 +1,190 @@
...
+	assert.Empty(s.T(), resp)
+}
+
+func (s *LogParamTestSuite) TestBatch_Ok() {
+	tests := []struct {
+		name    string
+		request *request.LogBatchRequest
+	}{
+		{
+			name: "Batch of one should succeed",
```
dsuhinin, line 92: please do not use whitespaces in test names. When test failed, all whitespaces will be removed and impossible to find failed test.

tests/integration/golang/mlflow/run/log_parameter_test.go
```diff
@@ -1,3 +1,190 @@
...
+
+	runID := uuid.New().String()
+	run := &models.Run{
+		ID:             strings.ReplaceAll(runID, "-", ""),
+		ExperimentID:   *exp.ID,
+		SourceType:     "JOB",
+		LifecycleStage: models.LifecycleStageActive,
+		Status:         models.StatusRunning,
+	}
+	run, err = s.runFixtures.CreateTestRun(context.Background(), run)
```
dsuhinin, line 61: probably will be nice to create `run` in each test. In that case logic is fully isolated and not related to other tests/results.

tests/integration/golang/mlflow/run/log_parameter_test.go
```diff
@@ -1,3 +1,190 @@
...
+		Value: "value1",
+	}
+	resp := map[string]any{}
+	err := s.client.DoPostRequest(
+		fmt.Sprintf("%s%s", mlflow.RunsRoutePrefix, mlflow.RunsLogParameterRoute),
+		req,
+		&resp,
+	)
+	assert.Nil(s.T(), err)
+	assert.Empty(s.T(), resp)
```
dsuhinin, line 83: probably Im not sure that we need to check that moment.

tests/integration/golang/mlflow/run/log_parameter_test.go
```diff
@@ -1,3 +1,190 @@
...
+					},
+					{
+						Key:   "key2",
+						Value: "value2",
+					},
+				},
+			},
+		},
+	}
+	var err error
```
dsuhinin, line 120: what do we need this var?

tests/integration/golang/mlflow/run/log_parameter_test.go
```diff
@@ -1,3 +1,190 @@
...
+		assert.Nil(s.T(), s.runFixtures.UnloadFixtures())
+	}()
+
+	var testData = []struct {
+		name    string
+		error   string
+		request *request.LogBatchRequest
+	}{
+		{
+			name:  "Missing required field",
```
dsuhinin, line 146: the same problem with naming.

tests/integration/golang/mlflow/run/log_parameter_test.go
```diff
@@ -1,3 +1,190 @@
...
+		SourceType:     "JOB",
+		LifecycleStage: models.LifecycleStageActive,
+		Status:         models.StatusRunning,
+	}
+	run, err = s.runFixtures.CreateTestRun(context.Background(), run)
+	assert.Nil(s.T(), err)
+	s.run = run
+}
+
+func (s *LogParamTestSuite) Test_Ok() {
```
dsuhinin, line 66: this test is not related to `batch`, can we please move it into own test file?

tests/integration/golang/mlflow/run/log_parameter_test.go
```diff
@@ -1,3 +1,190 @@
...
+		name    string
+		error   string
+		request *request.LogBatchRequest
+	}{
+		{
+			name:  "Missing required field",
+			error: "Missing value for required parameter 'run_id'",
+			request: &request.LogBatchRequest{
+				RunID: "",
+				Params: []request.ParamPartialRequest{
```
dsuhinin, line 150: probably you don't need to pass `params` here?

dsuhinin (changes requested):

pkg/api/mlflow/dao/repositories/param.go
suprjinx: interesting idea

pkg/api/mlflow/dao/repositories/param.go
suprjinx: no sure if there's a way to avoid without some "set operations" library or the like, which also range over collections as implementation

pkg/api/mlflow/dao/repositories/param.go
dsuhinin: I can see one problem here - how it should work for old records without the hash? maybe we can create a command line tool, which will migrate old data and add missing hashes?

pkg/api/mlflow/dao/repositories/param.go
dsuhinin: with `hashes` we can immediately remove all duplicates and re create new ones.

pkg/api/mlflow/dao/repositories/param.go
dsuhinin: let's try to use more plain constructs. now it is a bit hard to understand what is going on when you read. which error belongs to which block. maybe we can change this blocks so the could be in more plain way? without nesting?

dsuhinin: why we can't before clean? and then just create?

tests/integration/golang/mlflow/run/log_parameter_test.go
```diff
@@ -1,3 +1,190 @@
...
+func (s *LogParamTestSuite) SetupTest() {
+	s.client = helpers.NewHttpClient(os.Getenv("SERVICE_BASE_URL"))
+	runFixtures, err := fixtures.NewRunFixtures(os.Getenv("DATABASE_DSN"))
+	assert.Nil(s.T(), err)
+	s.runFixtures = runFixtures
+	expFixtures, err := fixtures.NewExperimentFixtures(os.Getenv("DATABASE_DSN"))
+	assert.Nil(s.T(), err)
+	s.experimentFixtures = expFixtures
+
+	expName := uuid.New().String()
```
dsuhinin, line 45: this is not re used somewhere so you use it inline:
```
Name: uuid.New().String()
```

tests/integration/golang/mlflow/run/log_parameter_test.go
```diff
@@ -1,3 +1,190 @@
...
+	exp := &models.Experiment{
+		Name:           expName,
+		LifecycleStage: models.LifecycleStageActive,
+	}
+	_, err = s.experimentFixtures.CreateTestExperiment(context.Background(), exp)
+	assert.Nil(s.T(), err)
+
+	runID := uuid.New().String()
```
dsuhinin, line 53: the same probably here.

tests/integration/golang/mlflow/run/log_parameter_test.go
dsuhinin: please add into the name, which required field is missing. it immediately give you all needed information when test failed.

pkg/api/mlflow/dao/repositories/param.go
suprjinx: I do like this idea, especially because we could then use `ON CONFLICT(hash_column) DO NOTHING` and make things much simpler. Perhaps old records wouldn't be getting new log params, so maybe it's okay without backfill?

pkg/api/mlflow/dao/repositories/param.go
suprjinx: well the issue is we have to subtract a set of keys from the hash in this implementation -- but maybe I'm not seeing it

tests/integration/golang/mlflow/run/log_parameter_test.go
suprjinx: the compiler should inline it ;-). But I will move.

tests/integration/golang/mlflow/run/log_parameter_test.go
suprjinx: well the Run instance isn't actually part of the test, right -- it's a fixture for Param.

tests/integration/golang/mlflow/run/log_parameter_test.go
suprjinx: will do

tests/integration/golang/mlflow/run/log_parameter_test.go
suprjinx: I think human readable in the test source file is probably more important than the error spit out by failure

tests/integration/golang/mlflow/run/log_parameter_test.go
dsuhinin: yep yep, but unfortunately we still need to support and read it =(

pkg/api/mlflow/dao/repositories/param.go
suprjinx: will rename

pkg/api/mlflow/dao/repositories/param.go
suprjinx: will remove var reuse

pkg/api/mlflow/dao/repositories/param.go
suprjinx: will wrap

suprjinx: well at this point we have 2 errors to report and Wrap() does not seem to allow multiple error args

pkg/api/mlflow/dao/repositories/param.go
suprjinx: moved

tests/integration/golang/mlflow/run/log_parameter_test.go
suprjinx: `--- PASS: TestLogBatchTestSuite (0.03s)
    --- PASS: TestLogBatchTestSuite/Test_Error (0.03s)
        --- PASS: TestLogBatchTestSuite/Test_Error/Missing_required_field (0.00s)
        --- PASS: TestLogBatchTestSuite/Test_Error/Duplicate_keys_with_different_values_should_error (0.00s)`

tests/integration/golang/mlflow/run/log_parameter_test.go
suprjinx: the API returns an empty body for this, seemed worth asserting

tests/integration/golang/mlflow/run/log_parameter_test.go
dsuhinin: convention is do not use `whitespaces` because as you can see they will be remove with `underscores` and when test failed, then you will try to fins something like `Missing_required_field` over the code but..... your test have `whitesapces`

tests/integration/golang/mlflow/run/log_parameter_test.go
suprjinx: well to illustrate the request is otherwise well formed?

tests/integration/golang/mlflow/run/log_parameter_test.go
dsuhinin: so the name has to be `MissingRequiredField` + add more details about which filed is missing.

tests/integration/golang/mlflow/run/log_parameter_test.go
dsuhinin: but you test this case, right -> `RunID: ""`, why do you need to pass other parameters?

dsuhinin: they are not necessary, because validation won't check them.

tests/integration/golang/mlflow/run/log_parameter_test.go
suprjinx: breaking up code into smaller steps is often helpful for us humans :-) but I'll move this inline

pkg/api/mlflow/dao/repositories/param.go
suprjinx: now discards first error

pkg/api/mlflow/dao/repositories/param.go
suprjinx: decided to keep this in the error-only path

tests/integration/golang/mlflow/run/log_parameter_test.go
suprjinx: is this really a style recommendation in Go community? I think human-readable messages are important

tests/integration/golang/mlflow/run/log_parameter_test.go
suprjinx: done

tests/integration/golang/mlflow/run/log_parameter_test.go
suprjinx: okay will remove

tests/integration/golang/mlflow/run/log_parameter_test.go
```diff
@@ -1,3 +1,102 @@
...
+		assert.Nil(s.T(), s.runFixtures.UnloadFixtures())
+	}()
+
+	// missing run_id
+	req := request.LogParamRequest{
+		Key:   "key1",
+		Value: "value1",
+	}
+	resp := map[string]any{}
+	_ = s.client.DoPostRequest(
```
dsuhinin, line 95: please check `err` here.

tests/integration/golang/mlflow/run/log_parameter_test.go
```diff
@@ -1,3 +1,102 @@
...
+		fmt.Sprintf("%s%s", mlflow.RunsRoutePrefix, mlflow.RunsLogParameterRoute),
+		req,
+		&resp,
+	)
+	assert.Equal(s.T(), "INVALID_PARAMETER_VALUE", resp["error_code"])
```
dsuhinin, line 100: can we please use existing error objects to check error? - `NewInvalidParameterValueError()`

tests/integration/golang/mlflow/run/log_parameter_test.go
```diff
@@ -1,3 +1,102 @@
...
+	defer func() {
+		assert.Nil(s.T(), s.runFixtures.UnloadFixtures())
+	}()
+
+	// missing run_id
+	req := request.LogParamRequest{
+		Key:   "key1",
+		Value: "value1",
+	}
+	resp := map[string]any{}
```
dsuhinin, line 94: since you are waiting for the error, please unwrap response into `Error` response - `api.ErrorResponse{}` and then use it to check error data.

dsuhinin (changes requested):
Some minor things.

tests/integration/golang/mlflow/run/log_parameter_test.go
suprjinx: will do, but in this case the test will obviously fail if the http request is unsuccessful so it is superfluous.

tests/integration/golang/mlflow/run/log_parameter_test.go
suprjinx: will do

tests/integration/golang/mlflow/run/log_parameter_test.go
suprjinx: will do

dsuhinin (dismissed):
looks good!

dsuhinin (dismissed):

suprjinx:
Moved to PR #89

PR #72, opened 2023-06-12
Fix run deletion

suprjinx:
This PR fixes the run deletion mechanism that was introduced in #35. It's adding cascade support for the objects associated to a run, and it renumbers `row_num` where needed.


pkg/api/mlflow/dao/repositories/run.go
```diff
@@ -209,3 +217,16 @@ func (r RunRepository) SetRunTagsBatch(ctx context.Context, run *models.Run, bat
...
+func (r RunRepository) renumberRows(ctx context.Context) error {
+	tx := r.db.WithContext(ctx).Raw(
+		`UPDATE runs
+	         SET row_num = rows.row_num
+	         FROM (
+	           SELECT run_uuid, ROW_NUMBER() OVER (ORDER BY start_time, run_uuid DESC) - 1 AS row_num
+		   FROM runs
+		 ) AS rows
+		 WHERE runs.run_uuid = rows.run_uuid`)
+	return tx.Error
```
dsuhinin, line 231: please can you wrap all the errors?

dsuhinin (changes requested):

jgiannuzzi (commented):
Optimisation idea: only renumber runs that are _after_ the deleted run

pkg/database/database.go
```diff
@@ -312,6 +312,7 @@ func ConnectDB(
 				}); err != nil {
 					return nil, fmt.Errorf("error migrating database to FastTrackML schema 8073e7e037e5: %w", err)
 				}
+				fallthrough
```
jgiannuzzi, line 315: good eye, thanks for spotting this!

pkg/database/database.go
```diff
@@ -329,6 +330,42 @@ func ConnectDB(
...
+					}
+					if err := tx.Migrator().CreateConstraint(&Run{}, "Tags"); err != nil {
+						return err
+					}
+					if err := tx.Migrator().CreateConstraint(&Run{}, "Metrics"); err != nil {
+						return err
+					}
+					if err := tx.Migrator().CreateConstraint(&Run{}, "LatestMetrics"); err != nil {
+						return err
+					}
```
jgiannuzzi, line 361: Could you make it a loop over the 4 columns in which you drop+create?

pkg/api/mlflow/dao/repositories/run.go
```diff
@@ -209,3 +243,34 @@ func (r RunRepository) SetRunTagsBatch(ctx context.Context, run *models.Run, bat
...
+	// get the lowest rownum and experiment id from batch
+	minRowNum := int32(0)
+	if err := tx.Raw(
+		`SELECT MIN(row_num)
+                 FROM runs
+                 WHERE run_uuid IN ?`, runIDs).First(&minRowNum).Error; err != nil {
+		return 0, eris.Wrapf(err, "error finding the lowest row num from run.IDs")
+	}
+	return minRowNum, nil
+}
```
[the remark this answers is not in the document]
dsuhinin: Jonathan, are you said this, because we want to be database agnostic as much as possible?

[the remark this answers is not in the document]
jgiannuzzi: Indeed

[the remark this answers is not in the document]
jgiannuzzi: This does not preclude us from switching to something else than Gorm in the future — however at this stage I'd like to keep consistency.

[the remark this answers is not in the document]
suprjinx: Will do, but I would suggest standard sql92 is also database agnostic and is better understood as a universal language than the Gorm api :-)

pkg/api/mlflow/dao/repositories/run.go
```diff
@@ -157,19 +157,13 @@ func (r RunRepository) ArchiveBatch(ctx context.Context, ids []string) error {
...
 	if err := r.db.Transaction(func(tx *gorm.DB) error {
-		// find the row_num for the run
-		minRowNum, err := r.getMinRowNum(tx, []string{run.ID})
-		if err != nil {
-			return err
-		}
-
 		// delete the row
-		if err := tx.Model(&run).Delete(run).Error; err != nil {
+		if err := tx.Clauses(clause.Returning{}).Model(&run).Delete(&run).Error; err != nil {
```
jgiannuzzi, line 161: could you please narrow down the `Returning` clause to only the `row_num` column?

jgiannuzzi, line 161: there should be no need to call `.Model()`, as we are passing a model to `.Delete()`

pkg/api/mlflow/dao/repositories/run.go
```diff
@@ -157,19 +157,13 @@ func (r RunRepository) ArchiveBatch(ctx context.Context, ids []string) error {
...
 			return eris.Wrapf(err, "error deleting run with id: %s", run.ID)
 		}
 
 		// renumber the remainder
-		if err := r.renumberRows(tx, minRowNum); err != nil {
+		if err := r.renumberRows(tx, getMinRowNum([]models.Run{*run})); err != nil {
```
jgiannuzzi, line 166: we know there is only one run here, so let's just pass `run.RowNum` to `renumberRows` directly

pkg/api/mlflow/dao/repositories/run.go
```diff
@@ -272,21 +260,24 @@ func (r RunRepository) SetRunTagsBatch(ctx context.Context, run *models.Run, bat
...
-                 WHERE run_uuid IN ?`, runIDs).First(&minRowNum).Error; err != nil {
-		return 0, eris.Wrapf(err, "error finding the lowest row num from run.IDs")
+// getMinRowNum will find the lowest row_num for the slice of runs
+// or 0 for an empty slice
+func getMinRowNum(runs []models.Run) int64 {
+	if len(runs) == 0 {
+		return int64(0)
+	}
+	// get the lowest row_num in the slice
+	minRowNum := int64(runs[0].RowNum)
```
jgiannuzzi, line 270: just set `minRowNum` to the zero value before looping

pkg/api/mlflow/dao/repositories/run.go
```diff
@@ -237,3 +259,37 @@ func (r RunRepository) SetRunTagsBatch(ctx context.Context, run *models.Run, bat
...
+	for _, run := range runs {
+		if int64(run.RowNum) < minRowNum {
+			minRowNum = int64(run.RowNum)
+		}
+	}
+	return minRowNum
```
jgiannuzzi, line 276: make `minRowNum` a `RowNum` type and only cast to `int64` when returning

pkg/api/mlflow/dao/repositories/run.go
```diff
@@ -237,3 +259,37 @@ func (r RunRepository) SetRunTagsBatch(ctx context.Context, run *models.Run, bat
...
+}
+
+// renumberRows will update the runs.row_num field with the correct ordinal
+func (r RunRepository) renumberRows(tx *gorm.DB, startWith int64) error {
```
jgiannuzzi, line 280: make sure we don't do a complete renumbering when there was no run to delete!

pkg/api/mlflow/dao/repositories/run.go
```diff
@@ -237,3 +259,37 @@ func (r RunRepository) SetRunTagsBatch(ctx context.Context, run *models.Run, bat
 	}
 	return nil
 }
+
+// getMinRowNum will find the lowest row_num for the slice of runs
+// or 0 for an empty slice
+func getMinRowNum(runs []models.Run) int64 {
```
jgiannuzzi, line 265: one more thing: you could get a slice of runs that have the zero value for their `RowNum` because they did not exist in the first place. This could all be avoided by checking that no bad IDs were passed by the user before we delete them.

pkg/api/mlflow/dao/repositories/run.go
```diff
@@ -237,3 +250,37 @@ func (r RunRepository) SetRunTagsBatch(ctx context.Context, run *models.Run, bat
 	}
 	return nil
 }
+
+// getMinRowNum will find the lowest row_num for the slice of runs
+// or 0 for an empty slice
+func getMinRowNum(runs []models.Run) models.RowNum {
+	// get the lowest row_num in the slice
+	minRowNum := models.RowNum(0)
```
jgiannuzzi, line 258: the comment is redundant, and I _think_ it's more readable to just declare the variable like this?
```suggestion
	var minRowNum models.RowNum
```

pkg/api/mlflow/dao/repositories/run.go
```diff
@@ -156,18 +156,31 @@ func (r RunRepository) ArchiveBatch(ctx context.Context, ids []string) error {
...
+			return eris.Errorf("non-existent run id provided")
+		}
+
+		// renumber the remainder
+		if err := r.renumberRows(tx, getMinRowNum(runs)); err != nil {
+			return eris.Wrapf(err, "error renumbering runs.row_num")
+		}
+		return nil
+	}); err != nil {
+		return eris.Wrapf(err, "error renumbering runs.row_num")
```
jgiannuzzi, line 183: I think this error is about deleting runs, not renumbering them per se, as it's at the transaction level

pkg/api/mlflow/dao/repositories/run.go
```diff
@@ -237,3 +250,37 @@ func (r RunRepository) SetRunTagsBatch(ctx context.Context, run *models.Run, bat
...
 	return nil
 }
+
+// getMinRowNum will find the lowest row_num for the slice of runs
+// or 0 for an empty slice
+func getMinRowNum(runs []models.Run) models.RowNum {
+	// get the lowest row_num in the slice
+	minRowNum := models.RowNum(0)
+	for _, run := range runs {
+		if minRowNum == models.RowNum(0) || run.RowNum < minRowNum {
```
jgiannuzzi, line 260: my suggestion to compare to the zero value was to avoid setting `minRowNum` to an invalid value, but it's not useful anymore now that we're checking that there are no non-existent run IDs provided

pkg/api/mlflow/dao/repositories/run.go
```diff
@@ -156,18 +156,31 @@ func (r RunRepository) ArchiveBatch(ctx context.Context, ids []string) error {
...
+		runs := []models.Run{}
+		tx.Clauses(clause.Returning{Columns: []clause.Column{{Name: "row_num"}}}).Where("run_uuid IN ?", ids).Delete(&runs)
+		if tx.Error != nil {
+			return eris.Wrapf(tx.Error, "error deleting existing runs with ids: %s", ids)
+		}
+
+		// verify deletion
+		if len(runs) != len(ids) {
+			return eris.Errorf("non-existent run id provided")
+		}
```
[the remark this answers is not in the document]
suprjinx: tx.RowsAffected is always zero for this query, but the runs slice returned is the length of affected rows

pkg/api/mlflow/dao/repositories/run.go
```diff
@@ -156,18 +156,31 @@ func (r RunRepository) ArchiveBatch(ctx context.Context, ids []string) error {
...
-	run := models.Run{}
-	if err := r.db.WithContext(ctx).Model(&run).Where("run_uuid IN ?", ids).Delete(run).Error; err != nil {
-		return eris.Wrapf(err, "error deleting existing runs with ids: %s", ids)
+	if err := r.db.Transaction(func(tx *gorm.DB) error {
+		// delete the rows
+		runs := []models.Run{}
+		tx.Clauses(clause.Returning{Columns: []clause.Column{{Name: "row_num"}}}).Where("run_uuid IN ?", ids).Delete(&runs)
+		if tx.Error != nil {
+			return eris.Wrapf(tx.Error, "error deleting existing runs with ids: %s", ids)
+		}
```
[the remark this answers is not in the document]
suprjinx: ah yes, this was left behind by trying to look at tx.RowsAffected. Will change

pkg/api/mlflow/dao/repositories/run.go
suprjinx: done

pkg/api/mlflow/dao/repositories/run.go
suprjinx: i think we need this to get the first value, otherwise the zero-value will always be lowest

pkg/api/mlflow/dao/repositories/run.go
suprjinx: oops, fixed

pkg/api/mlflow/dao/repositories/run.go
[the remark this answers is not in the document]
jgiannuzzi: ah alright, LGTM then!

pkg/api/mlflow/dao/repositories/run.go
jgiannuzzi: ah yes, good point!

pkg/api/mlflow/dao/repositories/run.go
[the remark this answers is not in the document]
dsuhinin: error message in this case a bit strange?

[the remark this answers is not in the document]
jgiannuzzi: maybe `invalid run ID`?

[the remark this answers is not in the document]
dsuhinin: I mean we compare lengths but say about not existing of run which is confusing when you read the code.

pkg/api/mlflow/dao/repositories/run.go
```diff
@@ -156,18 +156,31 @@ func (r RunRepository) ArchiveBatch(ctx context.Context, ids []string) error {
...
 }
 
 // DeleteBatch removes existing models.Run from the db.
 func (r RunRepository) DeleteBatch(ctx context.Context, ids []string) error {
-	run := models.Run{}
-	if err := r.db.WithContext(ctx).Model(&run).Where("run_uuid IN ?", ids).Delete(run).Error; err != nil {
-		return eris.Wrapf(err, "error deleting existing runs with ids: %s", ids)
+	if err := r.db.Transaction(func(tx *gorm.DB) error {
+		// delete the rows
+		runs := []models.Run{}
```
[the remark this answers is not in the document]
suprjinx: okay will do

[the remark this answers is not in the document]
jgiannuzzi: thanks!

pkg/api/mlflow/dao/repositories/run.go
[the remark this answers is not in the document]
suprjinx: if such a comment will get this PR approved, I can add. :-). Will change the error message for length mismatch

jgiannuzzi (approved):
LGTM 👍

Thanks a lot for your patience and flexibility @suprjinx, it's much appreciated!

jgiannuzzi (approved):

dsuhinin (approved):

PR #73, opened 2023-06-12
Integration tests for MLFlow experiments API

fabiovincenzi:
┆Issue is synchronized with this [Jira Task](https://gr-oss.atlassian.net/browse/FML-55) by [Unito](https://www.unito.io)


tests/integration/golang/mlflow/experiment/list_test.go
```diff
@@ -1,3 +1,80 @@
...
+		fmt.Sprintf(
+			"%s%s?%s", mlflow.ExperimentsRoutePrefix, mlflow.ExperimentsSearchRoute, query,
+		),
+		&resp,
+	)
+	assert.Nil(s.T(), err)
+	assert.Equal(s.T(), 4, len(resp.Experiments))
+}
+
+func (s *ListExperimentsTestSuite) Test_Error() {
```
dsuhinin, line 79: do you plan to add tests for error cases as well?

tests/integration/golang/mlflow/experiment/search_test.go
```diff
@@ -1,3 +1,255 @@
...
+			ArtifactLocation: "/artifact/location",
+		})
+		assert.Nil(s.T(), err)
+	}
+	defer func() {
+		assert.Nil(s.T(), s.fixtures.UnloadFixtures())
+	}()
+
+	// Test Filter
+	resp := GetSearchExperimentsResponse(s, request.SearchExperimentsRequest{
```
dsuhinin, line 141: you can probably re organise this tests to be table test. in that case you will have own sub test for each particular filter and test will be a bit more readable.

dsuhinin: also you won't need in that case to have a `GetSearchExperimentsResponse` function.

dsuhinin: and this function `GetExperimentNames` will be unnecessary as well.

tests/integration/golang/mlflow/experiment/search_test.go
```diff
@@ -1,3 +1,255 @@
...
+	// Test MaxResults
+	resp = GetSearchExperimentsResponse(s, request.SearchExperimentsRequest{
+		OrderBy:    []string{"name ASC"},
+		MaxResults: 3,
+	})
+
+	assert.ElementsMatch(s.T(), []string{"Default", "Test Experiment 1", "Test Experiment 2"}, GetExperimentNames(resp.Experiments))
+}
+
+func (s *SearchExperimentsTestSuite) Test_Error() {
```
dsuhinin, line 170: Niceeee! we have tests for this pretty complex endpoint so it will be much safer to change it after your changes. Great work!

tests/integration/golang/mlflow/experiment/set_experiment_tag_test.go
```diff
@@ -1,3 +1,208 @@
...
+	s.fixtures = fixtures
+}
+
+func SetExperimentTag(s *SetExperimentTagTestSuite, experiment *models.Experiment, key, value string) {
+	req := request.SetExperimentTagRequest{
+		ID:    fmt.Sprintf("%d", *experiment.ID),
+		Key:   key,
+		Value: value,
+	}
+	resp := fiber.Map{}
```
dsuhinin, line 49: if response returns nothing we probably don't need it? what do you think?

tests/integration/golang/mlflow/experiment/set_experiment_tag_test.go
```diff
@@ -1,3 +1,208 @@
...
+	exp, err = s.fixtures.GetExperimentByID(context.Background(), *experiment.ID)
+	assert.Nil(s.T(), err)
+
+	found = false
+	for _, tag := range exp.Tags {
+		if tag.Key == "KeyTag1" && tag.Value == "ValueTag2" {
+			found = true
+			break
+		}
+	}
```
dsuhinin, line 119: Let put such functions to `helper` file near the test. In that case your test will be much cleaner and understandable.

tests/integration/golang/mlflow/experiment/set_experiment_tag_test.go
```diff
@@ -1,3 +1,208 @@
...
+	assert.Nil(s.T(), err)
+	exp1, err := s.fixtures.GetExperimentByID(context.Background(), *experiment1.ID)
+	assert.Nil(s.T(), err)
+	found = false
+	for _, tag := range exp.Tags {
+		if tag.Key == "KeyTag1" && tag.Value == "ValueTag2" {
+			found = true
+			break
+		}
+	}
```
dsuhinin, line 139: yep, this guy has to be as a helper function.

tests/integration/golang/mlflow/experiment/set_experiment_tag_test.go
```diff
@@ -1,3 +1,208 @@
...
+		error   *api.ErrorResponse
+		request *request.SetExperimentTagRequest
+	}{
+		{
+			name:  "EmptyIDProperty",
+			error: api.NewInvalidParameterValueError("Missing value for required parameter 'experiment_id'"),
+			request: &request.SetExperimentTagRequest{
+				ID:    "",
+				Key:   "test_key",
+				Value: "test_value",
```
dsuhinin, line 164: probably you don't need to pass this values in this test, right? Test will fail immediately on `ID` check so you can just send an empty request.

tests/integration/golang/mlflow/experiment/set_experiment_tag_test.go
```diff
@@ -1,3 +1,208 @@
...
+			},
+		},
+		{
+			name:  "EmptyKeyProperty",
+			error: api.NewInvalidParameterValueError("Missing value for required parameter 'key'"),
+			request: &request.SetExperimentTagRequest{
+				ID:    "1",
+				Key:   "",
+				Value: "test_value",
```
dsuhinin, line 173: here you don't need to send `value`, because test will fail on `key`. Just send as minimum as possible in that case it is understand what is going on.

tests/integration/golang/mlflow/experiment/set_experiment_tag_test.go
```diff
@@ -1,3 +1,208 @@
...
+			},
+		},
+		{
+			name:  "IncorrectExperimentID",
+			error: api.NewBadRequestError(`Unable to parse experiment id 'incorrect_experiment_id': strconv.ParseInt: parsing "incorrect_experiment_id": invalid syntax`),
+			request: &request.SetExperimentTagRequest{
+				ID:    "incorrect_experiment_id",
+				Key:   "test_key",
+				Value: "test_value",
```
dsuhinin, line 182: here you probably don't need `key` and `value`.

tests/integration/golang/mlflow/experiment/set_experiment_tag_test.go
```diff
@@ -1,3 +1,208 @@
...
+			},
+		},
+		{
+			name:  "NotFoundExperiment",
```
dsuhinin, line 186: this is only one tests where you have to send all the parameters to pass first step - `validation` step.

tests/integration/golang/mlflow/experiment/update_test.go
```diff
@@ -1,3 +1,123 @@
...
+import (
+	"context"
+	"database/sql"
+	"fmt"
+	"os"
+	"testing"
+	"time"
+
+	"github.com/gofiber/fiber/v2"
+
```
dsuhinin, line 14: please group `import` near each other.

tests/integration/golang/mlflow/experiment/update_test.go
```diff
@@ -1,3 +1,123 @@
...
+	assert.Nil(s.T(), err)
+	defer func() {
+		assert.Nil(s.T(), s.fixtures.UnloadFixtures())
+	}()
+
+	req := request.UpdateExperimentRequest{
+		ID:   fmt.Sprintf("%d", *experiment.ID),
+		Name: "Test Updated Experiment",
+	}
+	resp := fiber.Map{}
```
dsuhinin, line 67: we probably don't need it?

tests/integration/golang/mlflow/experiment/update_test.go
```diff
@@ -1,3 +1,123 @@
...
+		name    string
+		error   *api.ErrorResponse
+		request *request.UpdateExperimentRequest
+	}{
+		{
+			name:  "EmptyIDProperty",
+			error: api.NewInvalidParameterValueError("Missing value for required parameter 'experiment_id'"),
+			request: &request.UpdateExperimentRequest{
+				ID:   "",
+				Name: "New Name",
```
dsuhinin, line 90: the same comment as above. please just send as minimum as you need in current test.

tests/integration/golang/mlflow/experiment/update_test.go
```diff
@@ -1,3 +1,123 @@
...
+			name:  "EmptyNameProperty",
+			error: api.NewInvalidParameterValueError("Missing value for required parameter 'new_name'"),
+			request: &request.UpdateExperimentRequest{
+				ID:   "1",
+				Name: "",
+			},
+		},
+		{
+			name:  "InvalidIDFormat",
+			error: api.NewBadRequestError("unable to parse experiment id 'invalid_id': strconv.ParseInt: parsing \"invalid_id\": invalid syntax"),
```
dsuhinin, line 103: you can use `backticks` to avoid `\"`, just wrap everything like this -> ``unable to parse experiment id 'invalid_id': strconv.ParseInt: parsing "invalid_id": invalid syntax``

dsuhinin (changes requested):
Looks good! Left few minor comments.

tests/integration/golang/mlflow/experiment/set_experiment_tag_test.go
fabiovincenzi: We need key here because we first check if key is empty then we search for the experiment id, but yes no value needed

tests/integration/golang/mlflow/experiment/search_test.go
```diff
@@ -1,3 +1,262 @@
...
+	fixtures, err := fixtures.NewExperimentFixtures(os.Getenv("DATABASE_DSN"))
+	assert.Nil(s.T(), err)
+	s.fixtures = fixtures
+}
+
+func (s *SearchExperimentsTestSuite) Test_Ok() {
+	// 1. prepare database with test data.
+	experiments := []models.Experiment{
+		{
+			Name: "Test Experiment 1",
```
dsuhinin, line 47: please do not use `whitespaces` in test names. When test will fail, all `whitespaces` will be changed to underscores, so when you will try to find this test by test name and copy past test name with underscores you wont find it.

tests/integration/golang/mlflow/experiment/search_test.go
```diff
@@ -1,3 +1,262 @@
...
+		assert.Nil(s.T(), s.fixtures.UnloadFixtures())
+	}()
+
+	tests := []struct {
+		name     string
+		request  request.SearchExperimentsRequest
+		expected []string
+	}{
+		{
+			name: "Test Filter",
```
dsuhinin, line 125: Please do not use `whitespaces` in test names. They will be changed to `underscores` so you won't find them in case if something just failed.

dsuhinin (approved):
Great!

PR #88, opened 2023-06-14
Integrate GolangCI linter.

dsuhinin:
┆Issue is synchronized with this [Jira Task](https://gr-oss.atlassian.net/browse/FML-70) by [Unito](https://www.unito.io)


Makefile
suprjinx, line 1: In general I think development is much faster if the make targets are designed to run inside a docker container or directly on a host instead of launching docker. golangci-lint is already built into .devcontainer/Dockerfile so it can be run directly in that container.

dsuhinin: I can't see it in .devcontainer/Dockerfile - emmm, where?

dsuhinin: also image build is like a final step. `linting` is a pre step before even tests I think.

dsuhinin: or maybe I misunderstood something?

dsuhinin: probably this target will works for everybody, I mean I even don't need to run a container which I probably not do on my local.

suprjinx: similarly, tests can be run in without additional docker invocations and Dockerfiles, so why add them?

dsuhinin: sorry, I don't understand you.

suprjinx: it's really slow to build docker containers for running tests, lint, etc

dsuhinin: em not sure, slow when? on which stage? Please can you give an example? Or it would be even better if you can just improve it. We probably need to start with something and then make performance improvements. So I probably will be happy if we can improve and speed up of what we have right now.

dsuhinin: I mean right now our goal is to make everything:
- testable
- to have well organised code for both `aim` and `mlflow` parts.
- to have code standards
- to have good linting process.
after that of course we can work with build pipeline improvements, tests performance and other things you mentioned.

dsuhinin: no no, maybe we don't need it. I just want to say that my work is iterative. create right now what I planned and then revisit it a bit later to improve. If you can you can probably adjust my PR or we can do that later as a separate task. so up to you really.

jgiannuzzi: I agree with @suprjinx that we should avoid introducing more docker things where possible. The target development environment is the devcontainer, so we just need to ensure any tools we need are in [its definition](https://github.com/G-Research/fasttrackml/blob/main/.devcontainer/Dockerfile). Adding something along these lines should do the trick:
```
# Install additional Go tools
RUN go install github.com/vektra/mockery/v2@v2.20.0 \
 && go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.53.2
```

(I realised that the current devcontainer is broken for building at the moment because mockery was not installed and the UIs can't be built _in my setup_ because the upstream repos are cloned in the mounted filesystem — please don't break the devcontainer even if you don't use it!)

dsuhinin: emmmm, but I didn't touch it, I mean devcontainer.

dsuhinin: @jgiannuzzi also, can we do that a bit later? like in separate PR?

dsuhinin: also if we want to integration it into `devcontainer` I think in that case it won't be like a separate step in github actions. everything will be included into Build docker image step.

jgiannuzzi: I meant that the `Makefile` doesn't work as-is on the devcontainer (because mockery is missing), and that the UI build scripts have been modified in #22 and broke my _specific_ setup (devcontainer in VSCode on macOS with a shared filesystem).

For this particular PR, I believe that adding a line that installs golangci-lint in the devcontainer (_instead_ of having an extra `Dockerfile`) and modifying the `Makefile` to just invoke the native `golangci-lint` command would do the trick.

dsuhinin: oook. it seems like it should be fixed.

.devcontainer/Dockerfile
```diff
@@ -12,12 +12,13 @@ RUN apt-get update && export DEBIAN_FRONTEND=noninteractive \
...
+RUN go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.53.3
+USER root
 
 # [Optional] Uncomment this line to install global node packages.
 # RUN su vscode -c "source /usr/local/share/nvm/nvm.sh && npm install -g <your-package-here>" 2>&1
 
 # Install xh
-RUN curl -sfL https://raw.githubusercontent.com/ducaale/xh/master/install.sh | XH_BINDIR=/usr/local/bin sh
\ No newline at end of file
+RUN curl -sfL https://raw.githubusercontent.com/ducaale/xh/master/install.sh | XH_BINDIR=/usr/local/bin sh
```
jgiannuzzi, line 24: nitpick: please remove the newline

.golangci.yml
jgiannuzzi, line 1: I am getting the following warnings, is this expected?

```
WARN [runner] The linter 'varcheck' is deprecated (since v1.49.0) due to: The owner seems to have abandoned the linter. Replaced by unused. 
WARN [runner] The linter 'golint' is deprecated (since v1.41.0) due to: The repository of the linter has been archived by the owner. Replaced by revive. 
WARN [runner] The linter 'interfacer' is deprecated (since v1.38.0) due to: The repository of the linter has been archived by the owner.  
WARN [runner] The linter 'maligned' is deprecated (since v1.38.0) due to: The repository of the linter has been archived by the owner. Replaced by govet 'fieldalignment'. 
```

jgiannuzzi (commented):
This is looking good! I just left a couple of comments below ⬇️

.devcontainer/Dockerfile
dsuhinin: sure! done.

.golangci.yml
dsuhinin: ahh, yes. this is my old config. let me try to find something more fresh =)

dsuhinin: ook, remove dead linters. right now we have:
```
    - lll
    - ineffassign
    - unparam
    - unused
    - unconvert
    - typecheck
    - megacheck
    - staticcheck
    - govet
    - goconst
    - gosimple
    - gosec
    - gocyclo
````
this is of course not a full list but good starting point. @jgiannuzzi please take a look when you will have some time and maybe you will find something that you want to include -> https://golangci-lint.run/usage/linters/ or I can do that also a bit later.

.devcontainer/Dockerfile
jgiannuzzi: Still there!

dsuhinin: emmmmmm

dsuhinin: ahhh, I found this guy. This guys is Goland IDE which add the new line =). done.

jgiannuzzi (approved):

PR #89, opened 2023-06-14
Applying fixes from other branch here as patch

### LINKED ISSUE #43: Replicate MLFlow behaviour during parameter logging
We differ from the MLFlow behaviour when logging parameters.

* when logging the same parameter with the same value, MLFLow treats it as valid and sends back `200`
* when logging the same parameter with a different value, MFLow treats it as `INVALID_PARAMETER_VALUE` and sends back `400`

We treat both cases as `INTERNAL_ERROR` and send back `500`, which has unfortunate side effects like https://github.com/optuna/optuna/issues/4722.

suprjinx:
Fixes #43

┆Issue is synchronized with this [Jira Task](https://gr-oss.atlassian.net/browse/FML-71) by [Unito](https://www.unito.io)


dsuhinin (approved):
approved

PR #110, opened 2023-06-21
Add integration tests for batch logging

dsuhinin:


pkg/api/mlflow/dao/models/metric.go
```diff
@@ -11,6 +11,11 @@ type Metric struct {
 	Iter      int64   `gorm:"index"`
 }
 
+// TableName returns actual table name.
+func (p Metric) TableName() string {
+	return "metrics"
+}
+
```
jgiannuzzi, line 18: I have seen in other PRs that we introduced these `TableName` methods. Out of curiosity, what is the reasoning behind them?

dsuhinin: ahh, yep. This is `tabler` interface and it needs to run this logic:
```
for _, table := range []schema.Tabler{
		models.Tag{},
		models.Param{},
		models.LatestMetric{},
		models.Metric{},
		models.Run{},
		models.ExperimentTag{},
		models.Experiment{},
	} {
		if err := f.db.Session(&gorm.Session{AllowGlobalUpdate: true}).Delete(table).Error; err != nil {
			return errors.Wrapf(err, "error deleting data from %s table", table.TableName())
		}
	}
```

pkg/api/mlflow/dao/repositories/metric.go
```diff
@@ -23,11 +23,15 @@ type MetricRepositoryProvider interface {
...
 	CreateBatch(ctx context.Context, run *models.Run, batchSize int, params []models.Metric) error
+	// GetMetricByKey returns metric by provided key.
+	GetMetricByKey(ctx context.Context, key string) (*models.Metric, error)
 	// GetMetricHistories returns metric histories by request parameters.
 	GetMetricHistories(
 		ctx context.Context,
 		experimentIDs []string, runIDs []string, metricKeys []string, viewType request.ViewType, limit int32,
 	) (*sql.Rows, func(*sql.Rows, interface{}) error, error)
+	// GetLatestMetricByKey returns the latest metric by provided key.
+	GetLatestMetricByKey(ctx context.Context, key string) (*models.LatestMetric, error)
```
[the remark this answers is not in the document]
dsuhinin: sorry, only one probably we need.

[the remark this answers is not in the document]
dsuhinin: just deleted `GetMetricByKey`

pkg/api/mlflow/dao/models/metric.go
jgiannuzzi: hmmm, I am pretty sure that there is a default naming method in Gorm that will give you the name instead

pkg/api/mlflow/dao/repositories/metric.go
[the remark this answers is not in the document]
jgiannuzzi: but is the other method only there for the tests and not for actual use in the service?

pkg/api/mlflow/dao/models/metric.go
dsuhinin: it is a pretty common practise. I can't see any problem to follow Gorm standards.

dsuhinin: gorm could generate it somewhere internally - yes, but I prefer to control it and use for logging as well like:
```
return errors.Wrapf(err, "error deleting data from %s table", table.TableName())
```

pkg/api/mlflow/dao/repositories/metric.go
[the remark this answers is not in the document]
dsuhinin: it is only for test. do you have any objections or concerns to do not have it?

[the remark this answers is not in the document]
dsuhinin: ok, removed it from `repo` now.

pkg/api/mlflow/dao/models/metric.go
jgiannuzzi: so this is only used for logging in tests — I would argue that we shouldn't define all of this code just for that, especially as the `err` returned by Gorm is already "no such table: runs", so simply having `return errors.Wrapf(err, "error unloading fixtures")` would give us all the information we need

dsuhinin: you are strong fighter =). ok let me check.

dsuhinin: done.

jgiannuzzi (approved):

PR #114, opened 2023-06-21
Ensure database is properly closed when shutting down the server

jgiannuzzi:


suprjinx (dismissed):

tests/integration/golang/mlflow/experiment/delete_test.go
```diff
@@ -40,6 +40,10 @@ func (s *DeleteExperimentTestSuite) SetupTest() {
 	s.fixtures = fixtures
 }
 
+func (s *DeleteExperimentTestSuite) TearDownTest() {
```
dsuhinin, line 43: this is not really important in tests.

dsuhinin: this is like with table names today =)

tests/integration/golang/mlflow/run/log_batch_test.go
```diff
@@ -65,6 +65,12 @@ func (s *LogBatchTestSuite) SetupTest() {
 	s.run = run
 }
 
+func (s *LogBatchTestSuite) TearDownTest() {
+	assert.Nil(s.T(), s.runFixtures.CloseDB())
+	assert.Nil(s.T(), s.metricFixtures.CloseDB())
+	assert.Nil(s.T(), s.experimentFixtures.CloseDB())
+}
+
```
dsuhinin, line 73: emmm, we definitely don't need to take case about that.

tests/integration/golang/mlflow/experiment/delete_test.go
jgiannuzzi: alright, I have removed all the test bits!

dsuhinin (approved):

PR #115, opened 2023-06-21
Updating website workflow

naskio:
In order to simplify the processing of Dependabot PRs, I suggest having the following behaviors based on the current branch:

1. Build and deploy if on the `main` branch.
2. Test if the website builds successfully if on a `dependabot` branch.


suprjinx (approved):

jgiannuzzi:
@naskio if the goal is this PR is to validate that any changes to the website files are not breaking the website itself, I would suggest building the website on pull requests and pushes to any branch, so that other kinds of changes can be validated too

.github/workflows/website.yml
```diff
@@ -1,9 +1,7 @@
-name: Deploy website to Pages
+name: Build and Deploy website to Pages
 
 on:
-  push: # Runs on pushes targeting the default branch
-    branches:
-      - main
+  push: # Runs on pushes targeting any branch
```
jgiannuzzi, line 4: we also need to check for pull requests — have a look at how it's done in `ci.yml`

naskio: Thanks, should be fixed now

naskio: I was just wondering if there are cases not covered by on `push` and covered by on `pull_request`, maybe when doing changes using GitHub editor ?

naskio: Ah forgot about forks xD never mind.

jgiannuzzi (changes requested):
one more thing: make sure we run only once for either PR or push to repo branch — again, have a look at how we do it in `ci.yml`!

naskio:
I tried to use `concurrency.group` instead of `if`, since it was already being used. Let me know if you prefer to do it otherwise.

jgiannuzzi (approved):
LGTM, thanks @naskio!

Issue #13, opened 2023-04-28
Support for string functions in Aim query parser


jgiannuzzi:
e.g. `startswith`, `endswith`, `contains`, ...

┆Issue is synchronized with this [Jira Task](https://gr-oss.atlassian.net/browse/FML-4) by [Unito](https://www.unito.io)

jgiannuzzi:
if the `in` syntax is too tricky, it's fine to split this into 2 PRs

dsuhinin:
ooook, lets start with `startswith, endswith, contains`

dsuhinin:
@jgiannuzzi can you please give me few examples how the end result should look like? I mean search query with `startswith`, `endswith`, `contains`?

PR #116, opened 2023-06-21
Require metric name in SearchMetrics

### LINKED ISSUE #45: Don't show all the metrics by default in the Aim UI, even when a search for some runs is provided
┆Issue is synchronized with this [Jira Task](https://gr-oss.atlassian.net/browse/FML-27) by [Unito](https://www.unito.io)

suprjinx:
Fixes #45


suprjinx:
image

Causes this error popup

jgiannuzzi:
I appreciate the screenshot, this is a nice touch and very helpful!

jgiannuzzi:
Although it does work in most cases, I think this may be a little bit brittle (e.g. `run.tags["mymetric.name"] == "something"` would be considered valid, even though it doesn't actually select a metric).

What about checking that we did indeed encounter a `metric.name` in the AST?

pkg/api/aim/runs.go
```diff
@@ -1039,3 +1039,8 @@ func toNumpy(values []float64) fiber.Map {
 		"blob":  buf.Bytes(),
 	}
 }
+
+// validateMetricNamePresent scans the query for metric.name condition
+func validateMetricNamePresent(query string) bool {
+	return strings.Contains(query, "metric.name == \"")
```
jgiannuzzi, line 1045: This won't work once we have proper string comparison support implemented! See #13

jgiannuzzi (changes requested):

pkg/api/aim/runs.go
suprjinx: updated to catch more cases, see test

pkg/api/aim/runs.go
```diff
@@ -1042,5 +1043,6 @@ func toNumpy(values []float64) fiber.Map {
 
 // validateMetricNamePresent scans the query for metric.name condition
 func validateMetricNamePresent(query string) bool {
-	return strings.Contains(query, "metric.name == \"")
+	re := regexp.MustCompile(`(in\s)?metric.name[\.|(\s?==)]+`)
```
jgiannuzzi, line 1046: The regex is a great idea! However I see a few issues with it:
* it won't match multiple whitespaces in `metric.name  == "accuracy"`
* it will match `metricsname`
* you use a character class (`[]`) so it will match all kinds of crazy things like `metric.name|`
* you force something to be behind `metric.name` so this won't match: `"accuracy" in metric.name`

I suggest the following regex instead — it may not be very elegant but I think it does the job!
```suggestion
	re := regexp.MustCompile(`in\s*metric\.name|metric\.name(?:\.|\s*==)`)
```

jgiannuzzi: also please make it a module var so it gets compiled only once!

jgiannuzzi (commented):
this is much much better, I love the regex idea!

pkg/api/aim/runs.go
suprjinx: made these changes!

jgiannuzzi (approved):

PR #117, opened 2023-06-21
Redact sensitive values in DSN

jgiannuzzi:


dsuhinin (approved):
hm, isn't possible to configure `glog`

PR #122, opened 2023-06-22
Support for string functions in Aim query parser

dsuhinin:


tests/integration/golang/aim/run/delete_test.go
```diff
@@ -76,22 +78,27 @@ func (s *DeleteRunTestSuite) Test_Ok() {
...
+			var resp fiber.Map
 			err = s.client.DoDeleteRequest(
 				fmt.Sprintf("/runs/%s", tt.request.RunID),
 				&resp,
 			)
 			assert.Nil(s.T(), err)
 
 			runs, err := s.runFixtures.GetTestRuns(context.Background(), s.runs[0].ExperimentID)
-			assert.NoError(s.T(), err)
+			assert.Nil(s.T(), err)
```
suprjinx, line 95: why do you change all of the NoError assertions to Nil?

tests/integration/golang/fixtures/run.go
```diff
@@ -74,11 +76,14 @@ func (f RunFixtures) CreateTestRuns(
 // GetTestRuns fetches all runs for an experiment
 func (f RunFixtures) GetTestRuns(
 	ctx context.Context, experimentID int32) ([]models.Run, error) {
-	runs := []models.Run{}
-	if err := f.db.WithContext(ctx).
-		Where("experiment_id = ?", experimentID).
-		Order("start_time desc").
-		Find(&runs).Error; err != nil {
+	var runs []models.Run
```
suprjinx, line 79: I think the original layout is much more readable. Are you following some particular style guide I should know about?

tests/integration/golang/aim/run/delete_test.go
dsuhinin: because under the hood, NoError do next stuff:
```
func NoError(t TestingT, err error, msgAndArgs ...interface{}) bool {
	if err != nil {
		if h, ok := t.(tHelper); ok {
			h.Helper()
		}
		return Fail(t, fmt.Sprintf("Received unexpected error:\n%+v", err), msgAndArgs...)
	}

	return true
}
```
is it is fully enough to just check for `NotNill`.

suprjinx: I think the assertion "NoError" is a better description of what we are asserting

docker/Dockerfile
```diff
@@ -31,4 +31,4 @@ RUN make go-build
 # Container settings.
 ENV LISTEN-ADDRESS 8080
 EXPOSE 8080
-ENTRYPOINT ["./fasttrack-service", "server", "--database-uri=postgres://postgres:postgres@fasttrack-postgres:5432/postgres", "--listen-address=:8080"]
+ENTRYPOINT ["./fasttrack-service", "server", "--database-uri=postgres://postgres:postgres@fasttrack-postgres:5432/postgres", "--listen-address=:8080", "--log-level=debug"]
```
dave-gantenbein, line 34: Do we want to leave this set at debug?

pkg/api/aim/encoding/decoder.go
```diff
@@ -0,0 +1,110 @@
...
+			return nil, eris.Wrap(err, "error reading data line")
+		}
+
+		var value any
+		switch valuebuf[0] {
+		case 0:
+			value = nil
+		case 1:
+			value = valuebuf[1] != 0
+		case 2:
```
dave-gantenbein, line 53: Could we use constants instead of magic numbers to make the intention of this switch clearer? ie: 

const INT = 2
const FLOAT = 3 // etc 

case INT16: // then do the length comparisons.

pkg/api/aim/encoding/decoder.go
jgiannuzzi, line 1: could you adapt the `decode` command to use this?

docker/Dockerfile
dsuhinin: it is pretty useful in general to see debug message and everything what is going on inside the service.

dsuhinin: so my vote - yes =)

pkg/api/aim/encoding/decoder.go
dsuhinin: I didn't touch the logic itself, because it isn't really in scope of this task. maybe reasonable to do that in separate task. what do you think?

pkg/api/aim/encoding/decoder.go
dsuhinin: can I do that in separate task? I didn't plan to change anything related to `decoder` in scope of this task.

jgiannuzzi: I think that if you want to move the code there, it would make sense to do that too rather than copying it over

docker/Dockerfile
suprjinx: this dockerfile only pertains to dev, yes?

pkg/api/aim/encoding/decoder.go
dsuhinin: ok. done.

pkg/api/aim/encoding/decoder.go
dsuhinin: adopted.

suprjinx (approved):
LGTM

tests/integration/golang/fixtures/run.go
dsuhinin: not really. I just started with this layout probably everywhere. We can switch to you layout - not a problem, but in that case we have to probably change in alllllll places which is time consuming process. that's why I decided to do vice versa right now.

tests/integration/golang/aim/run/delete_test.go
dsuhinin: probably the same answer. Not a big deal which to use. I've changed it just have the same approach everywhere. We can use `NoError` but in that case let's use it everywhere.

PR #128, opened 2023-06-22
Fix offset issue in Aim API GetExperimentRuns

### LINKED ISSUE #108: Aim API GetExperimentRuns does not respect the given offset

jgiannuzzi:
Fixes #108.

Since the refactor, we are now trying to update too many fields in multiple places (e.g. in `UpdateRun`). This is because we used to update only the required fields but we now pass around pointers to `Run` objects.
I chose to solve this by marking the fields as create+read in Gorm annotations.


suprjinx (approved):

dsuhinin:
> We could _and_ we should. This solution gets us there without having to refactor further, and it would remain useful as a 'seatbelt' even after refactoring further. What do you think?

oook. let's see.

dsuhinin (approved):

PR #133, opened 2023-06-23
Patch aim ui quickstart screen

### LINKED ISSUE #92: Patch Aim UI Quickstart Screen
When you first start the Aim UI and there are no data yet, a quick start guide is presented (see screenshot). This story involves adding a patch file and scripting that adapts these instructions for use with FasttrackML tracking server instead of Aim's backend. 

Because FasttrackML is API-compatible, we should be able to use the same Aim client usages with our tracking server. Perhaps use the current browser location to calculate the tracking server hostname?

image

galadd:
Fixes #92  made for a new file custom2.patch. I want to know i this is how it is going to look before I append it to custom.patch

Expectation: Aim uses fasttrack tracking server on the Aim UI Quickstart screen


pkg/ui/aim/embed/custom2.patch
```diff
@@ -0,0 +1,129 @@
...
+diff --git a/aim/web/ui/src/pages/Dashboard/components/QuickStart/QuickStart.tsx b/aim/web/ui/src/pages/Dashboard/components/QuickStart/QuickStart.tsx
+index 4608789..cec61aa 100644
+--- a/aim/web/ui/src/pages/Dashboard/components/QuickStart/QuickStart.tsx
++++ b/aim/web/ui/src/pages/Dashboard/components/QuickStart/QuickStart.tsx
+@@ -10,6 +10,9 @@ import { DOCUMENTATIONS } from 'config/references';
+ import './QuickStart.scss';
+ 
+ function QuickStart() {
++  // Get fasttrack server hostname from current browser location
++  const { hostname, port } = window.location;
```
suprjinx, line 116: great solution!

suprjinx:
@jgiannuzzi This looks great in the AIM ui, here's a pic below. I have question -- should the framework-specific help sections (the accordian of frameworks) receive any change of instructions? It seems like they are not referencing the tracking server, just local logging.

image

galadd:
I have observed in the shared screenshot that I overlooked an important detail: I forgot to convert the server into a string. However, I have rectified this issue in the latest commit.

suprjinx (dismissed):

jgiannuzzi (changes requested):
Thanks a lot for your PR @galadd! 🙏

I like how you're getting the tracking server information from the browser!

However, the instructions need to be changed to use MLFlow instead of Aim, as we do not support logging metrics with the Aim SDK (and we won't in the future either). Could you please adapt them with a simple equivalent script using MLFlow?

With regards to the accordion, we should adapt all of these examples to use MLFlow and our tracking server where possible. For the frameworks that would not be supported, we should simply remove them.

pkg/ui/aim/embed/custom.patch
```diff
@@ -638,3 +638,86 @@ index 0fc79c5..867376b 100644
...
+         </Text>
+         <CodeBlock
+-          code={`from aim import Run
++          code={`import tensorflow as tf
++import mlflow
++data = tf.keras.datasets.fashion_mnist
+ 
+-# Initialize a new run
+-run = Run()
++mlflow.set_tracking_uri("${fasttrack_server}")
```
suprjinx, line 693: definitely correct now -- using the mlflow client :-)

jgiannuzzi:
Superseded by #149

PR #140, opened 2023-06-26
Build with Goreleaser

### LINKED ISSUE #77: Add fml binary to releases
┆Issue is synchronized with this [Jira Task](https://gr-oss.atlassian.net/browse/FML-59) by [Unito](https://www.unito.io)

suprjinx:
Fixes #77


suprjinx:
image

Binary assets are uploaded to a release when created

.goreleaser.yaml
```diff
@@ -0,0 +1,38 @@
+project_name: fasttrackml
+builds:
+  - env:
+      - CGO_ENABLED=0
```
jgiannuzzi, line 4: I only started to look at your PR but I can see you disabled CGO here, which is required for FastTrackML to work — did you try launching any of the produced binaries?

suprjinx: i just assumed that goreleaser did the needful by targeting os and arch -- but I'll copy the build settings from root Dockerfile

jgiannuzzi:
do you have an example of this release workflow running somewhere?

suprjinx:
here's a run of the workflow: https://github.com/suprjinx/fasttrackml/actions/runs/5588264427/jobs/10214779343
here's a release, created in UI -- the assets were populated by the workflow: https://github.com/suprjinx/fasttrackml/releases/tag/v0.2.9

.goreleaser.yaml
```diff
@@ -0,0 +1,105 @@
...
+    binary: fasttrackml
+    goos:
+      - darwin
+    goarch:
+      - amd64
+    env:
+      - CC=o64-clang
+      - CXX=o64-clang++
+    flags:
+      - -mod=readonly
```
jgiannuzzi, line 16: those are default flags, do we really to set them?

.goreleaser.yaml
jgiannuzzi, line 1: we're missing `ldflags` on macOS — we want to remove symbols and set the version number!

.goreleaser.yaml
```diff
@@ -0,0 +1,105 @@
+project_name: fasttrackml
+env:
+  - CGO_ENABLED=1
+builds:
+  - id: darwin-amd64
+    main: ./
+    binary: fasttrackml
```
jgiannuzzi, line 7: binary should be `fml` 😄

.goreleaser.yaml
jgiannuzzi, line 1: could we please share most of the configuration between various platforms and architectures?
see https://goreleaser.com/customization/builds/#complex-template-environment-variables

.github/workflows/release.yml
```diff
@@ -1,103 +1,40 @@
...
-            fs.writeFileSync(`${process.env.GITHUB_WORKSPACE}/fasttrackml-oci-image.zip`, Buffer.from(download.data));
-
-      - name: Unzip artifact
-        run: unzip fasttrackml-oci-image.zip
-
-      - name: Login to Docker Hub
-        uses: docker/login-action@v2
+          docker run \
+          --rm \
+          -e CGO_ENABLED=1 \
```
jgiannuzzi: why is this environment variable set here and in the makefile instead of just in the goreleaser config?

.gitignore
```diff
@@ -21,7 +21,11 @@ mock_*
...
+ignore/
+
+# Goreleaser
+dist/
 
 # Optional workspace
 go.work
-*.code-workspace
\ No newline at end of file
+*.code-workspace
```
jgiannuzzi, line 31: please don't introduce whiteline changes, use `git add -p`

.gitignore
```diff
@@ -21,7 +21,11 @@ mock_*
 
 # Misc
 .DS_Store
+ignore/
```
jgiannuzzi, line 24: what's the rationale for ignoring this directory?

.goreleaser.yaml
```diff
@@ -0,0 +1,105 @@
...
+    goarch:
+      - amd64
+    env:
+      - CC=o64-clang
+      - CXX=o64-clang++
+    flags:
+      - -mod=readonly
+
+  - id: darwin-arm64
+    main: ./
```
jgiannuzzi, line 19: this is the default

.goreleaser.yaml
jgiannuzzi, line 1: build tags are missing!

jgiannuzzi (changes requested):

.goreleaser.yaml
suprjinx: i'll try to get it working but it was not happy with the original ldflags

.goreleaser.yaml
suprjinx: will do

.gitignore
suprjinx: that's a folder i use to put random stuff like patches, etc, so i don't have to see them in in my git view.

.goreleaser.yaml
suprjinx: build tags not happy for darwin, but I'll keep trying

jgiannuzzi: build tags work fine for me on darwin-amd64 and darwin-arm64 — are you not confusing them with the ldflags?

jgiannuzzi: essentially the static bits of the `ldflags` are for Linux only so that it works on any distro (`-linkmode external -extldflags -static`) — everything else should work on windows and darwin, where we don't require such tricks as the C libs are ABI compatible between releases of the OS.

suprjinx: I'm actually not having any success with linux builds and goreleaser-cross when I have those static/external flags

.gitignore
suprjinx: will do but this one won't go away

.goreleaser.yaml
jgiannuzzi: fixed in https://github.com/G-Research/fasttrackml/pull/140/commits/e2ad3a188a6494fa74f3f9ed1a85e38e75882441

.gitignore
jgiannuzzi: unrelated to this PR but useful

.gitignore
jgiannuzzi: fixed in https://github.com/G-Research/fasttrackml/pull/140/commits/3634b46ef265a3f2441f21eddb36d3453d995626

suprjinx:
not using

PR #146, opened 2023-07-02
Change 'FTML' to 'FastTrackML' in svg's

stackedsax:
Potentially the lowest-value contribution of all time: I noticed the title of these svg's was 'ftml'.  It feels like we're only using 'FastTrackML' and 'fml' in this project, so I'm adjusting the titles in these svg's to taste.


jgiannuzzi (approved):
Not giving any chance to 'FTML' — I like it! :+1:

PR #149, opened 2023-07-03
Fix Aim quickstart to use FastTrackML server

galadd:
This modifies only custom.patch file to include changes to aim quickstart ui. The example program used is the one presented in the repository readme. The tracking server uri is collected from the browser location

### DIFF
--- a/pkg/ui/aim/embed/custom.patch
+++ b/pkg/ui/aim/embed/custom.patch
@@ -638,3 +638,63 @@ index 0fc79c5..867376b 100644
      icon: 'audios',
      displayName: 'Audios',
      isExact: true,
+diff --git a/aim/web/ui/src/pages/Dashboard/components/QuickStart/QuickStart.tsx b/aim/web/ui/src/pages/Dashboard/components/QuickStart/QuickStart.tsx
+index 4608789..417b5bf 100644
+--- a/aim/web/ui/src/pages/Dashboard/components/QuickStart/QuickStart.tsx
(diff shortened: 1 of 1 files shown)


pkg/ui/aim/embed/custom.patch
```diff
@@ -638,3 +638,63 @@ index 0fc79c5..867376b 100644
...
+ 
+-# Log metrics
+-for i in range(10):
+-    run.track(i, name='loss', step=i, context={ "subset":"train" })
+-    run.track(i, name='acc', step=i, context={ "subset":"train" })`}
++model.compile(optimizer='adam',
++              loss='sparse_categorical_crossentropy',
++              metrics=['accuracy'])
++model.fit(training_images, training_labels, epochs=10)
++model.evaluate(test_images, test_labels)`}
```
jgiannuzzi, line 697: This should not depend on Tensorflow but rather but a one-to-one mapped example using the MLFlow API instead of the Aim one. e.g. it should create a run. log 2 parameters, and log 2 metrics 10 times.

galadd:
Hello @jgiannuzzi @suprjinx,

The way log metrics are handled in mlflow and Aim differs in terms of their input parameters. In mlflow's `log_metric` function, there is no `context` parameter like there is in Aim. On the other hand, Aim does not have a `value` parameter like MlFlow does. In MlFlow, the `value `parameter is used to specify a float value for the metric being logged. In this case, I have set the value to 0.00 + i.

jgiannuzzi:
Superseded by https://github.com/G-Research/fasttrackml-ui-aim/pull/7

PR #150, opened 2023-07-04
Add basic artifact support - S3 backend

dsuhinin:
This PR includes other related changes and improvements:
- `integration` tests for `/artifacts/list` endpoint.
- a bit changes logic to work with service configuration. now we have a configuration structure which we have to pass to other places were we want to something from config parameters.
- integrated `localstack` and `s3` service to the `integration` tests to properly test `/artifacts/list` endpoint.
- abstraction layer for other storages - not final and Im sure that it will be changed and adopted to other integrations.


pkg/api/mlflow/config/config.go
```diff
@@ -0,0 +1,35 @@
...
+	DatabaseReset         bool
+	DatabasePoolMax       int
+	DatabaseMigrate       bool
+	DatabaseSlowThreshold time.Duration
+}
+
+// NewServiceConfig creates new instance of ServiceConfig.
+func NewServiceConfig() *ServiceConfig {
+	return &ServiceConfig{
+		AuthUsername:          viper.GetString("auth-username"),
```
suprjinx, line 25: can't viper automatically populate a config struct that has annotations?

pkg/api/mlflow/service/artifact/service.go
```diff
@@ -6,29 +6,42 @@ import (
...
 	return &Service{
-		runRepository: runRepository,
+		runRepository:   runRepository,
+		artifactStorage: artifactStorage,
 	}
 }
 
-func (s Service) ListArtifacts(ctx context.Context, req *request.ListArtifactsRequest) error {
+// ListArtifacts handles business logic of `GET /artifacts/list` endpoint.
+func (s Service) ListArtifacts(
```
suprjinx, line 27: I kind of think the service layers should take more "neutral" argument types than api http request struct -- ie, controller should invoke service with string, int, etc, instead of passing on the http request -- this would aid in reuse across mflow and aim.

pkg/api/mlflow/service/artifact/storage/helpers.go
```diff
@@ -0,0 +1,11 @@
+package storage
+
+import "strings"
+
+// PrepareS3Prefix transforms input path to correct S3 prefix.
+// `runArtifactURI` represents path in next format -> s3://fasttrackml/2/30357ed2eaac4f2cacdbcd0e06e9e48a/artifacts.
+// to correctly pass it to S3, we have to remove `s3://fasttrackml/` prefix,
+// so final result has to be -> 2/30357ed2eaac4f2cacdbcd0e06e9e48a/artifacts.
+func PrepareS3Prefix(basePath, runArtifactURI string) string {
```
suprjinx, line 9: maybe this "ExtractsS3Path" instead of "PreparesS3Prefix"

pkg/api/mlflow/service/artifact/storage/nope.go
```diff
@@ -0,0 +1,14 @@
+package storage
+
+// Nope represents empty storage adapter.
+type Nope struct{}
+
+// NewNope creates new Nope instance.
+func NewNope() *Nope {
```
suprjinx, line 7: Maybe `Noop` (no op) or `Stub`?

pkg/api/mlflow/service/artifact/storage/storage.go
```diff
@@ -0,0 +1,52 @@
...
+	return o.IsDir
+}
+
+// Provider provides and interface to work with artifact storage.
+type Provider interface {
+	List(artifactURI, path, nextPageToken string) (string, string, []ArtifactObject, error)
+}
+
+// NewArtifactStorage creates
+func NewArtifactStorage(config *config.ServiceConfig) (Provider, error) {
```
suprjinx, line 39: nice :-)

pkg/api/mlflow/config/config.go
dsuhinin: this is a good question. I have to check.

pkg/api/mlflow/service/artifact/storage/helpers.go
dsuhinin: yeeep. this name is much better.

pkg/api/mlflow/service/artifact/storage/nope.go
dsuhinin: you are right. fixed!

pkg/api/mlflow/config/config.go
dsuhinin: ok, I did some sort of investigation. Viper supports reading configuration to `struct` -> https://github.com/spf13/viper#unmarshaling but a bit strange. I can't see that for `default` values for example. So I think I need more time to play with that so Im going to create a task and put it into `backlog` to investigate deeper a bit later.

dsuhinin: https://github.com/G-Research/fasttrackml/issues/154

.devcontainer/docker-compose.yml
```diff
@@ -4,6 +4,7 @@ volumes:
   home:
   workspaces:
   postgres-data:
+  minio-storage:
```
jgiannuzzi, line 7: let's call this `minio-data` for consistency

docker-compose.yml
```diff
@@ -5,16 +5,33 @@ services:
...
     depends_on:
+      - minio
       - postgres
     environment:
-      FML_DATABASE_URI: postgres://postgres:postgres@postgres/postgres
       FML_LOG_LEVEL: debug
+      FML_DATABASE_URI: postgres://postgres:postgres@postgres/postgres
+      FML_S3_ENDPOINT_URL: http://minio:9000
+      AWS_ACCESS_KEY_ID: "YFptmVrlVdOgnHZ8Idse"
+      AWS_SECRET_ACCESS_KEY: "ZeKsyUJikIAJQVtYsQNtdXx8qv9eevL94qFOx0Jh"
```
jgiannuzzi, line 15: why are you using different credentials here compared to the devcontainer?

.devcontainer/docker-compose.yml
dsuhinin: yep. done!

docker-compose.yml
```diff
@@ -5,16 +5,33 @@ services:
...
+    entrypoint: sh
+    command: -c 'mkdir -p /data/fasttrackml && minio server /data'
+    environment:
+      MINIO_ADDRESS: ":9000"
+      MINIO_ROOT_USER: "YFptmVrlVdOgnHZ8Idse"
+      MINIO_ROOT_PASSWORD: "ZeKsyUJikIAJQVtYsQNtdXx8qv9eevL94qFOx0Jh"
+      MINIO_CONSOLE_ADDRESS: ":9001"
+    ports:
+      - "9000:9000"
+      - "9001:9001"
```
jgiannuzzi, line 33: Exposing those ports would clash with the ones in the devcontainer and prevent us from running `make service-test` in there

docker-compose.yml
```diff
@@ -5,16 +5,33 @@ services:
...
     image: postgres:latest
     environment:
       - POSTGRES_PASSWORD=postgres
 
+  minio:
+    image: minio/minio:latest
+    entrypoint: sh
+    command: -c 'mkdir -p /data/fasttrackml && minio server /data'
+    environment:
+      MINIO_ADDRESS: ":9000"
```
jgiannuzzi, line 27: this is the default value

.devcontainer/docker-compose.yml
```diff
@@ -63,3 +64,15 @@ services:
...
+    image: minio/minio
+    ports:
+      - "9000:9000"
+      - "9001:9001"
+    volumes:
+      - minio-storage:/data
+    environment:
+      MINIO_ROOT_USER: user
+      MINIO_ROOT_PASSWORD: password
+    command: server --console-address ":9001" /data
```
jgiannuzzi, line 78: I think we should set this with `MINIO_CONSOLE_ADDRESS` for consistency

.devcontainer/docker-compose.yml
```diff
@@ -63,3 +64,15 @@ services:
...
+  minio:
+    image: minio/minio
+    ports:
+      - "9000:9000"
+      - "9001:9001"
+    volumes:
+      - minio-storage:/data
+    environment:
+      MINIO_ROOT_USER: user
+      MINIO_ROOT_PASSWORD: password
```
jgiannuzzi, line 77: could you please set these as `AWS_xxx` environment variables in the devcontainer too so that everything "just works"?

jgiannuzzi (commented):
A few comments about the tooling side of things — it looks really good though!

docker-compose.yml
dsuhinin: made them the same.

docker-compose.yml
dsuhinin: ok. removed.

.devcontainer/docker-compose.yml
dsuhinin: done.

.devcontainer/docker-compose.yml
dsuhinin: made them to be the same everywhere.

jgiannuzzi: I meant setting them in `.devcontainer/.env` 😃

docker-compose.yml
```diff
@@ -23,11 +36,15 @@ services:
       - go-cache:/go/pkg
     working_dir: /go/src
     depends_on:
-      - service
+      - minio
       - postgres
+      - service
     environment:
-      FML_DATABASE_URI: postgres://postgres:postgres@postgres/postgres
+      S3_BASE_URL: http://minio:9000
```
jgiannuzzi, line 43: could we use `FML_S3_ENDPOINT_URI` for both the app and the integration tests?

.devcontainer/docker-compose.yml
dsuhinin: aa, ok. done.

.devcontainer/.env
```diff
@@ -1,6 +1,13 @@
+# common service settings:
+LC_COLLATE=POSIX
```
jgiannuzzi, line 2: this one is a database setting!

jgiannuzzi: Also please capitalise the comments and choose either `:` or `.` (or nothing) so they're consistent

.devcontainer/devcontainer.json
jgiannuzzi, line 1: could you please add the `ghcr.io/christophermacgown/devcontainer-features/minio-client:1` feature so that the minio client is available within the devcontainer and can be used to create buckets?

jgiannuzzi: and then add a `postStartCommand` that will create the default bucket if it doesn't exist

jgiannuzzi (commented):
General comment: try using the devcontainer with vscode to validate your changes

docker-compose.yml
dsuhinin: ok. fixed.

.devcontainer/docker-compose.yml
```diff
@@ -63,3 +64,16 @@ services:
     # Add "forwardPorts": ["5432"] to **devcontainer.json** to forward PostgreSQL locally.
     # (Adding the "ports" property to this file will not forward from a Codespace.). Ports
     # section will work when running the compose file standalone.
+
+  minio:
+    image: minio/minio
+    ports:
+      - "9000:9000"
+      - "9001:9001"
```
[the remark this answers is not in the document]
dsuhinin: ahh, now I see what you mean. fixed and ports correctly exposed.

.devcontainer/.env
dsuhinin: fixed.

.devcontainer/devcontainer.json
dsuhinin: ooook. done! checked and it works. please take a look.

tests/integration/golang/mlflow/artifact/list_test.go
```diff
@@ -40,7 +40,7 @@ func TestListArtifactTestSuite(t *testing.T) {
 }
 
 func (s *ListArtifactTestSuite) SetupTest() {
-	s3Client, err := helpers.NewS3Client(os.Getenv("S3_BASE_URL"))
+	s3Client, err := helpers.NewS3Client(os.Getenv("FML_S3_ENDPOINT_URI"))
```
jgiannuzzi, line 43: I think we should use a helper which defaults to `http://localhost:9000` for consistency

pkg/cmd/server.go
```diff
@@ -188,6 +199,7 @@ func init() {
 
 	ServerCmd.Flags().StringP("listen-address", "a", "localhost:5000", "Address (host:post) to listen to")
 	ServerCmd.Flags().String("artifact-root", "s3://fasttrackml", "Artifact root")
+	ServerCmd.Flags().String("s3-endpoint-url", "", "S3 compatible storage base endpoint url")
```
jgiannuzzi, line 202:
```suggestion
	ServerCmd.Flags().String("s3-endpoint-uri", "", "S3 compatible storage base endpoint URI")
```

suprjinx (approved):

jgiannuzzi (changes requested):
I have left a couple more comments

pkg/api/mlflow/config/config.go
```diff
@@ -0,0 +1,35 @@
...
+		AuthPassword:          viper.GetString("auth-password"),
+		ArtifactRoot:          viper.GetString("artifact-root"),
+		S3EndpointURL:         viper.GetString("s3-endpoint-url"),
```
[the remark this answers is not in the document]
dsuhinin: fixed.

docker-compose.yml
```diff
@@ -5,16 +5,29 @@ services:
       args:
         tags: netgo osusergo sqlite_foreign_keys sqlite_math_functions sqlite_omit_load_extension sqlite_unlock_notify sqlite_vacuum_incr
     depends_on:
+      - minio
       - postgres
     environment:
-      FML_DATABASE_URI: postgres://postgres:postgres@postgres/postgres
       FML_LOG_LEVEL: debug
+      FML_DATABASE_URI: postgres://postgres:postgres@postgres/postgres
+      FML_S3_ENDPOINT_URL: http://minio:9000
```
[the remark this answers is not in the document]
dsuhinin: fixed.

.devcontainer/devcontainer.json
```diff
@@ -41,6 +43,8 @@
 		},
 		"ghcr.io/devcontainers/features/git:1": {},
 		"ghcr.io/devcontainers/features/github-cli:1": {},
-		"ghcr.io/dhoeric/features/k6:1": {}
-	}
+		"ghcr.io/dhoeric/features/k6:1": {},
+		"ghcr.io/christophermacgown/devcontainer-features/minio-client:1": {}
+	},
+	"postStartCommand": "mc config host add minio http://127.0.0.1:9000 user password && mc mb minio/fasttrackml"
```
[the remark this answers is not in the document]
dsuhinin: oook. fixed.

.devcontainer/.env
```diff
@@ -1,6 +1,13 @@
+# common service settings.
+FML_LOG_LEVEL=DEBUG
```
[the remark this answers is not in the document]
dsuhinin: fixed.

pkg/cmd/server.go
dsuhinin: fixed.

tests/integration/golang/mlflow/artifact/list_test.go
dsuhinin: fixed.

dsuhinin:
> The logic does not work when creating an experiment that points to a different bucket than the one specified in `--artifact-root`

dsuhinin:
> The logic does not work when creating an experiment that points to a different bucket than the one specified in `--artifact-root`

hmmm, you mean the case when we create experiment with not a default `artifact` path? if so, then probably I just totally forgot about this case. my bad. Ill extend logic without any problem.

pkg/api/mlflow/dao/convertors/experiment.go
```diff
@@ -33,7 +39,31 @@ func ConvertCreateExperimentToDBModel(req *request.CreateExperimentRequest) *mod
...
+			return nil, eris.Wrap(err, "error parsing artifact location")
+		}
+
+		switch u.Scheme {
+		case "s3":
+			artifactLocation := fmt.Sprintf("%s://%s", u.Scheme, strings.Trim(u.Host, "/"))
+			if path := strings.Trim(u.Path, "/"); path != "" {
+				artifactLocation = fmt.Sprintf("%s/%s", artifactLocation, path)
+			}
+			experiment.ArtifactLocation = artifactLocation
```
jgiannuzzi, line 54: why are we doing all of this trimming of slashes? if it's to be able to append experiment IDs and run IDs later, I would argue for using `url.JoinPath` instead

jgiannuzzi:
could you please resolve the conflicts with `main`?

dsuhinin:
> could you please resolve the conflicts with `main`?

done.

pkg/api/mlflow/dao/convertors/experiment.go
dsuhinin: this is because of different cases which I've tested:
- `s3://location/`
- s3://location/next`
- s3://location/next/
I can't see any reason to use `url.JoinPath` here.

dsuhinin: so we have to correctly trim latest slash. can't just use `url.JoinPath` and be happy. in that case the full url will be incorrect.

dsuhinin: probably I can use `url.JoinPath` but in any case I have to `trim` extra `slashes` that we don't need to be in path.

pkg/api/mlflow/api/request/artifact.go
```diff
@@ -2,15 +2,15 @@ package request
...
 	Path    string `query:"path"`
 	Token   string `query:"token"`
+	RunID   string `query:"run_id"`
+	RunUUID string `query:"run_uuid"`
 }
 
 // GetRunID returns Run ID.
 func (r ListArtifactsRequest) GetRunID() string {
-	if r.RunID == "" {
+	if r.RunID != "" {
```
suprjinx, line 13: good catch

suprjinx (dismissed):

pkg/api/mlflow/dao/convertors/experiment.go
jgiannuzzi: I mean `url.JoinPath` when you append the experiment IDs and run IDs later — but also why are you trimming the `Host` here? and why not just trim the slashes from the path and use `u.String()`?

pkg/api/mlflow/service/artifact/validators.go
```diff
@@ -10,5 +12,9 @@ func ValidateListArtifactsRequest(req *request.ListArtifactsRequest) error {
 	if req.RunID == "" && req.RunUUID == "" {
 		return api.NewInvalidParameterValueError("Missing value for required parameter 'run_id'")
 	}
+
+	if strings.Contains(req.Path, ".") || strings.Contains(req.Path, "..") {
```
jgiannuzzi, line 16: what about a path like `my.directory.containing.dots/whatever`?
I'd suggest using a regex and checking specifically for directories that have _only_ 1 or 2 dots

pkg/cmd/server.go
```diff
@@ -162,8 +162,14 @@ func initServer(config *mlflowConfig.ServiceConfig) *fiber.App {
...
-	if config.AuthUsername != "" && config.AuthPassword != "" {
-		log.Infof(`BasicAuth enabled with user "%s"`, config.AuthUsername)
+	if viper.GetBool("dev-mode") {
+		log.Info("Development mode - enabling CORS")
+		server.Use(cors.New())
+	}
+
+	authUsername := viper.GetString("auth-username")
+	authPassword := viper.GetString("auth-password")
+	if authUsername != "" && authPassword != "" {
```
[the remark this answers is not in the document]
dsuhinin: that's nice point.

pkg/cmd/server.go
```diff
@@ -162,8 +162,14 @@ func initServer(config *mlflowConfig.ServiceConfig) *fiber.App {
 		},
 	})
 
-	if config.AuthUsername != "" && config.AuthPassword != "" {
-		log.Infof(`BasicAuth enabled with user "%s"`, config.AuthUsername)
+	if viper.GetBool("dev-mode") {
+		log.Info("Development mode - enabling CORS")
+		server.Use(cors.New())
+	}
```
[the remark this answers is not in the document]
dsuhinin: that's nice point. Ill catch your new changes.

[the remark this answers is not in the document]
dsuhinin: done!

pkg/api/mlflow/service/artifact/validators.go
dsuhinin: I mean url.JoinPath when you append the experiment IDs and run IDs later — but also why are you trimming the Host here? and why not just trim the slashes from the path and use u.String()?
- hm, probably this is something old what you can see. let me a bit simplify this case.

dsuhinin: @jgiannuzzi about `url.JoinPath` . the problem and the point why I don't like to use it, because it returns `err` for such a simple operation, so it is really annoying to check it instead of to construct everything in just one line using `sprintf`.

dsuhinin:
@jgiannuzzi ooook, It seems like I've changed everything you mentioned. Im not fully sure that it looks like 100% you want to it should be near your requirements.

pkg/api/mlflow/service/artifact/validators.go
```diff
@@ -1,14 +1,32 @@
...
+
+	parsedUrl, err := url.Parse(req.Path)
+	if err != nil {
+		return api.NewInvalidParameterValueError("error parsing `path` parameter")
+	}
+	if parsedUrl.Scheme != "" || parsedUrl.Host != "" || parsedUrl.RawQuery != "" ||
+		parsedUrl.RawFragment != "" || parsedUrl.User != nil {
+		return api.NewInvalidParameterValueError("incorrect 'path' parameter has been provided")
+	}
+	if RelativePathRegExp.MatchString(parsedUrl.Path) {
```
jgiannuzzi, line 28: this path would pass validation: `..`

jgiannuzzi (commented):
Apart from the path validation bit, I think we're pretty much there — awesome work @dsuhinin!

pkg/api/mlflow/service/artifact/validators.go
```diff
@@ -9,7 +9,7 @@ import (
 )
 
 // RelativePathRegExp checks for the sequence `./` or `../` in provided path.
-var RelativePathRegExp = regexp.MustCompile(`\.{1,2}\/`)
+var RelativePathRegExp = regexp.MustCompile(`\.{2}\/?`)
```
jgiannuzzi, line 12: this is not allowing valid paths like: `some..directory/`
the _only_ thing we should disallow is "directories that are named `..` anywhere in the path"

jgiannuzzi (changes requested):

pkg/api/mlflow/service/artifact/validators.go
dsuhinin: fixed.

pkg/api/mlflow/service/artifact/validators.go
```diff
@@ -9,7 +9,7 @@ import (
 )
 
 // RelativePathRegExp checks for the sequence `./` or `../` in provided path.
```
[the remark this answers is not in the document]
dsuhinin: fixed

.devcontainer/.env
```diff
@@ -1,7 +1,17 @@
...
 POSTGRES_DB=postgres
 POSTGRES_HOSTNAME=localhost
-LC_COLLATE=POSIX
-FML_LOG_LEVEL=DEBUG
-FML_DEV_MODE=true
\ No newline at end of file
+
+# aws client settings.
+AWS_ACCESS_KEY_ID: "user"
+AWS_SECRET_ACCESS_KEY: "password"
```
jgiannuzzi, line 14: wrong syntax!
```suggestion
AWS_ACCESS_KEY_ID=user
AWS_SECRET_ACCESS_KEY=password
```

.devcontainer/.env
```diff
@@ -1,7 +1,17 @@
...
+
+# s3 compatible storage settings.
+FML_S3_ENDPOINT_URI=http://localhost:9000
```
jgiannuzzi, line 17: let's group all of these together as `s3 compatible storage settings.`

pkg/api/mlflow/service/artifact/validators.go
```diff
@@ -8,8 +8,8 @@ import (
 	"github.com/G-Research/fasttrackml/pkg/api/mlflow/api/request"
 )
 
-// RelativePathRegExp checks for the sequence `./` or `../` in provided path.
-var RelativePathRegExp = regexp.MustCompile(`\.{2}\/?`)
+// RelativePathRegExp checks for the `../` in provided path.
+var RelativePathRegExp = regexp.MustCompile(`\.{2}\/`)
```
jgiannuzzi, line 12: This is letting `..` go through (without `/` at the end). Please add a few unit tests like these:
- ok
  - `foo`
  - `./foo`
  - `./foo/`
  - `.foo`
  - `foo.bar`
  - `foo..bar`
- error
  - `..`
  - `foo/../bar`

.devcontainer/.env
dsuhinin: fixed.

pkg/api/mlflow/service/artifact/validators.go
```diff
@@ -9,7 +9,7 @@ import (
 )
 
 // RelativePathRegExp checks for the `../` in provided path.
-var RelativePathRegExp = regexp.MustCompile(`\.{2}\/`)
+var RelativePathRegExp = regexp.MustCompile(`\.{2}\/|^\.{2}$`)
```
jgiannuzzi, line 12: `foo..` will be considered invalid — I think you may want to just do a split and check for directories that are named exactly `..` instead of the regex

jgiannuzzi: never mind the above!

jgiannuzzi: `./..` would not be blocked by this, I think

dsuhinin: no no, foo.. it should work properly.

dsuhinin: yes yes. tests are already include such a case.

pkg/api/mlflow/service/artifact/validators.go
```diff
@@ -9,7 +9,7 @@ import (
 )
 
 // RelativePathRegExp checks for the `../` in provided path.
-var RelativePathRegExp = regexp.MustCompile(`\.{2}\/|^\.{2}$`)
+var RelativePathRegExp = regexp.MustCompile(`\/\.{2}|\.{2}\/|^\.{2}`)
```
jgiannuzzi, line 12: I think this is still not ok, as paths like `foo/..bar` or `foo../bar` would be considered invalid. I may be wrong, but I think that the regex is becoming really complex and it may be simpler to just check for any directories named exactly `..` in the path.

dsuhinin: no, this paths are fully valid. add them to unit tests.

jgiannuzzi: I'm very sorry Dmitriy, but the unit tests are wrong. You are setting the `RunID` instead of the `Path`!

dsuhinin: yes, Im a bit blinded already. now all good and tests still ok.

jgiannuzzi: sorry but the tests are still wrong — you are not passing the request to `ValidateListArtifactsRequest`!

dsuhinin: ohh, you are right. I have to disable my rush when you are online and answer so fast.

dsuhinin: then, yep. fixed and extend tests. looks good now.

dsuhinin: yep, you are right.  I have to stop rushing when you online and answer so fast. no should be all good. yes, probably `regexp` is not right(simplest) way to use it.

jgiannuzzi (approved):
LGTM 👍
Thanks a lot @dsuhinin!

PR #152, opened 2023-07-05
Various tooling improvements

jgiannuzzi:
* Build binary in Makefile with the right flags
* Fix Python integration tests in Makefile
* Fix gitignore for Python integration tests
* Refactor how integration tests are run
  - remove extra Dockerfiles and use the existing one instead
  - move docker-compose.yml to the root of the repository
  - cache building of the Docker image
  - do not clean the services when stopping them, but instead in a
    separate make target
  - run integration tests as a mounted volume instead of copying them
    into the container
  - introduce helpers to get the environment variables for the
    integration tests and provide defaults for them
  - add BuildX to CI
* Add Docker in Docker to devcontainer
* Rename source dir to fasttrackml within devcontainer
* Default to debug log level in devcontainer


tests/integration/golang/helpers/env.go
```diff
@@ -0,0 +1,19 @@
+package helpers
+
+import "os"
+
+func GetDatabaseUri() string {
+	uri, ok := os.LookupEnv("FML_DATABASE_URI")
+	if ok {
+		return uri
+	}
+	return "sqlite://fasttrackml.db"
```
suprjinx, line 10: nice!

suprjinx (approved):

tests/integration/golang/helpers/env.go
dsuhinin, line 10: fallback? probably much better to return an error or nothing(which could lead test just fail) so we always will see the problem. in case of fallback problem could be just hidden but w can try.

tests/integration/golang/aim/run/archive_batch_test.go
```diff
@@ -31,11 +30,11 @@ func TestArchiveBatchTestSuite(t *testing.T) {
 }
 
 func (s *ArchiveBatchTestSuite) SetupTest() {
-	s.client = helpers.NewAimApiClient(os.Getenv("SERVICE_BASE_URL"))
-	runFixtures, err := fixtures.NewRunFixtures(os.Getenv("DATABASE_DSN"))
+	s.client = helpers.NewAimApiClient(helpers.GetServiceUri())
```
dsuhinin, line 33: ooo, nice!.

docker-compose.yml
```diff
@@ -0,0 +1,33 @@
+services:
+  service:
```
dsuhinin, line 2: Im not sure about just a `service` instead of `prefix-service`. I've added `fastrack` prefix before to quickly see and find container just in case and also to avoid collisions with other running containers. what do you think?

dsuhinin (changes requested):
Few minor things.

docker-compose.yml
jgiannuzzi: Docker compose takes care of that for us. It always prefixes the container names with the project name (by default it's the directory containing `docker-compose.yml`).

dsuhinin: yeeeeep. it just looks ugly and you always have to find =((

tests/integration/golang/helpers/env.go
jgiannuzzi: I think it's a better experience when you have good defaults. In this particular case, it means that whether inside of outside of a devcontainer, you can just run `make run` in a terminal and `make test-go-integration` in another and it will just work. If you want to run against a different DB, you set the `FML_DATABASE_URI` to the same value in both terminals, and off you go!

Makefile
```diff
@@ -61,45 +52,58 @@ test-go-integration: ## run go integration tests.
 	go test -v -p 1 -tags="integration" ./tests/integration/golang/...
 
 PHONY: test-python-integration
```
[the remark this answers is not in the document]
jgiannuzzi: addressed in https://github.com/G-Research/fasttrackml/pull/152/commits/249ef2d714982db8f4f8da33acb52dcf8f1b98b6

Makefile
```diff
@@ -61,45 +52,58 @@ test-go-integration: ## run go integration tests.
...
 .PHONY: service-test
 service-test: service-stop service-start ## run tests over the service in docker.
 	@echo ">>> Running tests over service."
-	@docker-compose $(DOCKER_COMPOSE_OPTIONS) \
-		run fasttrack-integration-tests
+	@docker-compose \
+		run integration-tests
+
+.PHONY: service-clean
+service-clean:
```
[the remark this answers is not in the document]
jgiannuzzi: addressed in https://github.com/G-Research/fasttrackml/pull/152/commits/249ef2d714982db8f4f8da33acb52dcf8f1b98b6

tests/integration/golang/helpers/env.go
dsuhinin: ok. lets try.

dsuhinin (approved):

docker-compose.yml
jgiannuzzi: I disagree. Here is the output of `docker ps` on my machine, where the names are very obvious:
```
CONTAINER ID   IMAGE                              COMMAND                  CREATED          STATUS          PORTS                                                 NAMES
a2e6d53aa022   fasttrackml-service                "fml server"             11 seconds ago   Up 11 seconds                                                         fasttrackml-service-1
8a7d1fb20d3a   postgres:latest                    "docker-entrypoint.s…"   17 seconds ago   Up 17 seconds   5432/tcp                                              fasttrackml-postgres-1
```
(Scroll all the way to the left to see them)

dsuhinin: I mean when we have a prefix it is always a bit easier to see and group them.

dsuhinin: ok, no problem.

PR #153, opened 2023-07-05
Aim integration tests app

### LINKED ISSUE #57: Integration tests for Aim apps API
┆Issue is synchronized with this [Jira Task](https://gr-oss.atlassian.net/browse/FML-39) by [Unito](https://www.unito.io)

suprjinx:
Fixes #57

### DIFF
--- a/pkg/api/mlflow/dao/models/app.go
+++ b/pkg/api/mlflow/dao/models/app.go
@@ -15,8 +15,8 @@ type App struct {
 	CreatedAt  time.Time `json:"created_at"`
 	UpdatedAt  time.Time `json:"updated_at"`
 	IsArchived bool      `json:"-"`
-	Type  string   `gorm:"not null" json:"type"`
-	State AppState `json:"state"`
+	Type       string    `gorm:"not null" json:"type"`
+	State      AppState  `json:"state"`
 }
 
 type AppState map[string]any
--- a/pkg/api/mlflow/dao/models/app.go
+++ b/pkg/api/mlflow/dao/models/app.go
@@ -1,0 +1,42 @@
+package models
+
+import (
+	"database/sql"
+	"database/sql/driver"
+	"encoding/json"
+
+	"time"
+
+	"github.com/google/uuid"
+)
+
+type App struct {
+	ID         uuid.UUID `gorm:"type:uuid;primaryKey" json:"id"`
+	CreatedAt  time.Time `json:"created_at"`
+	UpdatedAt  time.Time `json:"updated_at"`
+	IsArchived bool      `json:"-"`
+	Type  string   `gorm:"not null" json:"type"`
+	State AppState `json:"state"`
+}
+
+type AppState map[string]any
+
+func (s AppState) Value() (driver.Value, error) {
+	v, err := json.Marshal(s)
+	if err != nil {
+		return nil, err
+	}
+	return string(v), nil
+}
+
+func (s *AppState) Scan(v interface{}) error {
+	var nullS sql.NullString
+	if err := nullS.Scan(v); err != nil {
+		return err
+	}
+	if nullS.Valid {
+		return json.Unmarshal([]byte(nullS.String), s)
+	}
+	s = nil
+	return nil
+}
--- a/tests/integration/golang/aim/app/create_app_test.go
+++ b/tests/integration/golang/aim/app/create_app_test.go
@@ -1,0 +1,99 @@
+package run
+
+import (
+	"context"
+	"os"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/suite"
+
+	"github.com/G-Research/fasttrackml/pkg/api/mlflow/dao/models"
+	"github.com/G-Research/fasttrackml/tests/integration/golang/fixtures"
+	"github.com/G-Research/fasttrackml/tests/integration/golang/helpers"
+)
+
+type CreateAppTestSuite struct {
+	suite.Suite
+	client      *helpers.HttpClient
+	appFixtures *fixtures.AppFixtures
+	apps        []*models.App
+}
+
+func TestCreateAppTestSuite(t *testing.T) {
+	suite.Run(t, new(CreateAppTestSuite))
+}
+
+func (s *CreateAppTestSuite) SetupTest() {
+	s.client = helpers.NewAimApiClient(os.Getenv("SERVICE_BASE_URL"))
+
+	appFixtures, err := fixtures.NewAppFixtures(os.Getenv("DATABASE_DSN"))
+	assert.Nil(s.T(), err)
+	s.appFixtures = appFixtures
+
+	s.apps, err = s.appFixtures.CreateTestApps(context.Background(), 10)
+	assert.Nil(s.T(), err)
+}
+
+func (s *CreateAppTestSuite) Test_Ok() {
+	defer func() {
+		assert.Nil(s.T(), s.appFixtures.UnloadFixtures())
+	}()
+	tests := []struct {
+		name        string
+		requestBody models.App
+	}{
+		{
+			name: "CreateValidApp",
+			requestBody: models.App{
+				Type: "app-type",
+				State: models.AppState{
+					"app-state-key": "app-state-value",
+				},
+			},
+		},
+	}
+	for _, tt := range tests {
+		s.T().Run(tt.name, func(T *testing.T) {
+
+			var resp models.App
+			err := s.client.DoPostRequest(
+				"/apps",
+				tt.requestBody,
+				&resp,
+			)
+			assert.Nil(s.T(), err)
+			assert.Equal(s.T(), "app-type", resp.Type)
+		})
+	}
+}
+
+func (s *CreateAppTestSuite) Test_Error() {
+	defer func() {
+		assert.Nil(s.T(), s.appFixtures.UnloadFixtures())
+	}()
+	tests := []struct {
+		name        string
+		requestBody any
+	}{
+		{
+			name: "CreateAppWithIncorrectBody",
+			requestBody: map[string]any{
+				"State": "this-cannot-unmarshal",
+			},
+		},
+	}
+	for _, tt := range tests {
+		s.T().Run(tt.name, func(T *testing.T) {
+
+			var resp map[string]any
+			err := s.client.DoPostRequest(
+				"/apps",
+				tt.requestBody,
+				&resp,
+			)
+			assert.Nil(s.T(), err)
+			assert.Contains(s.T(), resp["message"], "cannot unmarshal")
+		})
+	}
+}
--- a/tests/integration/golang/aim/app/delete_app_test.go
+++ b/tests/integration/golang/aim/app/delete_app_test.go
@@ -3,6 +3,7 @@ package run
 
 import (
 	"context"
+	"fmt"
 	"os"
 	"testing"
 
@@ -14,18 +15,18 @@ import (
 	"github.com/G-Research/fasttrackml/tests/integration/golang/helpers"
 )
 
-type CreateAppTestSuite struct {
+type DeleteAppTestSuite struct {
 	suite.Suite
 	client      *helpers.HttpClient
 	appFixtures *fixtures.AppFixtures
 	apps        []*models.App
 }
 
-func TestCreateAppTestSuite(t *testing.T) {
-	suite.Run(t, new(CreateAppTestSuite))
+func TestDeleteAppTestSuite(t *testing.T) {
+	suite.Run(t, new(DeleteAppTestSuite))
 }
 
-func (s *CreateAppTestSuite) SetupTest() {
+func (s *DeleteAppTestSuite) SetupTest() {
 	s.client = helpers.NewAimApiClient(os.Getenv("SERVICE_BASE_URL"))
 
 	appFixtures, err := fixtures.NewAppFixtures(os.Getenv("DATABASE_DSN"))
@@ -36,65 +37,78 @@ func (s *CreateAppTestSuite) SetupTest() {
 	assert.Nil(s.T(), err)
 }
 
-func (s *CreateAppTestSuite) Test_Ok() {
+func (s *DeleteAppTestSuite) Test_Ok() {
 	defer func() {
 		assert.Nil(s.T(), s.appFixtures.UnloadFixtures())
 	}()
 	tests := []struct {
 		name        string
-		requestBody models.App
+		wantAppCount int
 	}{
 		{
-			name: "CreateValidApp",
-			requestBody: models.App{
-				Type: "app-type",
-				State: models.AppState{
-					"app-state-key": "app-state-value",
-				},
-			},
+			name: "DeleteApp",
+			wantAppCount: 9,
 		},
 	}
 	for _, tt := range tests {
 		s.T().Run(tt.name, func(T *testing.T) {
 
-			var resp models.App
-			err := s.client.DoPostRequest(
-				"/apps",
-				tt.requestBody,
-				&resp,
+			var deleteResponse map[string]any
+			err := s.client.DoDeleteRequest(
+				fmt.Sprintf("/apps/%s", s.apps[0].ID),
+				&deleteResponse,
 			)
 			assert.Nil(s.T(), err)
-			assert.Equal(s.T(), "app-type", resp.Type)
+
+			var getResponse []models.App
+			err = s.client.DoGetRequest(
+				"/apps",
+				&getResponse,
+			)
+			assert.Nil(s.T(), err)
+			assert.Equal(s.T(), tt.wantAppCount, len(getResponse))
+
 		})
 	}
 }
 
-func (s *CreateAppTestSuite) Test_Error() {
+func (s *DeleteAppTestSuite) Test_Error() {
 	defer func() {
 		assert.Nil(s.T(), s.appFixtures.UnloadFixtures())
 	}()
 	tests := []struct {
 		name        string
 		requestBody any
+		wantAppCount int
 	}{
 		{
-			name: "CreateAppWithIncorrectBody",
+			name: "DeleteAppWithIncorrectBody",
 			requestBody: map[string]any{
 				"State": "this-cannot-unmarshal",
 			},
+			wantAppCount: 10,
 		},
 	}
 	for _, tt := range tests {
 		s.T().Run(tt.name, func(T *testing.T) {
 
-			var resp map[string]any
-			err := s.client.DoPostRequest(
-				"/apps",
+			var deleteResponse map[string]any
+			err := s.client.DoPutRequest(
+				fmt.Sprintf("/apps/%s", s.apps[0].ID),
 				tt.requestBody,
-				&resp,
+				&deleteResponse,
 			)
 			assert.Nil(s.T(), err)
-			assert.Contains(s.T(), resp["message"], "cannot unmarshal")
+			assert.Contains(s.T(), deleteResponse["message"], "cannot unmarshal")
+
+			var getResponse []models.App
+			err = s.client.DoGetRequest(
+				"/apps",
+				&getResponse,
+			)
+			assert.Nil(s.T(), err)
+			assert.Equal(s.T(), tt.wantAppCount, len(getResponse))
+
 		})
 	}
 }
--- a/tests/integration/golang/aim/app/get_app_test.go
+++ b/tests/integration/golang/aim/app/get_app_test.go
@@ -46,7 +46,7 @@ func (s *GetAppTestSuite) Test_Ok() {
 		expectedAppCount int
 	}{
 		{
-			name: "GetAppWithExistingRows",
+			name: "GetAppWithExistingID",
 		},
 	}
 	for _, tt := range tests {
--- a/tests/integration/golang/aim/app/get_app_test.go
+++ b/tests/integration/golang/aim/app/get_app_test.go
@@ -1,3 +1,4 @@
+//go:build integration
 package run
 
 import (
--- a/tests/integration/golang/aim/app/get_app_test.go
+++ b/tests/integration/golang/aim/app/get_app_test.go
@@ -2,9 +2,11 @@ package run
 
 import (
 	"context"
+	"fmt"
 	"os"
 	"testing"
 
+	"github.com/google/uuid"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/suite"
 
@@ -13,18 +15,18 @@ import (
 	"github.com/G-Research/fasttrackml/tests/integration/golang/helpers"
 )
 
-type GetAppsTestSuite struct {
+type GetAppTestSuite struct {
 	suite.Suite
-	client             *helpers.HttpClient
-	appFixtures        *fixtures.AppFixtures
-	apps               []*models.App
+	client      *helpers.HttpClient
+	appFixtures *fixtures.AppFixtures
+	apps        []*models.App
 }
 
-func TestGetAppsTestSuite(t *testing.T) {
-	suite.Run(t, new(GetAppsTestSuite))
+func TestGetAppTestSuite(t *testing.T) {
+	suite.Run(t, new(GetAppTestSuite))
 }
 
-func (s *GetAppsTestSuite) SetupTest() {
+func (s *GetAppTestSuite) SetupTest() {
 	s.client = helpers.NewAimApiClient(os.Getenv("SERVICE_BASE_URL"))
 
 	appFixtures, err := fixtures.NewAppFixtures(os.Getenv("DATABASE_DSN"))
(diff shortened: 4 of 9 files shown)


pkg/api/mlflow/dao/models/app.go
```diff
@@ -0,0 +1,42 @@
...
+	"database/sql"
+	"database/sql/driver"
+	"encoding/json"
+
+	"time"
+
+	"github.com/google/uuid"
+)
+
+type App struct {
```
dsuhinin, line 13: We try to avoid using shared objects like that, which contains `model` and `query` in the same object. can you please split them?

pkg/api/mlflow/dao/models/app.go
```diff
@@ -0,0 +1,42 @@
+package models
+
+import (
+	"database/sql"
```
dsuhinin, line 4: please organise imports in next way:
```
std packages
empty line
third part packages
empty line
internal pckages
```

pkg/api/mlflow/dao/models/app.go
```diff
@@ -0,0 +1,42 @@
...
+type App struct {
+	ID         uuid.UUID `gorm:"type:uuid;primaryKey" json:"id"`
+	CreatedAt  time.Time `json:"created_at"`
+	UpdatedAt  time.Time `json:"updated_at"`
+	IsArchived bool      `json:"-"`
+	Type       string    `gorm:"not null" json:"type"`
+	State      AppState  `json:"state"`
+}
+
+type AppState map[string]any
```
dsuhinin, line 22: probably if we add something will be nice to add comments everywhere.

tests/integration/golang/aim/app/create_app_test.go
```diff
@@ -0,0 +1,100 @@
...
+	defer func() {
+		assert.Nil(s.T(), s.appFixtures.UnloadFixtures())
+	}()
+	tests := []struct {
+		name        string
+		requestBody models.App
+	}{
+		{
+			name: "CreateValidApp",
+			requestBody: models.App{
```
dsuhinin, line 49: please split `models.App` to 2 objects: `request` and actual `model`.

tests/integration/golang/aim/app/create_app_test.go
```diff
@@ -0,0 +1,100 @@
...
+		s.T().Run(tt.name, func(T *testing.T) {
+
+			var resp models.App
+			err := s.client.DoPostRequest(
+				"/apps",
+				tt.requestBody,
+				&resp,
+			)
+			assert.Nil(s.T(), err)
+			assert.Equal(s.T(), "app-type", resp.Type)
```
dsuhinin, line 67: what about to check all other possible parameters? we probably have to check as much as possible.

tests/integration/golang/aim/app/create_app_test.go
```diff
@@ -0,0 +1,100 @@
...
+			name: "CreateAppWithIncorrectBody",
+			requestBody: map[string]any{
+				"State": "this-cannot-unmarshal",
+			},
+		},
+	}
+	for _, tt := range tests {
+		s.T().Run(tt.name, func(T *testing.T) {
+
+			var resp map[string]any
```
dsuhinin, line 90: maybe we can start using `response` objects for `AIM`? so we can use them and compare?

tests/integration/golang/aim/app/delete_app_test.go
```diff
@@ -0,0 +1,114 @@
...
+}
+
+func (s *DeleteAppTestSuite) SetupTest() {
+	s.client = helpers.NewAimApiClient(os.Getenv("SERVICE_BASE_URL"))
+
+	appFixtures, err := fixtures.NewAppFixtures(os.Getenv("DATABASE_DSN"))
+	assert.Nil(s.T(), err)
+	s.appFixtures = appFixtures
+
+	s.apps, err = s.appFixtures.CreateTestApps(context.Background(), 10)
```
dsuhinin, line 36: do we really need to create 10 apps? isn't enough to create just one app, then delete it and check that there is nothing in database right after that?

tests/integration/golang/aim/app/delete_app_test.go
```diff
@@ -0,0 +1,114 @@
...
+
+			var deleteResponse map[string]any
+			err := s.client.DoDeleteRequest(
+				fmt.Sprintf("/apps/%s", s.apps[0].ID),
+				&deleteResponse,
+			)
+			assert.Nil(s.T(), err)
+
+			var getResponse []models.App
+			err = s.client.DoGetRequest(
```
dsuhinin, line 64: in each `integration` test we try to be atomic and test only one particular endpoint. other calls has to go directly do database. can you please change it?

tests/integration/golang/aim/app/delete_app_test.go
```diff
@@ -0,0 +1,114 @@
...
+				"State": "this-cannot-unmarshal",
+			},
+			wantAppCount: 10,
+		},
+	}
+	for _, tt := range tests {
+		s.T().Run(tt.name, func(T *testing.T) {
+
+			var deleteResponse map[string]any
+			err := s.client.DoPutRequest(
```
dsuhinin, line 96: but this is not a `delete` request? or i misunderstood something?

dsuhinin: maybe we even don't need such a test in case of `delete`?

tests/integration/golang/aim/app/get_app_test.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+func (s *GetAppTestSuite) SetupTest() {
+	s.client = helpers.NewAimApiClient(os.Getenv("SERVICE_BASE_URL"))
+
+	appFixtures, err := fixtures.NewAppFixtures(os.Getenv("DATABASE_DSN"))
+	assert.Nil(s.T(), err)
+	s.appFixtures = appFixtures
+
+	s.apps, err = s.appFixtures.CreateTestApps(context.Background(), 10)
```
dsuhinin, line 37: we don't need to create 10 app here. this is a bit misleading. can you please create just one.

tests/integration/golang/aim/app/get_app_test.go
```diff
@@ -0,0 +1,88 @@
...
+	assert.Nil(s.T(), err)
+}
+
+func (s *GetAppTestSuite) Test_Ok() {
+	defer func() {
+		assert.Nil(s.T(), s.appFixtures.UnloadFixtures())
+	}()
+	tests := []struct {
+		name             string
+		expectedAppCount int
```
dsuhinin, line 47: this is unused, can you please remove it?

tests/integration/golang/aim/app/get_app_test.go
```diff
@@ -0,0 +1,88 @@
...
+				&resp,
+			)
+			assert.Nil(s.T(), err)
+			assert.Equal(s.T(), s.apps[0].ID, resp.ID)
+		})
+	}
+}
+
+func (s *GetAppTestSuite) Test_Error() {
+	assert.Nil(s.T(), s.appFixtures.UnloadFixtures())
```
dsuhinin, line 68: you probably even don't need to Unload here, because you call API with not existing `ID`.

tests/integration/golang/aim/app/delete_app_test.go
dsuhinin: it is a but confused to create 10 apps but we only need only 1.

tests/integration/golang/aim/app/get_app_test.go
```diff
@@ -0,0 +1,88 @@
...
+	for _, tt := range tests {
+		s.T().Run(tt.name, func(T *testing.T) {
+
+			var resp models.App
+			err := s.client.DoGetRequest(
+				fmt.Sprintf("/apps/%v", s.apps[0].ID),
+				&resp,
+			)
+			assert.Nil(s.T(), err)
+			assert.Equal(s.T(), s.apps[0].ID, resp.ID)
```
dsuhinin, line 62: will be nice to check as much as possible in each test. I mean to check not only `ID` but other props so we can make sure that endpoint works well and correctly after any change we do in the future.

tests/integration/golang/aim/app/get_apps_test.go
```diff
@@ -0,0 +1,91 @@
...
+		s.T().Run(tt.name, func(T *testing.T) {
+
+			var resp []models.App
+			err := s.client.DoGetRequest(
+				"/apps",
+				&resp,
+			)
+			assert.Nil(s.T(), err)
+			assert.Equal(s.T(), s.apps[0].ID, resp[0].ID)
+			assert.Equal(s.T(), tt.expectedAppCount, len(resp))
```
dsuhinin, line 63: will be nice to check not only the length but also each particular `app` object and all the fields. can you please change this moment?

tests/integration/golang/aim/app/get_apps_test.go
```diff
@@ -0,0 +1,91 @@
...
+		})
+	}
+}
+
+func (s *GetAppsTestSuite) Test_Error() {
+	assert.Nil(s.T(), s.appFixtures.UnloadFixtures())
+	tests := []struct {
```
dsuhinin, line 70: but this is not like a `error` tests. API return just an empty result which is fully fine and no error happened. can you please join this test with  `Test_Ok` if we really need it. Im not even sure that we have to test this case? What is the idea of it?

tests/integration/golang/aim/app/update_app_test.go
```diff
@@ -0,0 +1,101 @@
...
+}
+
+func (s *UpdateAppTestSuite) SetupTest() {
+	s.client = helpers.NewAimApiClient(os.Getenv("SERVICE_BASE_URL"))
+
+	appFixtures, err := fixtures.NewAppFixtures(os.Getenv("DATABASE_DSN"))
+	assert.Nil(s.T(), err)
+	s.appFixtures = appFixtures
+
+	s.apps, err = s.appFixtures.CreateTestApps(context.Background(), 10)
```
dsuhinin, line 36: can you please create just one particular app?

tests/integration/golang/aim/app/update_app_test.go
```diff
@@ -0,0 +1,101 @@
...
+		s.T().Run(tt.name, func(T *testing.T) {
+
+			var resp models.App
+			err := s.client.DoPutRequest(
+				fmt.Sprintf("/apps/%s", s.apps[0].ID),
+				tt.requestBody,
+				&resp,
+			)
+			assert.Nil(s.T(), err)
+			assert.Equal(s.T(), "app-type", resp.Type)
```
dsuhinin, line 68: Im not fully sure but can we check here that `State` also has been successfully updated?

tests/integration/golang/aim/app/update_app_test.go
```diff
@@ -0,0 +1,101 @@
...
+func (s *UpdateAppTestSuite) Test_Error() {
+	defer func() {
+		assert.Nil(s.T(), s.appFixtures.UnloadFixtures())
+	}()
+	tests := []struct {
+		name        string
+		requestBody any
+	}{
+		{
+			name: "UpdateAppWithIncorrectBody",
```
dsuhinin, line 82: can you please change test name from `UpdateAppWithIncorrectBody` -> `UpdateAppWithIncorrectState`, because only `State` incorrect here?

tests/integration/golang/fixtures/app.go
```diff
@@ -0,0 +1,80 @@
...
+		return nil, eris.Wrap(err, "error connection to database")
+	}
+	return &AppFixtures{
+		baseFixtures: baseFixtures{db: db.DB},
+		DbInstance:   db,
+	}, nil
+}
+
+// CreateTestApp creates a new test App.
+func (f AppFixtures) CreateTestApp(
```
dsuhinin, line 40: Im not sure that we need a `Test` prefix. Since the place where we use `fixtures` only `integration` tests we always know that we create a `test` records.

dsuhinin (approved):

pkg/api/mlflow/dao/models/app.go
jgiannuzzi: this will be enforced in https://github.com/G-Research/fasttrackml/pull/155 😄

pkg/api/mlflow/dao/models/app.go
suprjinx: will do

tests/integration/golang/aim/app/create_app_test.go
suprjinx: this will be the only one in the Aim side if I do that.

tests/integration/golang/aim/app/create_app_test.go
suprjinx: ok

tests/integration/golang/aim/app/create_app_test.go
suprjinx: Aim doesn't have any such types, so we'd be creating the first one here only to support a test.

tests/integration/golang/aim/app/delete_app_test.go
suprjinx: okay will reduce to 1

tests/integration/golang/aim/app/get_app_test.go
suprjinx: well the setup has created stuff

tests/integration/golang/aim/app/get_apps_test.go
suprjinx: yes I agree, very hard to have an error for this endpoint :-)

pkg/api/mlflow/dao/models/app.go
suprjinx: There's not a query in this file, just a scan. See also api/mlflow/dao/models/run.go, which also has scan logic.

I was trying to get us a step closer to having a shared "models" package for both aim and mlflow by putting here, but i can just delete this and use pkg/database model?

pkg/api/mlflow/dao/models/app.go
suprjinx: removing this file

pkg/api/mlflow/dao/models/app.go
suprjinx: removing file

tests/integration/golang/aim/app/create_app_test.go
suprjinx: the controller (handler func) itself is using adhoc request structs

tests/integration/golang/aim/app/get_apps_test.go
suprjinx: will do

tests/integration/golang/aim/app/get_apps_test.go
suprjinx: changed to "Test_Empty"

tests/integration/golang/aim/app/update_app_test.go
suprjinx: added assertion below

tests/integration/golang/aim/app/get_app_test.go
suprjinx: added assertions for all fields

tests/integration/golang/aim/app/get_apps_test.go
dsuhinin: can't we just organise this 2 cases in table test like we do in other cases?

dsuhinin (commented):
Left some comments.

tests/integration/golang/aim/app/delete_app_test.go
dsuhinin: the big problem of this approach which I faced before(in some previous projects), that if endpoint signature changed(endpoint which is in use in many places) you probably have to adjust many many tests(all the tests which use this endpoint). When you use database instead, of course something could also be changed but much more rarely and  usually you have to change the query which is under the hood of `methods`. In that case you probably have to do almost nothing to catch new changes. So that's why to have atomic test which tests exactly one `endpoint` instead of to have a test, which call other 2-3-4 endpoints to test actual endpoint.

tests/integration/golang/aim/app/get_apps_test.go
suprjinx: the main issue is that it's not convenient to have differing setup requirements in the same table test

tests/integration/golang/aim/app/get_app_test.go
```diff
@@ -0,0 +1,92 @@
...
+
+func (s *GetAppTestSuite) Test_Ok() {
+	defer func() {
+		assert.Nil(s.T(), s.appFixtures.UnloadFixtures())
+	}()
+	tests := []struct {
+		name string
+	}{
+		{
+			name: "GetAppWithExistingID",
```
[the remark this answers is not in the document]
suprjinx: i think it's prudent to setup for table test in all cases -- the occasional test which is not table driven is hard to read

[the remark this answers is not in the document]
dsuhinin: emmm, but why? this task for example -> https://github.com/G-Research/fasttrackml/blob/d1c362fca156b8069bfa61f1e3f43a54eae5782b/tests/integration/golang/mlflow/experiment/create_test.go#L37 looks pretty clean and understandable. I just want to say, that can we have setup for each test as small as possible? If we test only one case - just usual plain test. if we need to test only one entity - just create one entity instead of entities set. this is all adds purity to the test and understanding.

tests/integration/golang/aim/app/get_apps_test.go
dsuhinin: ok. maybe you are right. the moment which I'm not sure about - do we really need to test this case? @jgiannuzzi what do you think?

tests/integration/golang/aim/app/delete_app_test.go
```diff
@@ -0,0 +1,101 @@
...
+}
+
+func (s *DeleteAppTestSuite) Test_Error() {
+	defer func() {
+		assert.Nil(s.T(), s.appFixtures.UnloadFixtures())
+	}()
+	tests := []struct {
+		name         string
+		idParam      uuid.UUID
+		wantAppCount int
```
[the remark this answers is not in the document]
suprjinx: i think having a named param in the test struct is explaining what the magic number "0" means

tests/integration/golang/aim/app/get_app_test.go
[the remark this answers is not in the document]
suprjinx: Table-driven tests can explain what you're testing and looking for, whereas if there's not a table, "Test_Ok" and a collection of assertions are the only description which is not descriptive

tests/integration/golang/aim/app/update_app_test.go
```diff
@@ -0,0 +1,101 @@
...
+	for _, tt := range tests {
+		s.T().Run(tt.name, func(T *testing.T) {
+			var resp database.App
+			err := s.client.DoPutRequest(
+				fmt.Sprintf("/apps/%s", s.apps[0].ID),
+				tt.requestBody,
+				&resp,
+			)
+			assert.Nil(s.T(), err)
+			assert.Equal(s.T(), "app-type", resp.Type)
```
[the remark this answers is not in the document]
suprjinx: ok

tests/integration/golang/aim/app/delete_app_test.go
```diff
@@ -0,0 +1,101 @@
...
+	assert.Nil(s.T(), err)
+}
+
+func (s *DeleteAppTestSuite) Test_Ok() {
+	defer func() {
+		assert.Nil(s.T(), s.appFixtures.UnloadFixtures())
+	}()
+	tests := []struct {
+		name         string
+		wantAppCount int
```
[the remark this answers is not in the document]
suprjinx: "0" is magic number, "wantAppCount" describes it

tests/integration/golang/aim/app/delete_app_test.go
```diff
@@ -0,0 +1,101 @@
...
+	defer func() {
+		assert.Nil(s.T(), s.appFixtures.UnloadFixtures())
+	}()
+	tests := []struct {
+		name         string
+		idParam      uuid.UUID
+		wantAppCount int
+	}{
+		{
+			name:         "DeleteAppWithIncorrectID",
```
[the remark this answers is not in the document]
suprjinx: will do

tests/integration/golang/aim/app/get_app_test.go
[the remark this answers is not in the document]
suprjinx: but I will make the change

tests/integration/golang/aim/app/get_app_test.go
```diff
@@ -0,0 +1,92 @@
...
+}
+
+func (s *GetAppTestSuite) Test_Error() {
+	assert.Nil(s.T(), s.appFixtures.UnloadFixtures())
+	tests := []struct {
+		name    string
+		idParam uuid.UUID
+	}{
+		{
+			name:    "GetAppWithUnknownID",
```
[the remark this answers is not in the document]
suprjinx: kk

tests/integration/golang/aim/app/delete_app_test.go
```diff
@@ -0,0 +1,101 @@
...
+}
+
+func (s *DeleteAppTestSuite) SetupTest() {
+	s.client = helpers.NewAimApiClient(os.Getenv("SERVICE_BASE_URL"))
+
+	appFixtures, err := fixtures.NewAppFixtures(os.Getenv("DATABASE_DSN"))
+	assert.Nil(s.T(), err)
+	s.appFixtures = appFixtures
+
+	s.apps, err = s.appFixtures.CreateApps(context.Background(), 1)
```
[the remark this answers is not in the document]
suprjinx: CreateApp requests an input Application, which is more cumbersome -- some boilerplate is created with this function.

tests/integration/golang/aim/app/get_apps_test.go
```diff
@@ -0,0 +1,74 @@
...
+			apps, err := s.appFixtures.CreateApps(context.Background(), tt.expectedAppCount)
+			assert.Nil(s.T(), err)
+
+			var resp []database.App
+			err = s.client.DoGetRequest(
+				"/apps",
+				&resp,
+			)
+			assert.Nil(s.T(), err)
+			for idx := 0; idx < tt.expectedAppCount; idx++ {
```
dsuhinin, line 64: sorry, but this test still look strange:
- Im still no sure about `GetAppsWithNoRows`. what is the reason to test this moment? Then we probably can test for example `GetAppsWithOneRow` or `GetAppsWithTwoRows`?
- you iterate through the expected count but inside the cycle compare always with `apps[0]`. it isn't really clear, why? So when you read this test it is unclear really.

dsuhinin: also this test is not fail as I can see only, because you in second case just use `expectedAppCount` which is equal `0` and cycle won't just work in that case, but data in any case will be in database.

suprjinx: okay i'll take another look

tests/integration/golang/aim/app/update_app_test.go
```diff
@@ -0,0 +1,101 @@
...
+}
+
+func (s *UpdateAppTestSuite) SetupTest() {
+	s.client = helpers.NewAimApiClient(os.Getenv("SERVICE_BASE_URL"))
+
+	appFixtures, err := fixtures.NewAppFixtures(os.Getenv("DATABASE_DSN"))
+	assert.Nil(s.T(), err)
+	s.appFixtures = appFixtures
+
+	s.apps, err = s.appFixtures.CreateApps(context.Background(), 1)
```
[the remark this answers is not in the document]
suprjinx: i'll change suite field to singular app and capture apps[0] there.

tests/integration/golang/aim/app/update_app_test.go
```diff
@@ -0,0 +1,101 @@
...
+	assert.Nil(s.T(), err)
+}
+
+func (s *UpdateAppTestSuite) Test_Ok() {
+	defer func() {
+		assert.Nil(s.T(), s.appFixtures.UnloadFixtures())
+	}()
+	tests := []struct {
+		name        string
+		requestBody map[string]any
```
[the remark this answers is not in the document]
suprjinx: doing this

tests/integration/golang/aim/app/update_app_test.go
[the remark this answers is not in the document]
suprjinx: so i think this should stay as is -- use the structs from the db for comparison, except where they are out of date and then use literal.

tests/integration/golang/aim/app/create_app_test.go
```diff
@@ -0,0 +1,92 @@
...
+}
+
+func (s *CreateAppTestSuite) Test_Error() {
+	tests := []struct {
+		name        string
+		requestBody any
+	}{
+		{
+			name: "CreateAppWithIncorrectJson",
+			requestBody: map[string]any{
```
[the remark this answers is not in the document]
suprjinx: in other cases, introduced `request` and `response` objects

tests/integration/golang/aim/app/delete_app_test.go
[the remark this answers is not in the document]
suprjinx: but did change suite attribute to singular "app"

dsuhinin (approved):

PR #159, opened 2023-07-07
Add basic CORS support via --dev-mode hidden option

jgiannuzzi:
This enables basic CORS support in the server, which is required to get `npm start` to work


pkg/cmd/server.go
```diff
@@ -190,6 +197,7 @@ func init() {
 	ServerCmd.Flags().String("artifact-root", "s3://fasttrackml", "Artifact root")
 	ServerCmd.Flags().String("auth-username", "", "BasicAuth username")
 	ServerCmd.Flags().String("auth-password", "", "BasicAuth password")
+	ServerCmd.Flags().String("cors-allow-origins", "*", "CORS allowed origins")
```
suprjinx, line 200: should this default to all hosts?

jgiannuzzi: good question — here is my current thinking:
* figuring out how the server will actually be accessed is not trivial, especially if it's running behind a k8s ingress
* users will usually just run it locally and without any options
* for users that actually care about security, they will now have the option to set the allowed origins

happy to debate it, as I'm not entirely sure about this choice either!

jgiannuzzi: actually I'm very wrong about this!
we should instead default to no CORS headers at all when no origin is specified, so that it only works from the same browser origin, and then probably set `*` as the default in the devcontainer

dsuhinin: what about other CORS options?

jgiannuzzi: I have reduced the whole thing to a new hidden `--dev-mode` flag which enables CORS with default options so that `npm start` can be used to develop the UI and the backend together.

dsuhinin (approved):
Just a common question - don't we need to describe all of this parameters in README.md?

PR #171, opened 2023-07-11
Add ability to use regular expressions in Aim search queries

dsuhinin:
Implement `match` function which is analog of `python` match:
- modified `query` object to keep current `dialector`.
- use `github.com/multiprocessio/go-sqlite3-stdlib` - it represents big set of extensions, because natively `sqllite` doesn't support `regexp`.  Probably `sqlite3-pcre` is something that could help but in case of `local` development question is still opened? Also there is - https://github.com/nalgeon/sqlean but let's discuss it.
- modified `unit` tests to test different cases for different `dialectors`.
- not fully sure how to apply the same tests for `integration` tests. Maybe we need some sort of `matrix` to run service with different configurations: `postgres`, `sqlite`, `s3`, etc.

P.S Im not sure that this is final solution but I've spent a lot of time and seems like there is no normal way to do that for `sqlite`.


jgiannuzzi:
With regards to regexp support in SQLite, could we please benchmark a C regexp implementation vs a Go one?
Also, if we use a Go one, it would be enough to simply register a copy of this [trivial function](https://github.com/multiprocessio/go-sqlite3-stdlib/blob/6ac92baf2e8afd2c95e333d939edb744eea9ea5e/regexp.go#L7-L14) within a ConnectHook (example [here](https://github.com/multiprocessio/go-sqlite3-stdlib/blob/6ac92baf2e8afd2c95e333d939edb744eea9ea5e/stdlib.go#L18-L42)). I'm not very keen in overriding so many SQLite builtin functions just for this.

jgiannuzzi:
here is a [complete working example](https://github.com/mattn/go-sqlite3/blob/64bb93539113de32d6ae3daef1e09ca8239c743d/_example/custom_func/main.go) of registering a function written in Go with SQLite

dsuhinin:
> here is a [complete working example](https://github.com/mattn/go-sqlite3/blob/64bb93539113de32d6ae3daef1e09ca8239c743d/_example/custom_func/main.go) of registering a function written in Go with SQLite

yeesss. I also checked this stuff but it was a problem. When I tried to use `regexp` in `WHERE` statement, gorm always says that function is unregistered, that's why I just continued to move on. Ok. let me dive again.

jgiannuzzi:
another thing to benchmark is whether having a LRU cache of compiled regexps in the custom Go function would make a difference — my guess is that it does and so we should implement it that way

dsuhinin:
> here is a [complete working example](https://github.com/mattn/go-sqlite3/blob/64bb93539113de32d6ae3daef1e09ca8239c743d/_example/custom_func/main.go) of registering a function written in Go with SQLite

ahh, ok. now I remembered what is the problem with this solution. so when you registered `regexp` function such way, other functions that we use just stop working. probably when you we register like that -> https://github.com/mattn/go-sqlite3/blob/64bb93539113de32d6ae3daef1e09ca8239c743d/_example/custom_func/main.go it is no different with to use -> github.com/multiprocessio/go-sqlite3-stdlib. as a result when I registered new `driver`, other functions just stop working like for example `mod` and I think encryption also should stop working.

jgiannuzzi:
I just tested it and it works fine with both encryption and MOD — please implement it this way!

pkg/database/database.go
```diff
@@ -76,11 +81,27 @@ func ConnectDB(
...
+					c, err := regexp.Compile(re)
+					if err != nil {
+						return false
+					}
+
+					return c.MatchString(s)
+				}, true); err != nil {
+					return err
+				}
+				return nil
```
jgiannuzzi, line 96: you could return the error from `conn.RegisterFunc` directly

pkg/database/database.go
```diff
@@ -76,11 +81,27 @@ func ConnectDB(
...
+			},
+		})
+
+		s, err := sql.Open(SQLiteCustomDriverName, strings.Replace(dbURL.String(), "sqlite://", "file:", 1))
 		if err != nil {
 			return nil, fmt.Errorf("failed to connect to database: %w", err)
 		}
+
```
jgiannuzzi, line 104: nitpick: please don't introduce a newline here

pkg/database/database.go
```diff
@@ -8,9 +8,11 @@ import (
 	glog "log"
 	"net/url"
 	"os"
+	"regexp"
 	"strings"
 	"time"
 
+	goSqlite3 "github.com/mattn/go-sqlite3"
```
jgiannuzzi, line 15: the default name of `sqlite3` could be used instead

pkg/api/aim/query/query.go
```diff
@@ -217,6 +222,34 @@ func (pq *parsedQuery) parseAttribute(node *ast.Attribute) (any, error) {
...
+
+				arg, ok := args[0].(*ast.Str)
+				if !ok {
+					return nil, errors.New("unsupported argument type. has to be `string` only")
+				}
+
+				switch pq.qp.Dialector {
+				case sqlite.Dialector{}.Name():
+					return clause.Expr{
+						SQL: fmt.Sprintf(`"%s"."%s" regexp '%s'`, c.Table, c.Name, arg.S),
```
jgiannuzzi, line 243: we should use columns and vars in a new `clause.Expression` type — have a look at `clause.Like` for example

pkg/api/aim/query/query_test.go
```diff
@@ -32,37 +35,119 @@ func (s *QueryTestSuite) SetupTest() {
...
 		name         string
 		query        string
+		dialector    string
 		expectedSQL  string
 		expectedVars []interface{}
 	}{
 		{
 			name:         "TestRunNameWithoutFunction",
 			query:        `(run.name == 'run')`,
+			dialector:    postgres.Dialector{}.Name(),
```
jgiannuzzi, line 119: this should be the SQLite dialector everywhere in this test function

jgiannuzzi (changes requested):

pkg/api/aim/query/query.go
```diff
@@ -217,6 +222,37 @@ func (pq *parsedQuery) parseAttribute(node *ast.Attribute) (any, error) {
...
+				case postgres.Dialector{}.Name():
+					regexpOperator = "~"
+				default:
+					return nil, fmt.Errorf("unsupported dialector type: %s", pq.qp.Dialector)
+				}
+				return clause.Expr{
+					SQL: fmt.Sprintf(`"%s"."%s" %s ?`, c.Table, c.Name, regexpOperator),
+					Vars: []interface{}{
+						fmt.Sprintf("'%s'", arg.S),
+					},
```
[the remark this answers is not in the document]
dsuhinin: but `clause.Expr` is what implements `clause.Expression`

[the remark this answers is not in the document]
jgiannuzzi: sure, but we should implement our `clause.Regexp` expression that properly quotes columns instead of using `Sprintf`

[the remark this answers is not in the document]
jgiannuzzi: have a look at the code for `clause.Like` that I linked above

pkg/database/database.go
dsuhinin: done.

pkg/database/database.go
dsuhinin: done.

pkg/database/database.go
```diff
@@ -194,7 +217,7 @@ func checkAndMigrateDB(db *DbInstance, migrate bool) error {
 		tx.First(&schemaVersion)
 	}
 
-	if alembicVersion.Version != "97727af70f4d" || schemaVersion.Version != "5d042539be4f" {
+	if alembicVersion.Version != "97727af70f4d" || schemaVersion.Version != "1ce8669664d2" {
```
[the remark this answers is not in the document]
dsuhinin: done.

pkg/api/aim/query/query.go
```diff
@@ -217,6 +222,34 @@ func (pq *parsedQuery) parseAttribute(node *ast.Attribute) (any, error) {
...
+
+				switch pq.qp.Dialector {
+				case sqlite.Dialector{}.Name():
+					return clause.Expr{
+						SQL: fmt.Sprintf(`"%s"."%s" regexp '%s'`, c.Table, c.Name, arg.S),
+					}, nil
+				case postgres.Dialector{}.Name():
+					return clause.Expr{
+						SQL: fmt.Sprintf(`"%s"."%s" ~ '%s'`, c.Table, c.Name, arg.S),
+					}, nil
```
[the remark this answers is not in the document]

pkg/api/aim/query/query_test.go
dsuhinin: done.

pkg/api/aim/query/query.go
```diff
@@ -264,8 +267,9 @@ func (pq *parsedQuery) parseAttribute(node *ast.Attribute) (any, error) {
 				if !ok {
 					return nil, errors.New("unsupported argument type. has to be `string` only")
 				}
-				return clause.Expr{
-					SQL: fmt.Sprintf(`"%s"."%s" LIKE '%%%s'`, c.Table, c.Name, arg.S),
+				return clause.Like{
+					Value:  fmt.Sprintf(`'%%%s'`, arg.S),
+					Column: fmt.Sprintf(`"%s"."%s"`, c.Table, c.Name),
```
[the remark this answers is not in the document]

pkg/api/aim/query/query.go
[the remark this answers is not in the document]
jgiannuzzi: yes, exactly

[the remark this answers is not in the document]
jgiannuzzi: also, I have realised we implemented `str.contains` in #122, but this is not valid Python — the way to do it in Python is `"me" in "something"` or `"substring" in run.name` in our case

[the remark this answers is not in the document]
jgiannuzzi: and `re.match("pattern", "string")` is equivalent to `re.search("^pattern", "string")`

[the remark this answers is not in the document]
dsuhinin: emmm, but our syntax is `entity.column.function`?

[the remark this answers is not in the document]
dsuhinin: `also, I have realised we implemented str.contains in https://github.com/G-Research/fasttrackml/pull/122, but this is not valid Python — the way to do it in Python is "me" in "something" or "substring" in run.name in our case` - thats correct, but isn't what we have right now in the same style with `startswith`, `endswith`?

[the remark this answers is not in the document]
dsuhinin: ok, added custom `clause` implementation. tested everything and it seems like it looks good now.

pkg/database/database.go
dsuhinin: done

jgiannuzzi:
here are a few examples of queries using regexes that should work:

---

```python
re.match("(?:test|train)", run.name)
```
=>
```sql
SELECT * FROM "runs" WHERE "runs"."name" REGEXP '^(?:test|train)';
```

---

```python
re.search("(?:test|train)", run.name)
```
=>
```sql
SELECT * FROM "runs" WHERE "runs"."name" REGEXP '(?:test|train)';
```

---

```python
re.search("(?:test|train)", run.mode)
```
=>
```sql
SELECT * FROM "runs" LEFT JOIN params params_0 ON runs.run_uuid = params_0.run_uuid AND params_0.key = 'mode' WHERE "params_0"."value" REGEXP '(?:test|train)';
```

dsuhinin:
> here are a few examples of queries using regexes that should work:
> 
> ```python
> re.match("(?:test|train)", run.name)
> ```
> 
> =>
> 
> ```sql
> SELECT * FROM "runs" WHERE "runs"."name" REGEXP '^(?:test|train)';
> ```
> 
>  ```python
> re.search("(?:test|train)", run.name)
> ```
> 
> =>
> 
> ```sql
> SELECT * FROM "runs" WHERE "runs"."name" REGEXP '(?:test|train)';
> ```
> 
>  ```python
> re.search("(?:test|train)", run.mode)
> ```
> 
> =>
> 
> ```sql
> SELECT * FROM "runs" LEFT JOIN params params_0 ON runs.run_uuid = params_0.run_uuid AND params_0.key = 'mode' WHERE "params_0"."value" REGEXP '(?:test|train)';
> ```

ok, thank you! Ill take a look how we can handle that.

dsuhinin:
@jgiannuzzi oook, Im a bit refactored current functionality to support `match` and `search` functions to be like in `python`. As usually, the solution is not simple but it works and fully tested by `unit` and `integration` tests. please let me know what you think?

pkg/api/aim/query/query_test.go
```diff
@@ -32,36 +34,132 @@ func (s *QueryTestSuite) SetupTest() {
...
+			assert.Equal(s.T(), tt.expectedVars, result.Statement.Vars)
+		})
+	}
+}
+
+func (s *QueryTestSuite) TestSqliteDialector_Ok() {
+	tests := []struct {
+		name         string
+		query        string
+		dialector    string
```
suprjinx, line 118: since this test is entirely for sqlitedialector, this could be defaulted

pkg/api/aim/query/query_test.go
```diff
@@ -32,36 +34,132 @@ func (s *QueryTestSuite) SetupTest() {
 	s.db = db
 }
 
-func (s *QueryTestSuite) Test_Ok() {
+func (s *QueryTestSuite) TestPostgresDialector_Ok() {
 	tests := []struct {
 		name         string
 		query        string
+		dialector    string
```
suprjinx, line 41: Since this test is entirely posgres dialector, this parameter could be defaulted

suprjinx (approved):
LGTM

pkg/api/aim/query/query_test.go
dsuhinin: yep, removed.

pkg/api/aim/query/query_test.go
dsuhinin: yep, removed.

pkg/api/aim/query/query.go
```diff
@@ -581,6 +612,57 @@ func (pq *parsedQuery) parseName(node *ast.Name) (any, error) {
...
+						regexp, ok := args[1].(*ast.Str)
+						if !ok {
+							return nil, errors.New("first argument type for re.match function has to be ast.Str")
+						}
+						attribute, ok := args[2].(*ast.Attribute)
+						if !ok {
+							return nil, errors.New("second argument type for re.match function has to be ast.Attribute")
+						}
+
+						attributes, err := pq.parseName(attribute.Value.(*ast.Name))
```
jgiannuzzi, line 640: you should never call `parseName` directly but instead use `parseNode` which will help with proper error handling

pkg/api/aim/query/query.go
```diff
@@ -581,6 +612,57 @@ func (pq *parsedQuery) parseName(node *ast.Name) (any, error) {
 					}
 				},
 			), nil
+		case "re":
+			return callable(
```
jgiannuzzi, line 616: this should instead return an `attributeGetter` that switches between the 2 supported operations and returns a `callable`

pkg/api/aim/query/query.go
```diff
@@ -217,6 +220,22 @@ func (pq *parsedQuery) parseAttribute(node *ast.Attribute) (any, error) {
...
+				}
+				c, ok := parsedNode.(callable)
+				if !ok {
+					return nil, errors.New("unsupported node type. has to be callable type")
+				}
+
+				return c(append([]ast.Expr{&ast.Str{
+					S: py.String(strings.ToLower(attribute)),
+				}}, args...))
+			}), nil
```
jgiannuzzi, line 238: this is not the right place to deal with this — have a look at my comment about making `re` return an `attributeGetter`

pkg/api/aim/query/query.go
```diff
@@ -17,9 +17,11 @@ import (
...
 
 const (
-	OperationEndsWith   = "endswith"
-	OperationContains   = "contains"
-	OperationStartsWith = "startswith"
+	RegexpOperationMatch  = "match"
+	RegexpOperationSearch = "search"
+	OperationEndsWith     = "endswith"
+	OperationContains     = "contains"
+	OperationStartsWith   = "startswith"
```
jgiannuzzi, line 24: actual question: do we need constants if they are used in one place only? (I'm curious, I don't have a strong opinion either way)

pkg/api/aim/query/query.go
```diff
@@ -581,6 +612,57 @@ func (pq *parsedQuery) parseName(node *ast.Name) (any, error) {
...
+							return nil, errors.New("re.match function support exactly 2 arguments")
+						}
+
+						regexp, ok := args[1].(*ast.Str)
+						if !ok {
+							return nil, errors.New("first argument type for re.match function has to be ast.Str")
+						}
+						attribute, ok := args[2].(*ast.Attribute)
+						if !ok {
+							return nil, errors.New("second argument type for re.match function has to be ast.Attribute")
```
jgiannuzzi, line 637: second argument should be parsed with `parseNode` and the result should be checked to be a `clause.Column` — this will make sure that things like `re.match("regex", run.tags["mlflow.source.name"])` work and should make the code much simpler

pkg/database/database.go
```diff
@@ -75,7 +81,20 @@ func ConnectDB(
 			}
 		}
 
-		s, err := sql.Open(sqlite.DriverName, strings.Replace(dbURL.String(), "sqlite://", "file:", 1))
+		sql.Register(SQLiteCustomDriverName, &sqlite3.SQLiteDriver{
+			ConnectHook: func(conn *sqlite3.SQLiteConn) error {
+				return conn.RegisterFunc("regexp", func(re, s string) bool {
+					c, err := regexp.Compile(re)
```
jgiannuzzi, line 87: please add an LRU cache of compiled regexes — I did a super basic benchmark and cut the time in half by adding a super basic `map[string]*regexp.Regexp` cache

jgiannuzzi (changes requested):

pkg/api/aim/query/clause.go
```diff
@@ -0,0 +1,41 @@
...
+	clause.Eq
+	Dialector string
+}
+
+// Build builds positive statement.
+func (regexp Regexp) Build(builder clause.Builder) {
+	builder.WriteQuoted(regexp.Column)
+	operator := ""
+	switch regexp.Dialector {
+	case sqlite.Dialector{}.Name():
```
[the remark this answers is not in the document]
dsuhinin: yep, let's use `default` case.

pkg/api/aim/query/query.go
dsuhinin: not a big deal. we ca remove them. usually `constant` give you a more meaning but lets skip and see.

pkg/api/aim/query/query.go
dsuhinin: ook. let try to what you've suggested.

dsuhinin: done

pkg/api/aim/query/query.go

pkg/database/database.go
dsuhinin: do you think we need to take care about concurrency?

pkg/api/aim/query/query.go

pkg/api/aim/query/query.go

pkg/api/aim/query/query.go
```diff
@@ -613,53 +589,84 @@ func (pq *parsedQuery) parseName(node *ast.Name) (any, error) {
...
-						if !ok {
-							return nil, errors.New("second argument type for re.match function has to be ast.Attribute")
-						}
+								parsedNode, err := pq.parseNode(args[0])
+								if err != nil {
+									return nil, err
+								}
+								str, ok := parsedNode.(string)
+								if !ok {
+									return nil, errors.New("first argument type for re.match function has to be ast.Str")
```
jgiannuzzi, line 608:
```suggestion
									return nil, errors.New("first argument type for re.match function has to be a string")
```

pkg/api/aim/query/query.go
```diff
@@ -613,53 +589,84 @@ func (pq *parsedQuery) parseName(node *ast.Name) (any, error) {
...
+									},
+									Dialector: pq.qp.Dialector,
+								}, nil
 							},
-							Dialector: pq.qp.Dialector,
-						}, nil
+						), nil
+					case "search":
+						return callable(
+							func(args []ast.Expr) (any, error) {
```
jgiannuzzi, line 634: could you please make `match` and `search` be the same except for the `^` character?

pkg/api/aim/query/clause.go
```diff
@@ -19,10 +18,10 @@ func (regexp Regexp) Build(builder clause.Builder) {
...
 	operator := ""
 	switch regexp.Dialector {
-	case sqlite.Dialector{}.Name():
-		operator = "regexp"
 	case postgres.Dialector{}.Name():
 		operator = "~"
+	default:
+		operator = "regexp"
 	}
 	builder.WriteString(fmt.Sprintf(" %s ", operator))
```
jgiannuzzi, line 26: for consistency, can you please use the same style between `Build` and `NegationBuild`?

pkg/database/database.go
jgiannuzzi: yes, definitely!

pkg/api/aim/query/query.go
dsuhinin: done.

pkg/api/aim/query/query.go
dsuhinin: probably I can and it was before but I decided to just keep since a bit more clean even if they are almost the same. can we keep them split?

pkg/api/aim/query/clause.go
dsuhinin: yep. done.!

pkg/database/database.go
dsuhinin: ook.

dsuhinin: done.

pkg/api/aim/query/query.go
jgiannuzzi: no, I would really like us to stick to the DRY principle — maybe fold them into one private method of `parsedQuery`?

pkg/database/database.go
```diff
@@ -83,13 +86,22 @@ func ConnectDB(
...
+				if err != nil {
+					return eris.Wrap(err, "error creating lru cache to cache regexp statements")
+				}
 				return conn.RegisterFunc("regexp", func(re, s string) bool {
-					c, err := regexp.Compile(re)
-					if err != nil {
-						return false
+					result, ok := cache.Get(re)
+					if !ok {
+						complied, err := regexp.Compile(re)
```
jgiannuzzi, line 97: I think you should assign it straight to `result` instead so you avoid the duplicate `return MatchString`

pkg/api/aim/query/query.go

pkg/database/database.go
dsuhinin: done.

pkg/database/database.go
```diff
@@ -94,12 +94,10 @@ func ConnectDB(
 				return conn.RegisterFunc("regexp", func(re, s string) bool {
 					result, ok := cache.Get(re)
 					if !ok {
-						complied, err := regexp.Compile(re)
+						result, err = regexp.Compile(re)
 						if err != nil {
 							return false
 						}
```
jgiannuzzi, line 100: we need to add the result to the cache!

dsuhinin: done.

dsuhinin: btw - I don't like how it looks now. much better to have another `return` statement.

jgiannuzzi (approved):
Excellent stuff, thanks a lot @dsuhinin!

PR #179, opened 2023-07-23
ci: build Go binaries and upload them to a release

jgiannuzzi:
This PR adds a CI job to build the Go binaries for all the supported platforms and upload them as CI artifacts, enabling developers and users to try out a pull request, a branch, or a nightly build.

It also adds a CI job running after the CI has run for a tag that will create a GitHub release and upload the Go binaries to it.

The build produced by `make build` and the archive produced by `make dist` locally are the same as the ones produced by the CI.

The linux/arm64 and windows/x86_64 builds are cross-compiled from linux/x86_64 and the darwin/arm64 (macOS) is cross-compiled from darwin/x86_64. The reason for cross-compiling the Windows build is to speed up the CI (2 minutes instead of 6 minutes), as mingw is the only C compiler support by Go anyway.

All the Linux builds are statically-linked so that the same binary can be used on any distribution (even musl-based ones like Alpine).

All of the builds are done in parallel and thus the total build time takes 3 minutes in average, compared to [about 10 minutes](https://github.com/G-Research/fasttrackml/actions/runs/5630851571/job/15257220001) with GoReleaser (which also required using a [3rd-party Docker image](https://github.com/goreleaser/goreleaser-cross) that includes a [dodgy-licensed macOS cross-compiler](https://github.com/tpoechtrager/osxcross)).

Examples runs on my fork:
- [binary archives built for a tag push](https://github.com/jgiannuzzi/fasttrackml/actions/runs/5656348702) (notice the [artifacts](https://github.com/jgiannuzzi/fasttrackml/suites/14548248506/artifacts/823389228) created — their build properties can be validated with `go version -m <binary>`)
- [release created for a tag push](https://github.com/jgiannuzzi/fasttrackml/actions/runs/5656420264/job/15323414839#step:3:14) (the Docker publish failure is due to my fork not having the Docker Hub secrets — not a real problem as this PR _disables_ running releases on forks)
- [release skipped for a push to `main`](https://github.com/jgiannuzzi/fasttrackml/actions/runs/5635948180/job/15267494353)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -8,12 +8,14 @@ on:
       - main
       - v*
 
+permissions:
+  contents: read
+
 jobs:
-  release:
-    name: Publish container image to DockerHub
-    runs-on: ubuntu-latest
-    environment: release
+  validate:
+    name: Validate ref
     if: github.event.workflow_run.event == 'push' && github.event.workflow_run.conclusion == 'success' && !github.event.repository.fork
+    runs-on: ubuntu-latest
     steps:
       - name: Checkout
         uses: actions/checkout@v3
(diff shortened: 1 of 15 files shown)


jgiannuzzi:
this PR supersedes #140

dsuhinin (dismissed):
Looks good!

.github/workflows/release.yml
```diff
@@ -47,6 +49,56 @@ jobs:
...
+    needs: validate
+    if: github.event.workflow_run.head_branch != 'main'
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+    steps:
+      - name: Download artifact
+        uses: actions/github-script@v6
+        with:
+          script: |
```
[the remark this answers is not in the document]
jgiannuzzi: That's a good point, I'll see how this could be simplified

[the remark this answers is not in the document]
jgiannuzzi: Simplified to use GitHub CLI in cd2e00d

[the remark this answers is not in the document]
jgiannuzzi: Links in PR description updated to point to new runs using the change in cd2e00d

.github/workflows/release.yml
```diff
@@ -58,28 +58,9 @@ jobs:
...
-               repo: context.repo.repo,
-               artifact_id: matchArtifact.id,
-               archive_format: 'zip',
-            });
-            let fs = require('fs');
-            fs.writeFileSync(`${process.env.GITHUB_WORKSPACE}/fasttrackml-archives.zip`, Buffer.from(download.data));
-
-      - name: Unzip artifact
-        run: unzip -d dist fasttrackml-archives.zip
+        run: gh run download ${{ github.event.workflow_run.id }} --repo ${{ github.event.workflow_run.repository.full_name }} --name fasttrackml-archives --dir dist
```
suprjinx, line 61: that's definitely a lot simpler, I like it :-)

Makefile
```diff
@@ -6,8 +6,36 @@
...
+# Go ldflags.
+# Set version to git tag if available, otherwise use commit hash.
+# Strip debug symbols and disable DWARF generation.
+# Build static binaries on Linux.
+GO_LDFLAGS=-s -w -X github.com/G-Research/fasttrackml/pkg/version.Version=$(shell git describe --tags --always --dirty --match='v*' 2> /dev/null | sed 's/^v//')
+ifeq ($(shell go env GOOS),linux)
+  GO_LDFLAGS+=-linkmode external -extldflags -static
+endif
+# Go build tags.
+GO_BUILDTAGS=$(shell jq -r '."go.buildTags"' .vscode/settings.json)
```
suprjinx, line 23: should we put the tags here instead of the vscode project settings? This seems like a more authoritative location, and `jq` would not then be required

suprjinx (approved):
LGTM -- had one suggestion/comment about the build tags, take it or leave it

Makefile
jgiannuzzi: I'd love to have a more naturally authoritative place than the vscode settings to store this! However I haven't figured out a way to make vscode read the tags from a different place than it's settings and I would like codespaces/devcontainers to just work so that any contributor can pick up the project instantly.
If you figure out a way, then let's go for it!

jgiannuzzi: This had been bugging since very early on and your remark pushed to finally figure this out! I've centralised all build tags in `.go-build-tags` in 8047c6e. Thanks for pushing me @suprjinx, much appreciated!

suprjinx (dismissed):
Looks great

suprjinx (approved):

PR #183, opened 2023-07-25
Redesign `contains` function to be python like - `in`

dsuhinin:


suprjinx (approved):
LGTM

pkg/api/aim/query/query.go
```diff
@@ -766,7 +770,7 @@ func reverseComparison(op ast.CmpOp, left any, right clause.Column) (ast.CmpOp,
 		return ast.Lt, right, left, nil
 	case ast.GtE:
 		return ast.LtE, right, left, nil
-	case ast.Eq, ast.Is, ast.NotEq, ast.IsNot:
+	case ast.NotIn, ast.In, ast.Eq, ast.Is, ast.NotEq, ast.IsNot:
```
jgiannuzzi, line 773: `reverseComparison` is meant for those comparison operations that require a column to the left — in that case it will swap the values and return the reversed comparison operator.
For `in` and `not in`, it doesn't make sense, and by doing this you then end up with the reversed logic when handling these operations in `newSqlComparison`. I was extremely confused when reading this and kept on thinking "but surely the string needs to be on the left and the column needs to be on the right?".

Could you please handle this differently? One idea that springs to my mind is to handle the case of "left is a string AND op is `in` or `not in` AND right is a column" in `parseCompare` and make that one return a `clause.Like`.

jgiannuzzi (changes requested):

pkg/api/aim/query/query.go
dsuhinin: Yes, it looks a bit strange but this is like a one place where I managed to intercept this situation. ok, Ill take a look one more time.

dsuhinin: probably it looks strange a bit but it works.

dsuhinin: ooooook. it looks also a bit strange but.... take a look.

jgiannuzzi: I'm still not convinced by this approach, sorry!
Swapping the left and right operands makes the logic hard(er) to follow — the point of `newSqlComparison` was to have a 1-to-1 mapping between a Python comparison operator and a SQL comparison operator, but in this particular case, the Python `in` needs to be mapped to a SQL `like`, so I think we should directly return a `clause.Like` in `parseCompare`

dsuhinin: @jgiannuzzi no no. please take a `Files changed` tab.

jgiannuzzi (commented):
Unfortunately this is now breaking comparisons like `"hello" == run.name`

dsuhinin:
> Unfortunately this is now breaking comparisons like `"hello" == run.name`

ok, Ill close and remove this PR and do it from the scratch again.

PR #185, opened 2023-07-26
Integration tests for Aim experiments API

fabiovincenzi:


tests/integration/golang/aim/experiment/delete_test.go
```diff
@@ -40,31 +38,14 @@ func (s *DeleteExperimentTestSuite) Test_Ok() {
...
-			Valid: true,
-		},
-		LastUpdateTime: sql.NullInt64{
-			Int64: time.Now().UTC().UnixMilli(),
-			Valid: true,
-		},
-		LifecycleStage:   models.LifecycleStageActive,
-		ArtifactLocation: "/artifact/location",
-	})
+	test_experiments, err := s.fixtures.CreateExperiments(context.Background(), 1)
```
dsuhinin, line 41: can you please use `camelCase`?

dsuhinin: can we please create as minimum as possible test records for each test? so prev `s.fixtures.CreateExperiment` looks much better and understandable.

tests/integration/golang/aim/experiment/delete_test.go
```diff
@@ -40,31 +38,14 @@ func (s *DeleteExperimentTestSuite) Test_Ok() {
...
 	assert.Nil(s.T(), err)
+	experiment := test_experiments[0]
 
 	experiments, err := s.fixtures.GetTestExperiments(context.Background())
 	length := len(experiments)
 
-	var resp any
+	var resp response.DeleteExperiment
```
dsuhinin, line 48: nice!

tests/integration/golang/aim/experiment/delete_test.go
```diff
@@ -40,31 +38,14 @@ func (s *DeleteExperimentTestSuite) Test_Ok() {
...
-		LastUpdateTime: sql.NullInt64{
-			Int64: time.Now().UTC().UnixMilli(),
-			Valid: true,
-		},
-		LifecycleStage:   models.LifecycleStageActive,
-		ArtifactLocation: "/artifact/location",
-	})
+	test_experiments, err := s.fixtures.CreateExperiments(context.Background(), 1)
 	assert.Nil(s.T(), err)
+	experiment := test_experiments[0]
```
dsuhinin, line 43: In case of just one experiment you don't need to do all of this unnecessary stuff.

tests/integration/golang/aim/experiment/get_experiment_activity_test.go
```diff
@@ -0,0 +1,102 @@
...
+	runFixtures, err := fixtures.NewRunFixtures(helpers.GetDatabaseUri())
+	assert.Nil(s.T(), err)
+	s.runFixtures = runFixtures
+
+	expFixtures, err := fixtures.NewExperimentFixtures(helpers.GetDatabaseUri())
+	assert.Nil(s.T(), err)
+	s.experimentFixtures = expFixtures
+}
+
+func (s *GetExperimentActivityTestSuite) Test_Ok() {
```
dsuhinin, line 45: to have the same structure of each test, can you please start each test with:
```
defer func() {
 		assert.Nil(s.T(), s.runFixtures.UnloadFixtures())
 		assert.Nil(s.T(), s.experimentFixtures.UnloadFixtures())
 	}()
```

tests/integration/golang/aim/experiment/get_experiment_activity_test.go
```diff
@@ -0,0 +1,102 @@
...
+	var resp response.GetExperimentActivity
+	err = s.client.DoGetRequest(
+		fmt.Sprintf(
+			"/experiments/%d/activity", *experiment.ID,
+		),
+		&resp,
+	)
+	assert.Nil(s.T(), err)
+	activity := map[string]int{}
+	for _, r := range runs {
```
dsuhinin, line 68: can we move this into helper with name something like that -> `TransformRunsToActivityMap` or whatever you can suggest. In that case test will be a bit more clean and understandable.

tests/integration/golang/aim/experiment/get_experiment_activity_test.go
```diff
@@ -0,0 +1,102 @@
...
+		key := time.UnixMilli(r.StartTime.Int64).Format("2006-01-02T15:00:00")
+		activity[key] += 1
+	}
+
+	assert.Equal(s.T(), resp.NumRuns, len(runs))
+	assert.Equal(s.T(), resp.NumArchivedRuns, 0)
```
dsuhinin, line 74: maybe we also need to create `archived` runs? to fully fulfil test case?

tests/integration/golang/aim/experiment/get_experiment_runs_test.go
```diff
@@ -0,0 +1,107 @@
...
+	exp := &models.Experiment{
+		Name:           uuid.New().String(),
+		LifecycleStage: models.LifecycleStageActive,
+	}
+	experiment, err := s.experimentFixtures.CreateExperiment(context.Background(), exp)
+	assert.Nil(s.T(), err)
+
+	runs, err := s.runFixtures.CreateRuns(context.Background(), exp, 10)
+	assert.Nil(s.T(), err)
+	defer func() {
```
dsuhinin, line 54: the same comment here.

tests/integration/golang/aim/experiment/get_experiment_runs_test.go
```diff
@@ -0,0 +1,107 @@
...
+		assert.Nil(s.T(), s.runFixtures.UnloadFixtures())
+		assert.Nil(s.T(), s.experimentFixtures.UnloadFixtures())
+	}()
+
+	offset := 8
```
dsuhinin, line 59: why do we need it?

tests/integration/golang/aim/experiment/get_experiment_runs_test.go
```diff
@@ -0,0 +1,107 @@
...
+	err = s.client.DoGetRequest(
+		fmt.Sprintf(
+			"/experiments/%d/runs?limit=4&offset=%s", *experiment.ID, runs[offset].ID,
+		),
+		&resp,
+	)
+	assert.Nil(s.T(), err)
+
+	assert.Equal(s.T(), 4, len(resp.Runs))
+	assert.Equal(s.T(), runs[offset-1].ID, resp.Runs[0].ID)
```
dsuhinin, line 70: will be nice to check `run` object fully. not only ID.

tests/integration/golang/aim/experiment/get_experiment_runs_test.go
```diff
@@ -0,0 +1,107 @@
...
+			ID:    "123",
+		},
+	}
+
+	for _, tt := range testData {
+		s.T().Run(tt.name, func(t *testing.T) {
+			var resp api.ErrorResponse
+			err := s.client.DoGetRequest(
+				fmt.Sprintf(
+					"/experiments/%s/runs?limit=4", tt.ID,
```
dsuhinin, line 99: does `limit` parameters always required?

tests/integration/golang/aim/experiment/get_experiment_test.go
```diff
@@ -0,0 +1,111 @@
...
+		},
+		LastUpdateTime: sql.NullInt64{
+			Int64: time.Now().UTC().UnixMilli(),
+			Valid: true,
+		},
+		LifecycleStage:   models.LifecycleStageActive,
+		ArtifactLocation: "/artifact/location",
+	})
+	assert.Nil(s.T(), err)
+	defer func() {
```
dsuhinin, line 61: the same comment here.

tests/integration/golang/aim/experiment/get_experiments_test.go
```diff
@@ -0,0 +1,59 @@
...
+
+	fixtures, err := fixtures.NewExperimentFixtures(helpers.GetDatabaseUri())
+	assert.Nil(s.T(), err)
+	s.fixtures = fixtures
+}
+
+func (s *GetExperimentsTestSuite) Test_Ok() {
+	experiments, err := s.fixtures.CreateExperiments(context.Background(), 5)
+	assert.Nil(s.T(), err)
+	defer func() {
```
dsuhinin, line 40: the same comment here.

tests/integration/golang/aim/experiment/get_experiments_test.go
```diff
@@ -0,0 +1,59 @@
...
+		assert.Nil(s.T(), s.fixtures.UnloadFixtures())
+	}()
+	var resp []response.GetExperiment
```
dsuhinin, line 43: maybe we can create own type under the hood of `response` package -> type Experiments = []response.GetExperiment ?

tests/integration/golang/aim/experiment/get_experiments_test.go
```diff
@@ -0,0 +1,59 @@
...
+
+	err = s.client.DoGetRequest(
+		fmt.Sprintf(
+			"/experiments/",
+		),
+		&resp,
+	)
+	assert.Nil(s.T(), err)
+
+	for idx := 0; idx < 5; idx++ {
```
dsuhinin, line 53: Im not fully sure that this is correct. If endpoint will give 6 records instead of 5, but you iterate only over 5 so it will be a mistake in endpoint logic. So will be nice to compare the length of response. Also, you can iterate over `resp` directly and compare with what you have in `runs`, so you will always make sure that number of returned records is the same with what you've created.

dsuhinin (changes requested):

tests/integration/golang/aim/experiment/get_experiment_activity_test.go
suprjinx: also, I don't think we need two calls to UnloadFixtures(), since that method is included from base and clears everything

tests/integration/golang/aim/experiment/get_experiment_runs_test.go
```diff
@@ -0,0 +1,110 @@
...
+
+	expFixtures, err := fixtures.NewExperimentFixtures(helpers.GetDatabaseUri())
+	assert.Nil(s.T(), err)
+	s.experimentFixtures = expFixtures
+}
+
+func (s *GetExperimentRunsTestSuite) Test_Ok() {
+	defer func() {
+		assert.Nil(s.T(), s.runFixtures.UnloadFixtures())
+		assert.Nil(s.T(), s.experimentFixtures.UnloadFixtures())
```
dsuhinin, line 47: please, stay only one `unload`.

tests/integration/golang/aim/experiment/get_experiment_test.go
```diff
@@ -0,0 +1,111 @@
...
+		fmt.Sprintf(
+			"/experiments/%d", *experiment.ID,
+		),
+		&resp,
+	)
+	assert.Nil(s.T(), err)
+
+	assert.Equal(s.T(), *experiment.ID, resp.ID)
+	assert.Equal(s.T(), experiment.Name, resp.Name)
+	assert.Equal(s.T(), experiment.LifecycleStage == models.LifecycleStageDeleted, resp.Archived)
```
dsuhinin, line 76: won't it be enough to just check `assert.Equal(s.T(), false, resp.Archved)` ?
also can you please check other properties: `description, creation_time`. I know that description right now is `nil` but it is nice to catch this moment when it will be changed.

tests/integration/golang/aim/experiment/get_experiments_test.go
dsuhinin: The main point why I asked that, because `response.GetExperiment` is belongs to `get` endpoint and we know that from the name, but `[]response.GetExperiment` is an array of what? results from `get` endpoint? It is semantically wrong and confusing. So yep, it will still good to fix it and have something like `response.SearchExperiments` or `response.Experiments` or `response.GetExperiments` which give you exact meaning of this struct.

tests/integration/golang/aim/experiment/get_experiments_test.go
```diff
@@ -0,0 +1,59 @@
...
+	err = s.client.DoGetRequest(
+		fmt.Sprintf(
+			"/experiments/",
+		),
+		&resp,
+	)
+	assert.Nil(s.T(), err)
+	assert.Equal(s.T(), len(experiments), len(resp))
+	for idx := 0; idx < len(experiments); idx++ {
+		assert.Equal(s.T(), *experiments[idx].ID, resp[idx].ID)
```
dsuhinin, line 54: lets compare as much as endpoint can return.

dsuhinin (commented):
few small remarks.

tests/integration/golang/aim/experiment/get_experiments_test.go
dsuhinin: @suprjinx you did probably the same which is totally correct.
Screenshot 2023-07-31 at 09 52 49

tests/integration/golang/fixtures/run.go
```diff
@@ -49,6 +49,11 @@ func (f RunFixtures) CreateRun(
 	return run, nil
 }
 
+func (f RunFixtures) ArchiveRun(ctx context.Context, ids []string) error {
```
dsuhinin, line 52: can you please add comment which end with dot?

dsuhinin (approved):
good to go!

tests/integration/golang/aim/experiment/get_experiments_test.go
suprjinx: ok!

PR #187, opened 2023-07-26
Aim integration tests for runs endpoints

suprjinx:


tests/integration/golang/aim/run/get_run_metrics_test.go
```diff
@@ -0,0 +1,127 @@
...
+	exp := &models.Experiment{
+		Name:           uuid.New().String(),
+		LifecycleStage: models.LifecycleStageActive,
+	}
+	_, err = s.experimentFixtures.CreateExperiment(context.Background(), exp)
+	assert.Nil(s.T(), err)
+
+	runs, err := s.runFixtures.CreateRuns(context.Background(), exp, 1)
+	assert.Nil(s.T(), err)
+	s.run = runs[0]
```
dsuhinin, line 58: Sorry, I don't understand the point when you create `array` of `runs`? Here you create single `experiment` - which is nice, but then you create array of `runs` with single run? what is the point? we also have method to create just a single run.

tests/integration/golang/aim/run/get_runs_active_test.go
```diff
@@ -0,0 +1,116 @@
...
+			wantRunCount: 2,
+		},
+		{
+			name:         "GetActiveRunsWhenNotPresent",
+			wantRunCount: 0,
+		},
+	}
+	for _, tt := range tests {
+		s.T().Run(tt.name, func(T *testing.T) {
+			if tt.wantRunCount == 0 {
```
dsuhinin, line 79: it is hard to understand what is going on here when you read this moment first time. Can you please remove this test at all? I can't see any reason to have it like I mentioned before.

tests/integration/golang/aim/run/update_run_test.go
```diff
@@ -0,0 +1,130 @@
...
+	exp := &models.Experiment{
+		Name:           uuid.New().String(),
+		LifecycleStage: models.LifecycleStageActive,
+	}
+	_, err = s.experimentFixtures.CreateExperiment(context.Background(), exp)
+	assert.Nil(s.T(), err)
+
+	runs, err := s.runFixtures.CreateRuns(context.Background(), exp, 1)
+	assert.Nil(s.T(), err)
+	s.run = runs[0]
```
dsuhinin, line 53: probably the same question here -> why we can create one `experiment` but we can't create on run?

tests/integration/golang/aim/run/update_run_test.go
```diff
@@ -0,0 +1,130 @@
...
+	}{
+		{
+			name: "UpdateOneRun",
+			request: request.UpdateRunRequest{
+				RunID:    &(s.run.ID),
+				Name:     &newName,
+				Status:   &newStatus,
+				Archived: &archived,
+			},
+			verifyFn: func(t *testing.T, resp response.Success) {
```
dsuhinin, line 76: why do we need such function? why we can't verify this moment in test loop?:
```
s.T().Run(tt.name, func(T *testing.T) {
 			var resp response.Success
 			err := s.client.DoPutRequest(
 				fmt.Sprintf("/runs/%s", *tt.request.RunID),
 				tt.request,
 				&resp,
 			)
 			assert.Nil(s.T(), err)
 			if tt.verifyFn != nil {
 				tt.verifyFn(s.T(), resp)
 			}
 		})
```

tests/integration/golang/fixtures/run.go
```diff
@@ -119,3 +154,49 @@ func (f RunFixtures) FindMinMaxRowNums(
...
+func (f RunFixtures) CreateTag(
+	ctx context.Context, tag models.Tag,
+) error {
+	if err := f.tagRepository.CreateRunTagWithTransaction(ctx, f.db, tag.RunID, tag.Key, tag.Value); err != nil {
+		return eris.Wrap(err, "error creating run tag")
+	}
+	return nil
+}
+
+func (f RunFixtures) CreateMetrics(
```
dsuhinin, line 168: can you please add comments? every comment must end with a dot.

tests/integration/golang/fixtures/run.go
```diff
@@ -78,11 +82,42 @@ func (f RunFixtures) CreateRuns(
...
+		if err != nil {
+			return nil, err
+		}
+
 		runs = append(runs, run)
 	}
 	return runs, nil
 }
 
+// GetTestRun fetches one run
```
dsuhinin, line 106: every comment must end with a dot

tests/integration/golang/fixtures/run.go
```diff
@@ -59,7 +63,7 @@ func (f RunFixtures) UpdateRun(
 	return nil
 }
 
-// CreateRuns creates some num runs belonging to the experiment
+// CreateRuns creates some num runs belonging to the experiment, with tags and metrics
```
dsuhinin, line 66: every comment must end with a dot

dsuhinin (changes requested):

tests/integration/golang/aim/run/get_run_metrics_test.go
suprjinx: the `CreateRuns` method is more convenient because it populates boilerplate, adds metrics, etc. We could refactor to separate singular/plural versions of this, it is different than the the `CreateRun` method

tests/integration/golang/aim/run/update_run_test.go
suprjinx: as above, this method populates the Run for you

tests/integration/golang/aim/run/get_runs_active_test.go
suprjinx: this is the test for empty db table -- so it needs a different setup. I don't think we should remove it, seems like a valid test case. Either we move setup into the test for other cases or clean up in this case -- but I think it helps readability to leave setup outside the test function as much as possible.

tests/integration/golang/aim/run/get_run_metrics_test.go
```diff
@@ -0,0 +1,127 @@
...
+			err := s.client.DoPostRequest(
+				fmt.Sprintf("/runs/%s/metric/get-batch", tt.runID),
+				tt.request,
+				&resp,
+			)
+			assert.Nil(s.T(), err)
+			assert.Equal(s.T(), len(tt.request), len(resp))
+			for i := 0; i < 2; i++ {
+				assert.Equal(s.T(), 2, len(resp[i].Values))
+				assert.Equal(s.T(), 2, len(resp[i].Iters))
```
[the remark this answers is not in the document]
suprjinx: done

tests/integration/golang/aim/run/get_runs_active_test.go
suprjinx: added a comment, hopefully that helps? let me know

tests/integration/golang/aim/run/update_run_test.go
suprjinx: removed -- must have had a reason at some point 🤷

tests/integration/golang/fixtures/run.go
suprjinx: done

tests/integration/golang/fixtures/run.go
suprjinx: done

tests/integration/golang/fixtures/run.go
suprjinx: done

tests/integration/golang/fixtures/run.go
```diff
@@ -78,12 +94,44 @@ func (f RunFixtures) CreateRuns(
 		if err != nil {
 			return nil, err
 		}
+		tag := models.Tag{
```
dsuhinin, line 97: it is really, really hard to control over that in tests. for example you need to create different tags, metrics, params  with different parameters and then use them. right now it is impossible, because everything is hardcoded and the same for each test. Also you not always need to create `tags`, `metrics`, etc for each test. So it will be reasonable to create such data right in tests, where you can put any information you, information(parameters) could be different and you can see it.

tests/integration/golang/fixtures/run.go
```diff
@@ -78,12 +94,44 @@ func (f RunFixtures) CreateRuns(
...
+		}
+
 		runs = append(runs, run)
 	}
 	return runs, nil
 }
 
-// GetTestRuns fetches all runs for an experiment
+// GetTestRun fetches one run.
+func (f RunFixtures) GetTestRun(
```
dsuhinin, line 119: I totally confused. We have `CreateExampleRun` and `CreateRun` can we please move forward with only on function?

tests/integration/golang/aim/run/delete_batch_test.go
```diff
@@ -45,11 +45,14 @@ func (s *DeleteBatchTestSuite) SetupTest() {
 	_, err = s.experimentFixtures.CreateExperiment(context.Background(), exp)
 	assert.Nil(s.T(), err)
 
-	s.runs, err = s.runFixtures.CreateRuns(context.Background(), exp, 10)
+	s.runs, err = s.runFixtures.CreateExampleRuns(context.Background(), exp, 10)
```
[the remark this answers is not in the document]
suprjinx: how would you suggest we name this? it is a "Run which is populated"?

tests/integration/golang/fixtures/run.go
suprjinx: `CreateRun` persists a Run provided as parameter, but `CreateExampleRun` does not require a Run parameter, it creates for you.

tests/integration/golang/fixtures/run.go
suprjinx: I don't think you want to fill each test with a lot of repetitive object creation -- better to have this in a helper and let the test focus on the subject of testing.

tests/integration/golang/aim/run/update_run_test.go
```diff
@@ -0,0 +1,122 @@
...
+	assert.Nil(s.T(), err)
+}
+
+func (s *UpdateRunTestSuite) Test_Ok() {
+	defer func() {
+		assert.Nil(s.T(), s.runFixtures.UnloadFixtures())
+	}()
+	newName := fmt.Sprintf("%v%v", s.run.Name, "-new")
+	newStatus := string(models.StatusFinished)
+	archived := true
```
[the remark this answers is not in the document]
suprjinx: will look, not sure I understand

tests/integration/golang/aim/run/get_run_metrics_test.go
```diff
@@ -0,0 +1,147 @@
...
+				fmt.Sprintf("/runs/%s/metric/get-batch", tt.runID),
+				tt.request,
+				&resp,
+			)
+			assert.Nil(s.T(), err)
+			assert.Equal(s.T(), len(tt.request), len(resp))
+			// this is a test-side conversion of models to response to verify
+			// the response given by the endpoint
+			runMetricMap := convertModelToResponse(run.Metrics)
+			for _, respElement := range resp {
```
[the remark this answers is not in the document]
suprjinx: I think for this case we need to verify that the database has recorded the change, yes? `response` is not a confirmation of that.

tests/integration/golang/aim/run/get_run_metrics_test.go
```diff
@@ -0,0 +1,147 @@
...
+				{
+					Context: map[string]string{},
+					Name:    "key2",
+				},
+			},
+		},
+	}
+	for _, tt := range tests {
+		s.T().Run(tt.name, func(T *testing.T) {
+			run, err := s.runFixtures.GetTestRun(context.Background(), tt.runID)
```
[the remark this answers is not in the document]
suprjinx: will remove the `TestRun` method, but `Example` is trying to indicate that a populated example is being created -- how would you make that clear?

tests/integration/golang/aim/run/get_run_metrics_test.go
```diff
@@ -0,0 +1,147 @@
...
+				&resp,
+			)
+			assert.Nil(s.T(), err)
+			assert.Equal(s.T(), "Not Found", resp.Message)
+		})
+	}
+}
+
+// convertModelToResponse will convert []models.Metric to response struct like the endpoint response.
+func convertModelToResponse(metrics []models.Metric) map[string]response.RunMetrics {
```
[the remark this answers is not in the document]
suprjinx: i think for create/update we need to check database, not the response

tests/integration/golang/aim/run/get_run_metrics_test.go
```diff
@@ -0,0 +1,147 @@
...
+	}
+	for _, tt := range tests {
+		s.T().Run(tt.name, func(T *testing.T) {
+			var resp response.Error
+			err := s.client.DoGetRequest(
+				fmt.Sprintf("/runs/%s/metric/get-batch", tt.runID),
+				&resp,
+			)
+			assert.Nil(s.T(), err)
+			assert.Equal(s.T(), "Not Found", resp.Message)
```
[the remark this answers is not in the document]
suprjinx: added to table

tests/integration/golang/aim/run/update_run_test.go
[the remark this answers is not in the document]
suprjinx: changed to use GetPointer, but I think `&var` may be more concise and allows you to name the thing.

tests/integration/golang/aim/run/get_run_metrics_test.go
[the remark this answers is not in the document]
suprjinx: Nvm, I see what you're saying -- changed to compare to an expected response

tests/integration/golang/aim/run/get_run_metrics_test.go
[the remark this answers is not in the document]
suprjinx: Deleted this, just comparing response now

tests/integration/golang/aim/run/get_runs_active_test.go
suprjinx: removing

PR #188, opened 2023-07-27
Fix issue with logging batch metrics

jgiannuzzi:
Fix regression introduced in #63 that prevented from logging more than 32768 metrics at once. It also raises Fiber's `BodyLimit` from 4MB to 16MB to further increase the amount of metrics that can be logged in one call.


tests/integration/golang/mlflow/run/log_batch_test.go
```diff
@@ -215,6 +215,35 @@ func (s *LogBatchTestSuite) TestMetrics_Ok() {
 				"key2": 4,
 			},
 		},
+		{
+			name: "LogMany",
+			request: &request.LogBatchRequest{
+				RunID: s.run.ID,
+				Metrics: func() []request.MetricPartialRequest {
+					metrics := make([]request.MetricPartialRequest, 100*1000)
+					for k := 0; k < 100; k++ {
```
dsuhinin, line 224: why it couldn't be just one cycle?

suprjinx: It seems like LogMany would want more than one, so 2x?

jgiannuzzi: I'm not entirely sure I understand your question. Is it about merging LogSeveral and LogMany? Or is it about doing 100 metrics with 1000 values each?

suprjinx: my question was really just responding to Dmitriy's question :-)

jgiannuzzi: Oh I was asking @dsuhinin 😅

dsuhinin: no no, never mind. It seems like a got the idea. sorry for the late answer.

dsuhinin (approved):

PR #190, opened 2023-07-27
Increase integration tests coverage for MLFlow

dsuhinin:


tests/integration/golang/mlflow/run/delete_test.go
```diff
@@ -39,33 +39,39 @@ func (s *DeleteRunTestSuite) SetupTest() {
 	expFixtures, err := fixtures.NewExperimentFixtures(helpers.GetDatabaseUri())
 	assert.Nil(s.T(), err)
 	s.experimentFixtures = expFixtures
+}
+
+func (s *DeleteRunTestSuite) Test_Ok() {
+	defer func() {
+		assert.Nil(s.T(), s.runFixtures.UnloadFixtures())
+		assert.Nil(s.T(), s.experimentFixtures.UnloadFixtures())
```
suprjinx, line 47: i think all the fixtures inherit Base/UnloadFixtures() -- so only need one call

dsuhinin: yep yep, you are totally right. it seems I just missed this place.

tests/integration/golang/mlflow/run/delete_test.go
```diff
@@ -39,33 +39,39 @@ func (s *DeleteRunTestSuite) SetupTest() {
...
+	experiment, err := s.experimentFixtures.CreateExperiment(context.Background(), &models.Experiment{
 		Name:           uuid.New().String(),
 		LifecycleStage: models.LifecycleStageActive,
-	}
-	_, err = s.experimentFixtures.CreateExperiment(context.Background(), exp)
+	})
 	assert.Nil(s.T(), err)
 
 	// create run for the experiment
-	runs, err := s.runFixtures.CreateRuns(context.Background(), exp, 1)
```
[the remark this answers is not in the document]
dsuhinin: my main point which I follow in any tests - is when you red it like I first time you should have everything before your eyes and you have to understand what is going on. So when you read test which creates by some reason array but then we operate with only first item from this array - it is a bit strange at least of me. Also, when I read a test I want to see like immediately what is going on? which test records has been created? how many? with which data? So all of that just helps me to quickly understand the logic. Usually I don't care about repetition in tests if it makes clear test in general.

tests/integration/golang/mlflow/run/update_test.go
```diff
@@ -1,3 +1,135 @@
...
+			Name:    "UpdatedName",
+			Status:  string(models.StatusScheduled),
+			EndTime: 1111111111,
+		},
+		&resp,
+	)
+	assert.Nil(s.T(), err)
+	assert.NotEmpty(s.T(), resp.RunInfo.ID)
+	assert.NotEmpty(s.T(), resp.RunInfo.UUID)
+	assert.Equal(s.T(), "UpdatedName", resp.RunInfo.Name)
```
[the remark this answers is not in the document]
dsuhinin: we probably can. yep, let's do that. thank you!

tests/integration/golang/mlflow/run/delete_test.go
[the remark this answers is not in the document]
suprjinx: i think it's worth creating fixture/helper methods that create the typical setup we need for the test without having to do a lot of repetitive setup. I'll rename this in my PR and create a "singlular" version.

tests/integration/golang/mlflow/run/update_test.go
```diff
@@ -1,3 +1,141 @@
...
+	assert.Equal(s.T(), fmt.Sprintf("%d", *experiment.ID), resp.RunInfo.ExperimentID)
```
suprjinx, line 93: I think we should be comparing the expectation to the database record, instead of the response struct -- just to make sure the change is actually persisted. The response may not reflect the state of the db.

dsuhinin: yep, that's done.

PR #195, opened 2023-07-30
Delete first run in AimUI

### LINKED ISSUE #158: Unable to delete the first run in Aim UI
Here is an example of trying to delete the first run in Aim UI.
The run was created in v0.2.0 of FML.

https://github.com/G-Research/fasttrackml/assets/3692455/7daee554-4e6c-4162-a8b3-46897c151dbf

galadd:
Description
The PR fixes #158 
As mentioned in the issue by @jgiannuzzi, the row nums starts from 0. I haven’t changed that. I have only allowed the error to be for values less than 0


pkg/api/mlflow/dao/repositories/run.go
```diff
@@ -266,8 +266,8 @@ func getMinRowNum(runs []models.Run) models.RowNum {
 
 // renumberRows will update the runs.row_num field with the correct ordinal
 func (r RunRepository) renumberRows(tx *gorm.DB, startWith models.RowNum) error {
-	if startWith <= models.RowNum(0) {
-		return eris.Errorf("attempting to renumber with 0 or less row number value")
+	if startWith < models.RowNum(0) {
+		return eris.Errorf("attempting to renumber with less than 0 row number value")
```
suprjinx, line 270: nice simple fix :-) Can you add a test for this situation? I think `tests/integration/golang/aim/run/delete_test.go` is verifying the renumbering, but it only exercises delete when there are a few records, not a single record.

suprjinx (dismissed):
Nice!

dsuhinin:
@galadd can we try to cover this case with `unit` or `integration` tests or even with both of them?

pkg/api/mlflow/dao/repositories/run.go
dsuhinin, line 270: I know it was before, but please change `Errorf` to `New`

pkg/api/mlflow/dao/repositories/run.go
```diff
@@ -266,8 +266,8 @@ func getMinRowNum(runs []models.Run) models.RowNum {
 
 // renumberRows will update the runs.row_num field with the correct ordinal
 func (r RunRepository) renumberRows(tx *gorm.DB, startWith models.RowNum) error {
-	if startWith <= models.RowNum(0) {
```
suprjinx, line 269: Can we add a private unit test in the `repositories` package which will test this function?

pkg/api/mlflow/dao/repositories/run.go
galadd: I would also like to point out that Errorf is also used in 2 other places with just strings. 
```go
// pkg/api/mlflow/dao/repositories/run.go
// line 175
if len(runs) != len(ids) {
			return eris.Errorf("count of deleted runs does not match length of ids input (invalid run ID?)")
		}
		
// and
// pkg/api/mlflow/dao/repositories/experiment.go
// line 138
if len(experiments) != len(ids) {
			return eris.Errorf("count of deleted experiments does not match length of ids input (invalid experiment ID?)")
		}
```

Should I change these to `New` too

dsuhinin: yes please! nice catch!

pkg/api/mlflow/dao/repositories/run.go
galadd: I have been working on this, and I was using a mock for the `tx *gorm.DB`. Though I can test for the function to fail (`startWith <= models.RowNum(0)`), to test for the function to pass, there is an execution that also takes arguments (line 279). For the mock, I always write and expect the execution, ie

```go
mock.ExpectExec(`LOCK TABLE runs`).
		WillReturnResult(sqlmock.NewResult(0, 0))
```

 but there's no Expect Execution with Arguments I have seen. I will keep looking on this, or is there an alternative way to how we intend to approach this

dsuhinin: yep, I know that `unit` test could be a bit complicated because of that. Can you also take a look into `integration` test for this particular case? It should be a bit easier I hope.

galadd: It has been covered in this [commit](https://github.com/G-Research/fasttrackml/pull/195/files#diff-948cf7e6f69d6f4393eeb54b77a7322b533ce6775807b37190e9dd86c567b7b9) to the delete_test integration

dsuhinin (dismissed):

pkg/api/mlflow/dao/repositories/run_test.go
```diff
@@ -0,0 +1,32 @@
...
+
+	"github.com/stretchr/testify/assert"	
+	"gopkg.in/DATA-DOG/go-sqlmock.v1"	
+	"gorm.io/driver/postgres"
+	"gorm.io/gorm"
+
+	"github.com/G-Research/fasttrackml/pkg/api/mlflow/dao/models"
+)
+
+func TestRenumberRowsForNegativeRowNumber(t *testing.T) {
```
suprjinx, line 14: Looks very good -- nice job with the mock. Since this is private, I'd say rename to "Test_renumberRows" -- and make it table-driven with a couple of cases (-1, 0, 1) as the startsWith argument

suprjinx (approved):

pkg/api/mlflow/dao/repositories/run_test.go
```diff
@@ -0,0 +1,62 @@
...
+
+	dialector := postgres.New(postgres.Config{
+		Conn:       mockDb,
+		DriverName: "postgres",
+	})
+	db, _ := gorm.Open(dialector, &gorm.Config{})
+
+	repo := NewRunRepository(db)
+
+	for _, tc := range testData {
```
suprjinx, line 52: for table driven, usually need something like `t.Run(tc.Name, func(t *testing.T) { })` wrapped around the code in this loop (which helps it log your name field)

suprjinx (approved):

PR #134, opened 2023-06-23
Ability to delete experiments in Aim UI

fabiovincenzi:
This PR only implements the backend side. The UI side is implemented in https://github.com/G-Research/fasttrackml-ui-aim/pull/6.

PR #211, opened 2023-08-04
Fix database migrations that recreate table twice and delete run data on SQLite

jgiannuzzi:
In #72 and #134, we introduced new database migrations. Unfortunately they had issues on SQLite 😞

#72 is recreating multiple tables twice, one of them being the one containing all the metrics, which takes quite a bit of time. This PR changes the migration to recreate the tables only once. Ideally we wouldn't need to recreate any tables, but this is a very unfortunate limitation of SQLite.

#134 is deleting all the run data (metrics, params, and tags) because the foreign key constraints were not disabled and the run table got recreated. This PR temporarily disables foreign key constraints for SQLite when running this particular migration to avoid the cascading delete.

### DIFF
--- a/pkg/database/database.go
+++ b/pkg/database/database.go
@@ -229,6 +229,13 @@ func checkAndMigrateDB(db *DbInstance, migrate bool) error {
 			return fmt.Errorf("unsupported database schema versions alembic %s, FastTrackML %s", alembicVersion.Version, schemaVersion.Version)
 		}
 
+		runWithoutForeignKeyIfPossible := func(fn func() error) error { return fn() }
+		switch db.Dialector.Name() {
+		case "sqlite":
+			migrator := db.Migrator().(sqlite.Migrator)
+			runWithoutForeignKeyIfPossible = migrator.RunWithoutForeignKey
+		}
+
 		switch alembicVersion.Version {
 		case "c48cb773bb87":
 			log.Info("Migrating database to alembic schema bd07f7e963c5")
@@ -410,8 +417,12 @@ func checkAndMigrateDB(db *DbInstance, migrate bool) error {
 				if err := db.Transaction(func(tx *gorm.DB) error {
 					constraints := []string{"Params", "Tags", "Metrics", "LatestMetrics"}
 					for _, constraint := range constraints {
-						if err := tx.Migrator().DropConstraint(&Run{}, constraint); err != nil {
-							return err
+						// SQLite tables need to be recreated to add or remove constraints.
+						// By not dropping the constraint, we can avoid having to recreate the table twice.
+						if db.Dialector.Name() != "sqlite" {
+							if err := tx.Migrator().DropConstraint(&Run{}, constraint); err != nil {
+								return err
+							}
 						}
 						if err := tx.Migrator().CreateConstraint(&Run{}, constraint); err != nil {
 							return err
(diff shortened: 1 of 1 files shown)


dsuhinin (approved):

pkg/database/database.go
```diff
@@ -229,6 +229,13 @@ func checkAndMigrateDB(db *DbInstance, migrate bool) error {
 			return fmt.Errorf("unsupported database schema versions alembic %s, FastTrackML %s", alembicVersion.Version, schemaVersion.Version)
 		}
 
+		runWithoutForeignKeyIfPossible := func(fn func() error) error { return fn() }
```
suprjinx, line 232: this function doesn't seem to do anything except wrap another function?

pkg/database/database.go
```diff
@@ -428,22 +439,33 @@ func checkAndMigrateDB(db *DbInstance, migrate bool) error {
...
-					constraints := []string{"Tags", "Runs"}
-					for _, constraint := range constraints {
-						if err := tx.Migrator().DropConstraint(&Experiment{}, constraint); err != nil {
-							return err
-						}
-						if err := tx.Migrator().CreateConstraint(&Experiment{}, constraint); err != nil {
-							return err
+				// We need to run this migration without foreign key constraints to avoid
+				// the cascading delete to kick in and delete all the run data.
+				if err := runWithoutForeignKeyIfPossible(func() error {
```
suprjinx, line 444: this is crazy, dropping the constraint causes a delete-cascade??

pkg/database/database.go
jgiannuzzi: yes, by default it doesn't do anything — it only does something on SQLite

pkg/database/database.go
jgiannuzzi: most table alterations in SQLite are not possible and thus require recreating the table — if you recreate a table that is referenced by other tables with `ON DELETE CASCADE`, then their content will be deleted 😭

suprjinx: that's rough, thanks for catching!

suprjinx: so this means we cannot migrate sqlite to have delete-cascades on those associations, yes? Do we need to reopen that ticket and manually clear associations?

pkg/database/database.go
suprjinx: maybe the name is confusing to me -- perhaps "disableForeignKeysFunc"? I assume for PG this is possible, just not needed?

pkg/database/database.go
jgiannuzzi: no no, the `CreateConstraint` is enough to modify the constraint (hence why we skip the `DropConstraint` in SQLite and avoid recreating twice.
and we _do_ want to have `ON DELETE CASCADE` on those associations, but we want to disable it temporarily whilst migrating some tables.

pkg/database/database.go
jgiannuzzi: I called it like that because the function with use with the SQLite migrator is called `RunWithoutForeignKey`

jgiannuzzi: it's not necessary for Postgres, as generally speaking we do _not_ want to disable foreign key constraints, even whilst migrating, but SQLite's minimal support of `ALTER TABLE` forces us to do that

suprjinx: "...IfNeeded"?

jgiannuzzi: addressed in https://github.com/G-Research/fasttrackml/pull/211/commits/f31805448f24abf095ee0b4dc6c8cbae55757c8f

PR #214, opened 2023-08-07
Python wheels

### LINKED ISSUE #81: Add Python wheels
┆Issue is synchronized with this [Jira Task](https://gr-oss.atlassian.net/browse/FML-61) by [Unito](https://www.unito.io)

fabiovincenzi:
Fixes #81


jgiannuzzi (commented):
`cibuildwheel` is a very heavy-weight process that is only useful when building libraries that depend on Python internals. In our case we just want to copy our binary, so we would be better off having a simple `make python-dist` that does something like `python -m pip wheel . --wheel-dir=wheelhouse --no-deps`.

We can then just have `dist` depend on this target, like it already does depend on `go-dist`, and change our golang-build job to be a generic dist job.

We also need to make sure versioning works (proper parsing of `fml --version`, and that the wheels are not tied to a particular version of Python (so `py3-none`) — I think the `ext_modules` and `build_ext` stuff is unnecessary and triggers that.

In the end we should have 5 wheels that are named something like this:
```
fml-0.2.0-py3-none-manylinux1_x86_64.whl
fml-0.2.0-py3-none-manylinux1_aarch64.whl
fml-0.2.0-py3-none-macosx_x86_64.whl
fml-0.2.0-py3-none-macosx_arm64.whl
fml-0.2.0-py3-none-win_amd64.whl
```

suprjinx (approved):
LGTM

.github/workflows/ci.yml
```diff
@@ -190,6 +190,64 @@ jobs:
...
+        run: make python-dist
+        env:
+          GOOS: ${{ matrix.os }}
+          GOARCH: ${{ matrix.arch }}
+
+      - uses: actions/upload-artifact@v3
+        with:
+          name: fasttrackml-wheels
+          path: ./wheelhouse/*.whl
+
```
jgiannuzzi, line 250: Change `golang-build` to become `build` and run `make dist` that will depend on `python-dist` as well as `go-dist`. This way each os/arch combination will build its own binary and wheel in one go.

.github/workflows/release.yml
```diff
@@ -49,6 +49,22 @@ jobs:
...
+      id-token: write
+    steps:
+      - name: Download artifact
+        run: gh run download ${{ github.event.workflow_run.id }} --repo ${{ github.event.workflow_run.repository.full_name }} --name fasttrackml-wheels --dir wheelshouse
+        env:
+          GH_TOKEN: ${{ github.token }}
+      - name: Publish package distributions to PyPI
+        uses: pypa/gh-action-pypi-publish@release/v1
+        with:
+          packages-dir: wheelshouse/
```
jgiannuzzi, line 67:
```suggestion
        run: gh run download ${{ github.event.workflow_run.id }} --repo ${{ github.event.workflow_run.repository.full_name }} --name fasttrackml-wheels --dir wheelhouse
        env:
          GH_TOKEN: ${{ github.token }}
      - name: Publish package distributions to PyPI
        uses: pypa/gh-action-pypi-publish@release/v1
        with:
          packages-dir: wheelhouse/
```

Makefile
```diff
@@ -39,6 +39,18 @@ endif
...
+    PLATFORM_NAME := macosx_10_13_x86_64
+else ifeq ($(shell go env GOOS)_$(shell go env GOARCH),darwin_arm64)
+    PLATFORM_NAME := macosx_11_0_arm64
+else ifeq ($(shell go env GOOS)_$(shell go env GOARCH),linux_x86_64)
+    PLATFORM_NAME := manylinux1_x86_64
+else ifeq ($(shell go env GOOS)_$(shell go env GOARCH),linux_arm64)
+    PLATFORM_NAME := manylinux1_aarch64
+else ifeq ($(shell go env GOOS)_$(shell go env GOARCH),windows_x86_64)
+    PLATFORM_NAME := win_amd64
+endif
```
jgiannuzzi, line 53: do this within `setup.py` to return the proper `plat` in `bdist_wheel.get_tag()`

setup.py
```diff
@@ -0,0 +1,68 @@
...
+def get_version():
+    path = './dist/fasttrackml_linux_x86_64' if platform.system() == "Linux" else './dist/fasttrackml_macos_x86_64'
+    bin = tarfile.open(f'{path}.tar.gz')
+    bin.extractall(path)
+    bin.close()
+    version_string = subprocess.check_output(
+        [f"{path}/fml", "--version"], universal_newlines=True
+    ).strip()
+    matches = re.findall(r'\d+\.\d+\.\d+', version_string)
+    return matches[0]
```
jgiannuzzi, line 23: Use the `VERSION` environment variable (with a fallback to `'dev'`)

Makefile
```diff
@@ -81,6 +93,18 @@ go-dist: go-build ## archive app binary.
 	@if [ -f $(ARCHIVE_NAME) ]; then rm -f $(ARCHIVE_NAME); fi
 	@$(ARCHIVE_CMD) $(ARCHIVE_NAME) $(ARCHIVE_FILES)
 
+.PHONY: python-dist
+python-dist: ## build python wheels.
```
jgiannuzzi, line 97: depend on `go-build`

jgiannuzzi: and make `dist` depend on this one

Makefile
```diff
@@ -81,6 +93,18 @@ go-dist: go-build ## archive app binary.
...
+	@echo '>>> Building Python Wheels.'
+	@echo $(VERSION)
+	@python3 -m pip wheel . --wheel-dir=tmp-wheels --no-deps
+	@mkdir -p wheelhouse
+	@for file in tmp-wheels/*.whl; do \
+		base_name="$$(basename $$file)"; \
+		new_name="$$(echo $$base_name | rev | cut -d- -f2- | rev)-$(PLATFORM_NAME).whl"; \
+		cp "$$file" "wheelhouse/$$new_name"; \
+	done
```
jgiannuzzi, line 106: scrap the renaming by handling the platform name within `setup.py`

setup.py
```diff
@@ -0,0 +1,68 @@
...
+
+def get_fml_executable():
+     for file_name in os.listdir("bin"):
+         if "fml" in file_name:
+             return os.path.join("bin", file_name)
```
jgiannuzzi, line 28: the executable is at the root of the repo

setup.py
```diff
@@ -0,0 +1,68 @@
...
+        def finalize_options(self):
+            _bdist_wheel.finalize_options(self)
+            # Mark us as not a pure python package
+            self.root_is_pure = False
+
+        def get_tag(self):
+            python, abi, plat = _bdist_wheel.get_tag(self)
+            # We don't contain any python source
+            python, abi = 'py3', 'none'
+            return python, abi, plat
```
jgiannuzzi, line 49: use the `GOOS` and `GOARCH` environment variables to determine which `plat` to return

pyproject.toml
jgiannuzzi, line 1: move the 2 Python files into their own `python` directory, ideally with `batch.py` as well

jgiannuzzi: _not for this PR, this can be done in another one_: make sure `batch.py` is included in the wheel, so that a user can do
```python
import fasttrackml

fasttrackml.get_metric_histories(...)
```

this will require adding a `__init__.py` file that exports `batch.get_metric_histories`.

jgiannuzzi (changes requested):
great stuff @fabiovincenzi

I have written down the comments that we discussed yesterday

setup.py
fabiovincenzi: I believe 'dev' is not a valid version for setuptools

python/setup.py
```diff
@@ -0,0 +1,66 @@
...
+
+    def get_tag(self):
+        python, abi, plat = _bdist_wheel.get_tag(self)
+        # We don't contain any python source
+        python, abi = 'py3', 'none'
+        return 'py3', 'none', get_platform()
+
+
+setup(
+    name="fml",
```
jgiannuzzi, line 52: package name should actually be `fasttrackml`
I should have realised this sooner, sorry!

jgiannuzzi (commented):
another round of comments 😅 

please also add the following files to `.gitignore`:
```
python/build/
python/*.egg-info/
wheelhouse/
```

.github/workflows/ci.yml
```diff
@@ -178,18 +178,24 @@ jobs:
           sudo apt-get install -y --no-install-recommends gcc-mingw-w64-x86-64-win32
           echo CC=x86_64-w64-mingw32-gcc >> $GITHUB_ENV
 
-      - name: Build and archive Go binary
-        run: make go-dist
+      - name: Build software distribution
+        run: |
+          make dist
```
jgiannuzzi, line 183: put it on one line please

.github/workflows/ci.yml
```diff
@@ -178,18 +178,24 @@ jobs:
...
         env:
           GOOS: ${{ matrix.os }}
           GOARCH: ${{ matrix.arch }}
-
-      - name: Upload artifact
+      - name: Upload binary artifact
```
jgiannuzzi, line 187: keep the empty line above (you had a line with spaces before)

Makefile
```diff
@@ -81,6 +81,14 @@ go-dist: go-build ## archive app binary.
 	@if [ -f $(ARCHIVE_NAME) ]; then rm -f $(ARCHIVE_NAME); fi
 	@$(ARCHIVE_CMD) $(ARCHIVE_NAME) $(ARCHIVE_FILES)
 
+#
+# Python targets.
+#
+.PHONY: python-dist
+python-dist: go-build ## build python wheels.
+	@echo '>>> Building Python Wheels.'
+	@@export VERSION=$(VERSION); python3 -m pip wheel ./python --wheel-dir=wheelhouse --no-deps
```
jgiannuzzi, line 90:
```suggestion
	@VERSION=$(VERSION) python3 -m pip wheel ./python --wheel-dir=wheelhouse --no-deps
```

python/setup.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+def get_fml_executable():
+    os = subprocess.check_output(["go", "env", "GOOS"]).strip().decode("utf-8")
+    return "../fml.exe" if os == "windows" else "../fml"
+
+def get_version():
+    version = os.environ.get("VERSION")
+    matches = re.findall(r'\d+\.\d+\.\d+', version)
+    return matches[0]
```
jgiannuzzi, line 21:
```suggestion
    return version.replace("-", "+", 1)
```

python/setup.py
```diff
@@ -0,0 +1,66 @@
...
+    arch = subprocess.check_output(["go", "env", "GOARCH"]).strip().decode("utf-8")
+    plat = f"{os}_{arch}"
+    if plat == "darwin_amd64":
+        return "macosx_10_13_x86_64"
+    elif plat == "darwin_arm64":
+        return "macosx_11_0_arm64"
+    elif plat == "linux_amd64":
+        return "manylinux1_x86_64"
+    elif plat == "linux_arm64":
+        return "manylinux1_aarch64"
```
jgiannuzzi, line 34: just checked this locally and this is required to install properly everywhere:
```suggestion
    elif plat == "linux_amd64":
        return "manylinux_2_17_x86_64.manylinux2014_x86_64.musllinux_1_1_x86.64"
    elif plat == "linux_arm64":
        return "manylinux_2_17_aarch64.manylinux2014_aarch64.musllinux_1_1_aarch64"
```

fabiovincenzi: I think this is not a valid name for a wheel

setup.py
jgiannuzzi: you're right!

python/setup.py
jgiannuzzi: correct, I edited the suggestion above

jgiannuzzi (approved):
Awesome, thanks @fabiovincenzi!

dsuhinin (approved):

PR #215, opened 2023-08-07
Implement `IN` and `NOT IN` string comparison function in Aim query

dsuhinin:


pkg/api/aim/query/query.go
```diff
@@ -364,13 +342,36 @@ func (pq *parsedQuery) parseCompare(node *ast.Compare) (any, error) {
...
-				o, l, r, err := reverseComparison(op, left, right)
-				if err != nil {
-					return nil, err
-				}
-				exprs[i], err = newSqlComparison(o, l, r)
-				if err != nil {
-					return nil, err
+				switch op {
+				case ast.In:
+					return clause.Like{
```
jgiannuzzi, line 347: we should probably check that `left` is a string in those 2 cases

jgiannuzzi (commented):
I like it!!!

just a couple of comments below

pkg/api/aim/query/query.go
dsuhinin: done.

pkg/api/aim/query/query.go
```diff
@@ -364,13 +342,36 @@ func (pq *parsedQuery) parseCompare(node *ast.Compare) (any, error) {
...
+						Value: fmt.Sprintf("%%%s%%", left),
+						Column: clause.Column{
+							Table: right.Table,
+							Name:  right.Name,
+						},
+					}, nil
+				case ast.NotIn:
+					return clause.NotConditions{
+						Exprs: []clause.Expression{
```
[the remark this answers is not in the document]

pkg/api/aim/query/query.go
```diff
@@ -38,12 +39,20 @@ type parsedQuery struct {
 	conditions []clause.Expression
 }
 
-type attributeGetter func(attr string) (any, error)
-
 type callable func(args []ast.Expr) (any, error)
 
+var negativeClause = func(expression clause.Expression) clause.Expression {
```
jgiannuzzi, line 44: why do you make it a func variable and not directly a func? 🤔

pkg/api/aim/query/query.go
```diff
@@ -344,25 +353,23 @@ func (pq *parsedQuery) parseCompare(node *ast.Compare) (any, error) {
 			case clause.Column:
 				switch op {
 				case ast.In:
+					// for `IN` statement, left parameter has to be always `string`.
+					if _, ok := left.(string); !ok {
+						return nil, eris.New("left parameter has to be a string")
```
jgiannuzzi, line 358: why do you use `eris.New` instead of the stdlib `errors.New` here?

jgiannuzzi: can you change all the `errors.New` and `fmt.Errorf` of this file to use `eris` then?

dsuhinin: ok, I reverted to use `errors` instead. lets change this when we will have good unit tests for such important module.

pkg/api/aim/query/query.go
dsuhinin: changed in this style.

pkg/api/aim/query/query.go
```diff
@@ -38,12 +38,12 @@ type parsedQuery struct {
...
 }
 
-type attributeGetter func(attr string) (any, error)
-
 type callable func(args []ast.Expr) (any, error)
 
 type subscriptSlicer func(index ast.Slicer) (any, error)
 
+type attributeGetter func(attr string) (any, error)
+
```
jgiannuzzi, line 46: any reason why you moved `attributeGetter` here?

jgiannuzzi (commented):
just one last nitpick, but otherwise LGTM 👍

pkg/api/aim/query/query.go
jgiannuzzi: hmmm, should it not be before `subscriptSlicer` then?

dsuhinin: no no, not alphabetical - it is pain, just from short to long.

dsuhinin: but yep, lets sort it fully.

jgiannuzzi: but they have the same length for their names, and then `subscriptSlicer` has a longer signature — I'm cool with going with non-alphabetical and go for length, but we need to be consistent!

jgiannuzzi (approved):

pkg/api/aim/query/query.go
dsuhinin: yep, the use this way. agreed. done.

Issue #135, opened 2023-06-23
Add new logo in all the UIs


jgiannuzzi:
We have essentially 2 types of logos: the "full logo" (aka project-logo) and the "icon logo" (aka project-icon). They can be found here: https://github.com/G-Research/fasttrackml/tree/main/website/static/logo.
The new logos should be used in all the following places:

- [x] chooser UI (full + favicon) - dark mode + light mode
  - [x] https://github.com/G-Research/fasttrackml/pull/276
- [x] MLFlow UI (full + favicon)
  - [x] https://github.com/G-Research/fasttrackml-ui-mlflow/pull/9
  - [x] https://github.com/G-Research/fasttrackml/pull/253
- [x] Aim UI (icon + spinner + favicon)
  - [x] https://github.com/G-Research/fasttrackml-ui-aim/pull/20
  - [x] https://github.com/G-Research/fasttrackml-ui-aim/pull/21
  - [x] https://github.com/G-Research/fasttrackml/pull/278

suprjinx:
For MLH fellow

PR #216, opened 2023-08-07
Add full logo and favicon to Chooser UI

galadd:
This PR fixes 1 of the 3 main parts of #135

### DIFF
--- a/pkg/ui/chooser/embed/index.html
+++ b/pkg/ui/chooser/embed/index.html
@@ -16,11 +16,21 @@
             grid-template-columns: 1fr 1fr;
             column-gap: 2rem;
         }
+
+        .logo {
+            display: block;
+            width: 250px;
+            height: auto;
+        }
     </style>
+
+    <link rel="icon" type="image/x-icon" href="../../../../website/static/favicon.ico">
 </head>
 
 <body>
     <header>
+        <img class="logo" src="../../../../website/static/logo/project-logo-dark.svg" alt="full logo">
+
         <h1><i>FastTrackML</i></h1>
         <p>A <i>very fast</i> experiment tracker</p>
     </header>


pkg/ui/chooser/embed/index.html
```diff
@@ -16,11 +16,21 @@
...
         }
+
+        .logo {
+            display: block;
+            width: 250px;
+            height: auto;
+        }
     </style>
+
+    <link rel="icon" type="image/x-icon" href="../../../../website/static/favicon.ico">
```
jgiannuzzi, line 27: those files would actually need to be copied over so that they can be embedded in the Go binary

pkg/ui/chooser/embed/index.html
```diff
@@ -16,11 +16,21 @@
...
 </head>
 
 <body>
     <header>
+        <img class="logo" src="../../../../website/static/logo/project-logo-dark.svg" alt="full logo">
```
jgiannuzzi, line 32: the theme we use can adapt to both dark and light mode — could you please ensure the right icon gets used for the right mode?

pkg/ui/chooser/embed/index.html
galadd: Ah yes! that was an oversight

galadd: Please, can you elaborate more on what you mean by copy over

pkg/ui/chooser/embed/index.html
galadd: Presently, the chooser ui does not have a light mode

jgiannuzzi: oh it totally does!
Dark:
image
Light:
image

pkg/ui/chooser/embed/index.html
jgiannuzzi: I mean that we shouldn't reference files in the repo, but instead reference files that will be embedded in the Go binary, like `simple.min.css`

galadd: Like the aim and mlflow ui src/src, does the chooser ui already have a directory for modifications. I presently can’t find an existing one in this repo (fasttrackml)

pkg/ui/chooser/embed/index.html
galadd: Interesting. I can’t seem to figure this out on my own end. There is no visible toggle. How are you changing it?

jgiannuzzi: by changing the mode at the OS level

pkg/ui/chooser/embed/index.html
jgiannuzzi: yeah the other UIs require "building" and so that is why it's then converted — the chooser UI is (so far) just plain HTML/CSS/JS, so things have to be dropped in the right place instead

pkg/ui/chooser/embed/logo/organization-dark.svg
jgiannuzzi, line 1: many of those files are not required — please only keep the favicon and the dark and light logos

pkg/ui/chooser/embed/simple.min.css
jgiannuzzi, line 1: Please don't modify `simple.min.css`

jgiannuzzi (changes requested):

pkg/ui/chooser/embed/index.js
jgiannuzzi, line 1: please remove this file

pkg/ui/chooser/embed/index.html
```diff
@@ -16,15 +17,35 @@
             grid-template-columns: 1fr 1fr;
             column-gap: 2rem;
         }
+
+        header {
+            display: flex;
+            flex-direction: column;
+            align-items: center;
+            text-align: center;
```
jgiannuzzi, line 25: `text-align: center` is already defined in the CSS

pkg/ui/chooser/embed/index.html
```diff
@@ -16,15 +17,35 @@
...
+        }
+
+        .logo {
+            max-width: 100%;
```
jgiannuzzi, line 29: `max-width: 100%` is already defined in the CSS for `img`

pkg/ui/chooser/embed/index.html
```diff
@@ -16,15 +17,35 @@
...
+            height: 100%;
```
jgiannuzzi, line 30: let's go with `height: 150px;` to keep the logo quite impactful

jgiannuzzi: you'll probably need to go with something like `margin-top: -25px;` for `.logo` and `margin-top: -35px;` for `.title p` to have something similar to what we used to have

jgiannuzzi (commented):
This is very close to finished, well done @galadd!

Issue #213, opened 2023-08-04
Add golangci-lint


Smaug123:
There are a number of lint failures, of which these ones are pretty concerning because they indicate unchecked errors:

```
pkg/api/aim/encoding/encoder.go:86:16                                      errcheck     Error return value of `binary.Write` is not checked
pkg/api/aim/encoding/encoder.go:140:15                                     errcheck     Error return value of `binary.Write` is not checked
tests/integration/golang/helpers/client.go:96:23                           errcheck     Error return value of `resp.Body.Close` is not checked
tests/integration/golang/helpers/client.go:138:23                          errcheck     Error return value of `resp.Body.Close` is not checked
pkg/database/database.go:125:12                                            errcheck     Error return value of `DB.Close` is not checked
pkg/database/database.go:163:11                                            errcheck     Error return value of `DB.Close` is not checked
pkg/database/database.go:168:9                                             errcheck     Error return value of `DB.Use` is not checked
pkg/database/database.go:178:10                                            errcheck     Error return value of `DB.DB.DB` is not checked
pkg/database/database.go:185:13                                            errcheck     Error return value of `DB.Close` is not checked
pkg/api/mlflow/dao/convertors/helpers.go:10:5                              errcheck     Error return value of `rand.Int` is not checked
pkg/api/mlflow/dao/convertors/helpers.go:11:5                              errcheck     Error return value of `rand.Int` is not checked
pkg/api/mlflow/dao/convertors/helpers.go:12:5                              errcheck     Error return value of `rand.Int` is not checked
pkg/api/aim/query/clause.go:21:22                                          errcheck     Error return value of `builder.WriteString` is not checked
pkg/api/aim/query/clause.go:23:22                                          errcheck     Error return value of `builder.WriteString` is not checked
pkg/api/aim/query/clause.go:33:22                                          errcheck     Error return value of `builder.WriteString` is not checked
pkg/api/aim/runs.go:639:19                                                 errcheck     Error return value of `rows.Close` is not checked
pkg/api/aim/runs.go:841:19                                                 errcheck     Error return value of `rows.Close` is not checked
pkg/api/aim/runs.go:1039:15                                                errcheck     Error return value of `binary.Write` is not checked
pkg/ui/chooser/routes.go:17:7                                              errcheck     Error return value of `fs.Sub` is not checked
pkg/api/mlflow/controller/metric.go:73:19                                  errcheck     Error return value of `rows.Close` is not checked
[...]

PR #218, opened 2023-08-08
Improve linter configuration

jgiannuzzi:
This is first step towards addressing #213. The `golangci-lint` configuration is slightly adapted to include the linters already used in the CI pipeline, and the linting is also enabled in VSCode to help devs catch issues during development.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -4,5 +4,9 @@
     "gopls": {
         "formatting.local": "github.com/G-Research/fasttrackml",
         "formatting.gofumpt": true
-    }
+    },
+    "go.lintTool": "golangci-lint",
+    "go.lintFlags": [
+        "--fast"
+    ]
 }
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -23,8 +23,8 @@ run:
   # can use regexp here: generated.*, regexp is applied on full path;
   # default value is empty creators, but next dirs are always skipped independently
   # from this option's value:
-  #   	vendor$, third_party$, testdata$, examples$, Godeps$, builtin$
-  skip-dirs:
+  #   vendor$, third_party$, testdata$, examples$, Godeps$, builtin$
+  skip-dirs: []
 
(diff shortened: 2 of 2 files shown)


suprjinx (approved):

.vscode/settings.json
```diff
@@ -4,5 +4,9 @@
     "gopls": {
         "formatting.local": "github.com/G-Research/fasttrackml",
         "formatting.gofumpt": true
-    }
+    },
+    "go.lintTool": "golangci-lint",
+    "go.lintFlags": [
+        "--fast"
```
dsuhinin, line 10: maybe just add `-v` for more information?

dsuhinin (approved):
Looks good!

.vscode/settings.json
dsuhinin: also -> ` --tests Analyze tests (*_test.go) (default true)` will be nice to disable `test` linting, because sometimes it could be a pain to follow all the rules, buuut. what do you think?

dsuhinin: also, thinking about next linters to enable:
- [prealloc](https://github.com/alexkohler/prealloc)
- [varnamelen](https://github.com/blizzy78/varnamelen)
- [nilerr](https://github.com/gostaticanalysis/nilerr)
- [nestif](https://github.com/nakabonne/nestif)
- [misspell](https://github.com/client9/misspell)
- [ireturn](https://github.com/butuzov/ireturn)
- [gomnd](https://github.com/tommy-muehle/go-mnd)
- [godot](https://github.com/tetafro/godot)

jgiannuzzi: `-v` does not seem to make a difference in VS Code. In both cases, linting errors look like this:
image

jgiannuzzi: about `--tests`, I think let's keep it for now — ideally we do want to follow all the rules, but we'll see once we actually enforce it how it goes!

PR #220, opened 2023-08-08
Namespaces

dsuhinin:


pkg/api/mlflow/dao/repositories/experiment.go
```diff
@@ -60,10 +66,25 @@ func (r ExperimentRepository) Create(ctx context.Context, experiment *models.Exp
...
+			return nil, eris.Wrapf(err, "error getting default experiment for namespace code %q", nsCode)
+		}
+	}
 	var experiment models.Experiment
-	if err := r.db.WithContext(ctx).Where(
+	if err := r.db.WithContext(ctx).Preload(
+		"Tags",
+	).Joins(
+		"RIGHT JOIN namespaces ON namespaces.id = experiments.namespace_id AND namespaces.code = ?",
+		namespace.GetCodeFromContext(ctx),
```
dsuhinin, line 84: Why when we create `experiment` we use -> `experiment.NamespaceID, err = namespace.GetIDFromContext(r.db, ctx)` but here we use `namespace.GetCodeFromContext(ctx)`? Can't we use just `namespace.GetIDFromContext(r.db, ctx)` ? it is a bit confusing that we are using different functions

dsuhinin: so it will be like:
```
RIGHT JOIN namespaces ON namespaces.id = experiments.namespace_id AND namespaces.id = ?
```

dsuhinin: or even, why do we need to join each time to validate that this `namespace` is exists? we've already validated that namespace` is already exists in `middleware` layer, so here we can just trust without any risk like:
```
WHERE experiments.namespace_id = ?
```

PR #233, opened 2023-08-10
Use ruby image + sequel gem for DB import/export

suprjinx:
Fixes #200


suprjinx:
not practical for users

PR #241, opened 2023-08-14
Check how we create the required check by using a separate workflow

jgiannuzzi:
The "All required checks done" CI job has been designed to be a required check, but GitHub considers it successful when it's skipped, and it gets skipped when some of the jobs it depends on fail. A first attempt at fixing that was made in #237 but it was not enough.

This PR fixes that by checking in a separate workflow whether the job conclusion is `success` and creates a new check run "All required checks **succeeded**" when it's the case. When it's not the case, the check run is not created and PRs will thus not be mergeable.

Example runs:
- [successful run](https://github.com/jgiannuzzi/fasttrackml/actions/runs/5858130823) — the [check run](https://github.com/jgiannuzzi/fasttrackml/runs/15881484894) is created by this [workflow run](https://github.com/jgiannuzzi/fasttrackml/actions/runs/5858187124/job/15881638649#step:2:47)
- [skipped run](https://github.com/jgiannuzzi/fasttrackml/actions/runs/5858128921) — the check run was _not_ created by this [workflow run](https://github.com/jgiannuzzi/fasttrackml/actions/runs/5858130484/job/15881468808#step:2:47)
- [failed run](https://github.com/jgiannuzzi/fasttrackml/actions/runs/5858124644) — the check run was _not_ created by this [workflow run](https://github.com/jgiannuzzi/fasttrackml/actions/runs/5858128916/job/15881462725#step:2:47)

I changed the workflow output slightly to show a ❌ when "All required checks done" is not successful, but this is not reflected in the above examples.

### DIFF
--- a/.github/workflows/check-required.yml
+++ b/.github/workflows/check-required.yml
@@ -1,0 +1,62 @@
+name: Check required jobs
+
+# This workflow is triggered when a workflow run for the CI is completed.
+# It checks if the "All required checks done" job was actually successful
+# (and not just skipped) and creates a check run if that is the case. The
+# check run can be used to protect the main branch from being merged if the
+# CI is not passing.
+
+on:
+  workflow_run:
+    types: [completed]
+    workflows: [CI]
+
+permissions:
+  actions: read
+  checks: write
+
+jobs:
+  required-jobs:
+    name: Check required jobs
+    runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


suprjinx (approved):
It seems a little cuckoo that Github workflows don't stop on a failure without such a script -- isn't that CI 101?

PR #251, opened 2023-08-17
Python wheels fixes

jgiannuzzi:
- Set README type to Markdown
- Only publish to PyPI for tags


dsuhinin (approved):

suprjinx (approved):
So this looks like it will release for all branches but main, yes? including WIP branches?

jgiannuzzi:
> So this looks like it will release for all branches but main, yes? including WIP branches?

No, this is actually the opposite :)
This workflow only runs for `main` and tags, and we exclude `main` for publishing to PyPI. The result is that we will only publish to PyPI for tags.

PR #258, opened 2023-08-18
Installation script

### LINKED ISSUE #198: Add an installation script to the website
We should add a script to the website that can be curled and piped into a shell to install the latest FML binary for the host platform.
On Linux and macOS, it should be a bash script. On Windows, this should be a PowerShell script.
Have a look at how they do it at https://github.com/ducaale/xh. Their install scripts should be an excellent starting point.

- [ ] create bash script and add it to the website structure (should end up as https://fasttrackml.io/install.sh)
- [ ] create PowerShell script and it to the website structure (should end up as https://fasttrackml.io/install.ps1)
- [ ] adapt website quickstart to have the command to run the script — make it copyable!
  - [ ] "Run FastTrackML" should become 3 steps: "Install FastTrackML", "Run FastTrackML", and "Run FastTrackML via Docker"
  - [ ] "Install FastTrackML" should reference how to install on the supported platforms:
    - [ ] `curl -fsSL https://fasttrackml.io/install.sh | sh` on Linux and macOS
    - [ ] `iwr -useb https://fasttrackml.io/install.ps1 | iex` on Windows
  - [ ] "Run FastTrackML" should be the command to run the installed binary
  - [ ] "Run FastTrackML via Docker" should be the content of the original "Run FastTrackML" section

fabiovincenzi:
fixes #198


website/static/install.sh
```diff
@@ -0,0 +1,87 @@
...
+        tac | tac | grep -wo -m1 "https://.*$target.tar.gz" || true
+)
+if ! test "$url"; then
+    echo "Could not find release info"
+    exit 1
+fi
+
+echo "Downloading fml..."
+
+temp_dir=$(mktemp -dt fml.XXXXXX)
```
suprjinx, line 42: possibly some content filters might block the script from download because of this name string. maybe just "fml.tmp"?

website/static/install.sh
```diff
@@ -0,0 +1,87 @@
...
+
+set -e
+
+if [ "$(uname -s)" = "Darwin" ] && [ "$(uname -m)" = "x86_64" ]; then
+    target="macos_x86_64"
+elif [ "$(uname -s)" = "Linux" ] && [ "$(uname -m)" = "x86_64" ]; then
+    target="linux_x86_64"
+elif [ "$(uname -s)" = "Linux" ] && [ "$(uname -m)" = "arm64" ]; then
+    target="linux_arm64"
+elif [ "$(uname -s)" = "Darwin" ] && [ "$(uname -m)" = "arm64" ]; then
```
suprjinx, line 11: move the two Darwin cases next to each other?

suprjinx (approved):
Looks great!

website/static/install.sh
jgiannuzzi: the `X`'s are actually required by the GNU coreutils version of `mktemp`, so changing the _template_ to `fml.tmp` makes the script now fail on Linux

jgiannuzzi: also a content filter that looks at `XXXXXX` within a TLS stream and decides to block it would most likely block the download of our binary as well

PR #259, opened 2023-08-18
Add initial support for namespaces

dsuhinin:
What's inside the box:
- `namespace` functionality fully integrated into `mlflow` part.
- `mlflow` `integration` and `unit` tests adopted for `namespace` logic.
- `namespace` functionality integrated into `aim` part. not fully. needs attention from @jgiannuzzi 
- `aim` `integration` tests  adopted for `namespace` logic.
- introduced basic `admin` routing.
- moved `db` models to `common` package, since it will be shared across `mlflow` and `aim` apis.
- refactored `middleware` logic to catch current `namespace`.
- redesign/simplify `integration` tests.
- started fixing minor problems across the project -> `golangci-lint`.

What's left(we can complete it in other tasks a bit later):
- add more `unit`, `integration` tests related to `namespace` functionality.
- as I can see our `repositories` started receiving a lot of input `parameters`. sometimes it could be 3, sometimes event 5. It doesn't look neat. it will be nice to introduce `dto` objects and pass this objects right into `repositories` functions in case if for example function waits more then 3-4 parameters. Also, I think it could help and remove some functions which looks almost similar but differ in one or more parameters. So we can remove duplicate functions.
- `aim` api, needs refactoring and `integration` tests. it is hard to support, read and integrate something new.
- fully fix all the `golangci-lint` remarks, so we can integrate this step as a required step into our `pipeline`.


pkg/api/mlflow/api/response/metric.go
```diff
@@ -1,7 +1,7 @@
 package response
 
 import (
-	"github.com/G-Research/fasttrackml/pkg/api/mlflow/dao/models"
+	"github.com/G-Research/fasttrackml/pkg/common/dao/models"
```
suprjinx, line 4: Nice refactor -- I'd say we don't even need "common/dao" in path and could just place under pkg, eg `pkg/models`, `pkg/repositories`, etc

pkg/api/aim/experiments.go
```diff
@@ -5,7 +5,8 @@ import (
 	"strconv"
 	"time"
 
-	"github.com/G-Research/fasttrackml/pkg/api/mlflow/dao/models"
+	"github.com/G-Research/fasttrackml/pkg/common/dao/models"
```
suprjinx, line 8: Like Dave was saying, if you just made a preliminary PR of the package rename, it would reduce the number of files changed in this PR and both would be easier to review

.gitignore
```diff
@@ -10,6 +10,8 @@ fasttrackml
 # Local database
 fasttrackml.db*
 encrypted.db*
+fasttrackml-in.*
```
dsuhinin, line 13: this is for `import` integration test.

dsuhinin:
@jgiannuzzi can you please take a look into `aim` part, into this functions:
- `SearchRuns` function -> `414` line.
- `SearchMetrics` function -> `675` line.
- `SearchAlignedMetrics` function -> `871` line.
these are the places where I'm not at all sure.

pkg/api/admin/routes.go
```diff
@@ -0,0 +1,42 @@
...
+
+// NewRouter creates new instance of `mlflow` router.
+func NewRouter(controller *controller.Controller) *Router {
+	return &Router{
+		controller: controller,
+	}
+}
+
+// Init makes initialization of all `mlflow` routes.
+func (r Router) Init(server fiber.Router) {
```
dave-gantenbein, line 33: Can we at least get an integration test here?

pkg/api/aim/dashboards.go
```diff
@@ -5,16 +5,27 @@ import (
...
 func GetDashboards(c *fiber.Ctx) error {
+	ns, err := namespace.GetNamespaceFromContext(c.Context())
+	if err != nil {
+		return api.NewInternalError("error getting namespace from context")
+	}
+	log.Debugf("getDashboards namespace: %s", ns.Code)
+
 	var dashboards []database.Dashboard
 	if err := database.DB.
+		Preload("App").
```
dave-gantenbein, line 25: This seems like more mixing of model and controller. While it was here already, maybe now is a good time to abstract dashboards and apps out? It doesn't seem like a controller responsible for rendering json should be talking to a database...

pkg/api/aim/dashboards.go
```diff
@@ -5,16 +5,27 @@ import (
...
+	log "github.com/sirupsen/logrus"
 	"gorm.io/gorm"
 
+	"github.com/G-Research/fasttrackml/pkg/api/mlflow/api"
+	"github.com/G-Research/fasttrackml/pkg/common/middleware/namespace"
 	"github.com/G-Research/fasttrackml/pkg/database"
 )
 
 func GetDashboards(c *fiber.Ctx) error {
+	ns, err := namespace.GetNamespaceFromContext(c.Context())
```
dave-gantenbein, line 17: Does golang/fiber provide any sort of before hook that would let us DRY this up?

pkg/api/mlflow/api/response/metric.go
dave-gantenbein: +1

pkg/api/mlflow/dao/repositories/metric.go
```diff
@@ -26,10 +26,15 @@ type MetricRepositoryProvider interface {
 	// GetMetricHistories returns metric histories by request parameters.
 	GetMetricHistories(
 		ctx context.Context,
-		experimentIDs []string, runIDs []string, metricKeys []string, viewType request.ViewType, limit int32,
+		namespaceID uint,
```
dave-gantenbein, line 29: Don't we want to test these repositories?

pkg/api/mlflow/dao/repositories/run.go
```diff
@@ -89,13 +93,43 @@ func (r RunRepository) GetByIDAndLifecycleStage(
...
 			return nil, nil
 		}
-		return nil, eris.Wrapf(err, "error getting 'run' entity by id: %s", id)
+		return nil, eris.Wrapf(err, "error getting 'run' entity by id: %s", runID)
+	}
+	return &run, nil
+}
+
+// GetByNamespaceIDAndRunID returns models.Run entity by Namespace ID and its ID.
+func (r RunRepository) GetByNamespaceIDAndRunID(
```
dave-gantenbein, line 112: Testme.

pkg/api/aim/experiments.go
```diff
@@ -86,6 +103,7 @@ func GetExperiment(c *fiber.Ctx) error {
 		).
 		Joins("LEFT JOIN runs USING(experiment_id)").
 		Group("experiments.experiment_id").
+		Where("experiments.namespace_id = ?", ns.ID).
```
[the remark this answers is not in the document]
dsuhinin: yes, we still have have such call over `aim` because it isn't yet refactored properly. so nothing to do  right now.

pkg/api/aim/projects.go
```diff
@@ -20,18 +23,40 @@ func GetProject(c *fiber.Ctx) error {
...
+	log.Debugf("getProjectActivity namespace: %s", ns.Code)
+
 	tzOffset, err := strconv.Atoi(c.Get("x-timezone-offset", "0"))
 	if err != nil {
 		return fiber.NewError(fiber.StatusUnprocessableEntity, "x-timezone-offset header is not a valid integer")
 	}
 
 	var numExperiments int64
-	if tx := database.DB.Model(&database.Experiment{}).Where("lifecycle_stage = ?", database.LifecycleStageActive).Count(&numExperiments); tx.Error != nil {
+	if tx := database.DB.Model(
```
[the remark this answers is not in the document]
dsuhinin: yes unfortunately. we have to refactor it and split everything to be on different levels.

pkg/api/admin/dao/repositories/namespace.go
```diff
@@ -0,0 +1,53 @@
+package repositories
```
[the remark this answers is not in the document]
dsuhinin: we don't cover with tests `db` layer. there is no need to do that. we have `integration` tests which tests all level at once.

pkg/api/admin/routes.go
dsuhinin: yes we can. when we create actual logic for that we can add `integration` tests as well.

pkg/api/aim/dashboards.go
dsuhinin: yes, we have to refactor `aim` and make it similar to what we have in `mlflow`. right now `aim` is in bad shape.

pkg/api/aim/dashboards.go
dsuhinin: no. this is how we can attach `middleware`.

dsuhinin: no, this is usual approach to get something from the context.

pkg/api/mlflow/dao/repositories/metric.go
dsuhinin: no, we usually don't need to test `repositories`. `integration` tests do this for us.

pkg/api/mlflow/dao/repositories/run.go
dsuhinin: no, we usually don't need to test repositories. integration tests do this for us.

pkg/common/dao/models/namespace.go
```diff
@@ -0,0 +1,19 @@
...
+
+	"gorm.io/gorm"
+)
+
+// Namespace represents model to work with `namespaces` table.
+type Namespace struct {
+	ID                  uint   `gorm:"primaryKey;autoIncrement"`
+	Code                string `gorm:"unique;index;not null"`
+	Description         string
+	CreatedAt           time.Time
```
suprjinx, line 14: We have quite a mix in this project, but I think timestamps are usually stored in bigint for the other models (int64 or sql.NullInt64). should we do that here for consistency?

pkg/api/mlflow/api/response/metric.go
suprjinx: additionally, we need to move `api/mlflow/dao/repositories` to a common location

pkg/api/admin/routes.go
suprjinx: this is covered by integration tests which use the routes, right?

pkg/common/dao/models/namespace.go
dsuhinin: that's true. we also have a mix of types for `ids` lie: `uuid`, `string`, `uint`, etc. will be nice to normalise this moment as well.

suprjinx (dismissed):
Let's merge it

suprjinx (dismissed):
reapprove

pkg/database/migrate.go
```diff
@@ -306,6 +309,40 @@ func checkAndMigrate(migrate bool, dbProvider DBProvider) error {
...
+						}
+						if err := tx.Migrator().CreateConstraint(&Namespace{}, "Experiments"); err != nil {
+							return err
+						}
+						if err := tx.Migrator().AlterColumn(&Experiment{}, "Name"); err != nil {
+							return err
+						}
+						if err := tx.Migrator().CreateIndex(&Experiment{}, "idx_namespace_name"); err != nil {
+							return err
+						}
```
jgiannuzzi, line 334: We're still missing the migration for Apps

pkg/api/aim/apps.go
```diff
@@ -62,6 +86,7 @@ func GetApp(c *fiber.Ctx) error {
 	}
 	if err := database.DB.
 		Where("NOT is_archived").
+		Where("namespace_id = ?", ns.ID).
```
jgiannuzzi, line 89: move this into the `app` struct initialised above

pkg/api/aim/apps.go
```diff
@@ -136,6 +174,7 @@ func DeleteApp(c *fiber.Ctx) error {
 	if err := database.DB.
 		Select("ID").
 		Where("NOT is_archived").
+		Where("namespace_id = ?", ns.ID).
```
jgiannuzzi, line 177: move this into `app` struct

pkg/api/aim/dashboards.go
```diff
@@ -5,16 +5,27 @@ import (
...
+
 	var dashboards []database.Dashboard
 	if err := database.DB.
+		Preload("App").
+		Joins("LEFT JOIN apps ON apps.id = dashboards.app_id").
 		Where("NOT dashboards.is_archived").
-		Joins("App", database.DB.Select("ID", "Type", "IsArchived")).
+		Where("apps.namespace_id = ?", ns.ID).
 		Order("dashboards.updated_at").
 		Find(&dashboards).
```
jgiannuzzi, line 30: The whole SQL query building should be like this instead:

```go
		InnerJoins("App",
			database.DB.
				Select("ID", "Type").
				Where("namespace_id = ? AND NOT App.is_archived", ns.ID)).
		Where("NOT dashboards.is_archived").
		Order("dashboards.updated_at").
		Find(&dashboards).
```

This will ensure:
* we do a _single_ query to get both the dashboards and their associated apps
* we only get the information we need from the `apps` table
* we filter by namespace in the JOIN statement itself
* (new, we were not doing this before) we filter out archived apps

pkg/api/aim/dashboards.go
```diff
@@ -76,26 +100,34 @@ func GetDashboard(c *fiber.Ctx) error {
...
 		},
 	}
 	if err := database.DB.
 		Where("NOT dashboards.is_archived").
-		Joins("App", database.DB.Select("ID", "Type", "IsArchived")).
-		First(&dash).
+		Preload("App").
+		Joins("LEFT JOIN apps ON apps.id = dashboards.app_id").
+		Where("apps.namespace_id = ?", ns.ID).
+		First(&dashboard).
```
jgiannuzzi, line 113: same the remark above for `GetDashboards`

pkg/api/aim/dashboards.go
```diff
@@ -119,7 +151,9 @@ func UpdateDashboard(c *fiber.Ctx) error {
 		},
 	}
 	if err := database.DB.
-		Where("NOT is_archived").
+		Joins("LEFT JOIN apps ON dashboards.app_id = apps.id").
+		Where("NOT dashboards.is_archived").
+		Where("apps.namespace_id = ?", ns.ID).
```
jgiannuzzi, line 156: same remark as for `GetDashboards`

pkg/api/aim/dashboards.go
```diff
@@ -156,8 +196,10 @@ func DeleteDashboard(c *fiber.Ctx) error {
 		},
 	}
 	if err := database.DB.
-		Select("ID").
-		Where("NOT is_archived").
+		Select("dashboards.id").
+		Joins("LEFT JOIN apps ON dashboards.app_id = apps.id").
+		Where("NOT dashboards.is_archived").
+		Where("apps.namespace_id = ?", ns.ID).
```
jgiannuzzi, line 202: same remark as for `GetDashboards`

.github/workflows/ci.yml
```diff
@@ -151,7 +151,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        api: [aim, mlflow]
+        api: [ aim, mlflow ]
```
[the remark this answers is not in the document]
dsuhinin: ok.

pkg/api/aim/apps.go
dsuhinin: ahh. should be done.

pkg/database/migrate.go
dsuhinin: you are right. done.

pkg/api/aim/apps.go
dsuhinin: ok, I can do that. We have a mix when we use `where` statements via `model` and when we use `Where` function. I've decided to use `Where` function because we use it more often than inside the `model`.

pkg/api/aim/dashboards.go
dsuhinin: probably the final result has to be:
```
InnerJoins(
			"App",
			database.DB.Select("ID", "Type").Where(`"App".namespace_id = ? AND NOT "App".is_archived`, ns.ID),
		)
```
it is why I tried to do not use this syntax.

pkg/api/aim/dashboards.go
```diff
@@ -22,10 +22,15 @@ func GetDashboards(c *fiber.Ctx) error {
...
 	var dashboards []database.Dashboard
 	if err := database.DB.
-		Preload("App").
-		Joins("LEFT JOIN apps ON apps.id = dashboards.app_id").
+		InnerJoins(
+			"App",
+			database.DB.Select(
+				"ID", "Type",
+			).Where(
+				`"App".namespace_id = ? AND NOT "App".is_archived`, ns.ID,
```
jgiannuzzi, line 30: why the quoting?

dsuhinin: otherwise  it is just doesn't work. look at this:
Screenshot 2023-09-15 at 13 54 55

pkg/api/aim/dashboards.go
dsuhinin: done.

pkg/api/aim/dashboards.go
dsuhinin: done.

pkg/api/aim/dashboards.go
jgiannuzzi: I see, then I would recommend this:

```go
		InnerJoins("App",
			database.DB.
				Select("ID", "Type").
				Where(&database.App{NamespaceID: ns.ID}, "NamespaceID", "IsArchived")).
		Where("NOT dashboards.is_archived").
		Order("dashboards.updated_at").
		Find(&dashboards).
```

pkg/database/migrate.go
```diff
@@ -306,6 +309,43 @@ func checkAndMigrate(migrate bool, dbProvider DBProvider) error {
...
+				// We need to run this migration without foreign key constraints to avoid
+				// the cascading delete to kick in and delete all the runs.
+				if err := runWithoutForeignKeyIfNeeded(func() error {
+					if err := db.Transaction(func(tx *gorm.DB) error {
+						if err := tx.AutoMigrate(&Namespace{}); err != nil {
+							return err
+						}
+						if err := tx.Migrator().AddColumn(&App{}, "NamespaceID"); err != nil {
+							return err
+						}
```
jgiannuzzi, line 325: The constraint is missing

pkg/api/aim/encoding/encoder.go
```diff
@@ -42,7 +42,7 @@ func encodeTree(w io.Writer, v any, p []any) error {
 			return err
 		}
 		for i := 0; i < rv.Len(); i++ {
-			v := rv.Index(i).Interface()
+			v = rv.Index(i).Interface()
```
jgiannuzzi, line 45: if this was to fix the linting issues for shadowing variables, I'd rather we give it a new name. It feels dangerous to me to assign something to this top-level variable when it's clearly not our intention.

pkg/api/aim/encoding/decoder.go
```diff
@@ -49,7 +49,7 @@ func Decode(data io.Reader) (map[string]interface{}, error) {
 	for {
 		key, err := d.readField()
 		if err != nil {
-			if err := eris.Unwrap(err); err == io.EOF {
+			if err = eris.Unwrap(err); err == io.EOF {
```
jgiannuzzi, line 52: This feels dangerous to me! Better to use a different name if you want to avoid the linter yelling at you for overshadowing a variable. In this case, I _think_ that `err` will become the unwrapped `err` — whether it's an `io.EOF` or not — and then we may wrap it and return it and then loose some parts of it.

pkg/api/aim/experiments.go
```diff
@@ -149,7 +173,7 @@ func GetExperimentRuns(c *fiber.Ctx) error {
 		run := &database.Run{
 			ID: q.Offset,
 		}
-		if tx := database.DB.Select("row_num").First(&run); tx.Error != nil && tx.Error != gorm.ErrRecordNotFound {
+		if err = database.DB.Select("row_num").First(&run).Error; err != nil && tx.Error != gorm.ErrRecordNotFound {
 			return fmt.Errorf("unable to find search runs offset %q: %w", q.Offset, tx.Error)
```
jgiannuzzi, line 177:
```suggestion
		if err = database.DB.Select("row_num").First(&run).Error; err != nil && err != gorm.ErrRecordNotFound {
			return fmt.Errorf("unable to find search runs offset %q: %w", q.Offset, err)
```

pkg/api/aim/experiments.go
```diff
@@ -192,18 +222,21 @@ func GetExperimentActivity(c *fiber.Ctx) error {
...
 		return fiber.NewError(fiber.StatusUnprocessableEntity, fmt.Sprintf("unable to parse experiment id %q: %s", p.ID, err))
 	}
-	id32 := int32(id)
 
-	if tx := database.DB.Select("ID").First(&database.Experiment{
-		ID: &id32,
+	if tx := database.DB.Select(
+		"ID",
+	).Where(
+		"experiments.namespace_id = ?", ns.ID,
```
jgiannuzzi, line 237: Move this into the `Experiment` struct below

pkg/api/aim/dashboards.go
```diff
@@ -179,6 +179,7 @@ func UpdateDashboard(c *fiber.Ctx) error {
 	}
 
 	if err := database.DB.
+		Omit(clause.Associations).
```
jgiannuzzi, line 182: this is because you select too many fields above — we have no need for `Type` or `IsArchived` from the `App`

pkg/api/aim/dashboards.go
suprjinx: it could be DRY'd up if we moved the "fetch from context" step down to the repository

pkg/api/aim/experiments.go
dsuhinin: fixed.

pkg/api/aim/encoding/decoder.go
dsuhinin: yes yes, fixed.

pkg/api/aim/experiments.go
dsuhinin: done.

pkg/api/aim/dashboards.go
dsuhinin: reverted back for now.

pkg/api/aim/encoding/encoder.go
dsuhinin: hm this is a bit strange. you are right. reverted.

pkg/database/migrate.go
dsuhinin: done.

pkg/api/aim/apps.go
```diff
@@ -98,6 +129,7 @@ func UpdateApp(c *fiber.Ctx) error {
 	}
 	if err := database.DB.
 		Where("NOT is_archived").
+		Where("namespace_id = ?", ns.ID).
```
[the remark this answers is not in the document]
dsuhinin: done.

dave-gantenbein (approved):

PR #308, opened 2023-09-12
Fix database migrations to be fully compatible with MLFlow on Postgres

### LINKED ISSUE #305: Fix database migrations to be fully compatible with MLFlow on Postgres
#72 and #134 unfortunately introduced new migrations that are not compatible with an existing MLFlow database created with Postgres. The result is that those databases cannot be opened with FastTrackML anymore.

jgiannuzzi:
Fixes #305.

### DIFF
--- a/pkg/database/postgres_db_instance.go
+++ b/pkg/database/postgres_db_instance.go
@@ -11,7 +11,6 @@ import (
 	"gorm.io/driver/postgres"
 	"gorm.io/gorm"
 	"gorm.io/gorm/logger"
-	"gorm.io/plugin/dbresolver"
 )
 
 // PostgresDBInstance is the Postgres-specific DbInstance variant.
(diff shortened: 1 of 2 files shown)


pkg/database/postgres_db_instance.go
```diff
@@ -39,23 +38,15 @@ func NewPostgresDBInstance(
...
-	sourceConn = postgres.Open(dsnURL.String())
+	conn := postgres.Open(dsnURL.String())
 
-	logURL := dsnURL
-	q := logURL.Query()
-	if q.Has("_key") {
-		q.Set("_key", "xxxxx")
-	}
-	logURL.RawQuery = q.Encode()
-	log.Infof("Using database %s", logURL.Redacted())
```
dsuhinin, line 52: hah, this is what I also delete in my current PR, because this code is not necessary. I feel I have a lot of conflicts :)

pkg/database/postgres_db_instance.go
```diff
@@ -75,14 +66,5 @@ func NewPostgresDBInstance(
...
 
-	if replicaConn != nil {
-		pgdb.Use(
-			dbresolver.Register(dbresolver.Config{
-				Replicas: []gorm.Dialector{
-					replicaConn,
-				},
-			}),
-		)
-	}
```
dsuhinin, line 86: and yes, this also not needed.

dsuhinin (approved):

PR #309, opened 2023-09-13
Improve logic for validating whether metrics are selected in the metrics explorer

### LINKED ISSUE #302: Check for metric name in Aim UI metrics explorer does not consider the new regex syntax
We implemented a check for metric name in the Aim UI metrics explorer in #116. Unfortunately this now prevents users from searching for their metrics with the new regex syntax introduced in #171.

Example of a search query that should have been allowed, but isn't:
image

jgiannuzzi:
Fixes #302


dsuhinin (approved):
Nice!

PR #311, opened 2023-09-13
Improve Python wheel metadata

jgiannuzzi:
This PR improves the Python wheel metadata so that the information available on PyPI is more complete and relevant.

Here is an example on test.pypi.org before this change — https://test.pypi.org/project/FastTrackML/0.3.0/
image

And here it after this change — https://test.pypi.org/project/FastTrackML/0.3.1/
image


suprjinx (approved):

dsuhinin (approved):
Look good from what I can see but I think we need more eyes.

PR #312, opened 2023-09-13
Add initial cache support for namespaces

dsuhinin:
What's inside the box:
- caching layer for `namespace` logic.
- attach `namespace` middleware to only `mlflow` and `aim` routes instead of global router.


pkg/api/mlflow/routes.go
```diff
@@ -57,25 +59,31 @@ const (
...
+		controller:          controller,
+		namespaceRepository: namespaceRepository,
 	}
 }
 
 // Init makes initialization of all `mlflow` routes.
 func (r Router) Init(server fiber.Router) {
 	for _, prefix := range r.prefixList {
 		mainGroup := server.Group(prefix)
+		mainGroup.Use(namespaceMiddleware.New(r.namespaceRepository))
```
dsuhinin, line 86: moved middleware here, to connect it only to `mlflow` and `aim` routes instead of to all the possible like it was before.

.github/workflows/ci.yml
```diff
@@ -118,7 +118,7 @@ jobs:
           - database-backend: sqlite
             database-uri: sqlite:///tmp/fasttrackml.db
           - database-backend: postgres
-            database-uri: postgres://postgres:postgres@postgres/postgres
+            database-uri: postgres://postgres:postgres@postgres/postgres?sslmode=disable
```
dsuhinin, line 121: `pq.NewListener` asks for `ssl`, that's why it was disabled.

go.mod
```diff
@@ -67,6 +67,7 @@ require (
 	github.com/klauspost/asmfmt v1.3.2 // indirect
 	github.com/klauspost/compress v1.16.7 // indirect
 	github.com/klauspost/cpuid/v2 v2.2.4 // indirect
+	github.com/lib/pq v1.10.9 // indirect
```
jgiannuzzi, line 70: we shouldn't introduce another library to connect to Postgres — please use `github.com/jackc/pgx/v5` instead

dsuhinin: done.

PR #313, opened 2023-09-13
Fix install scripts

jgiannuzzi:
* fix the install script to work on Linux (wrong architecture name for aarch64/arm64 + wrong use of GNU coreutils `mktemp`)
* fix the install script to work on macOS (`tac` is not installed by default)
* general cleanup


fabiovincenzi (commented):
Looks good!

fabiovincenzi (dismissed):

dave-gantenbein (approved):

PR #367, opened 2023-09-20
Fix `ListArtifacts` to return correct paths

### LINKED ISSUE #355: Fix the MLFlow `ListArtifacts` endpoint to return correct paths
We are currently returning the server's default artifact root as the `root_uri` instead of the run's artifact location, and the full path to an artifact, instead of the one relative to the run's artifact location.

Here is an example of what we currently return:
```json
{
    "files": [
        {
            "path": "/workspaces/fasttrackml/artifacts/0/6521a6589fb144469e567aad58559cf4/artifacts/foo.txt",
            "is_dir": false,
            "file_size": 123
        }
    ],
    "root_uri": "/workspaces/fasttrackml/artifacts"
}
```

And this is what it should look like instead:
```json
{
    "files": [
        {
            "path": "foo.txt",
            "is_dir": false,
            "file_size": 123
        }
    ],
    "root_uri": "/workspaces/fasttrackml/artifacts/0/6521a6589fb144469e567aad58559cf4/artifacts"
}
```

jgiannuzzi:
Fixes #355.

- the `root_uri` key now returns the run's artifact location
- the `path` keys of the returned files are now relative to the `root_uri`
- an empty directory returns an empty list of files instead of an error
- avoids a race condition where a file is removed after we read the directory (local storage)
- properly list directories in S3 storage


pkg/database/migrate.go
```diff
@@ -2,9 +2,10 @@ package database
...
 import (
 	"fmt"
+	"slices"
 
 	log "github.com/sirupsen/logrus"
-	"golang.org/x/exp/slices"
+
 	"gorm.io/driver/sqlite"
 	"gorm.io/gorm"
 	"gorm.io/gorm/logger"
```
dsuhinin, line 11: please fix groping.

dsuhinin (approved):
It looks great! one tiny moment.

pkg/database/migrate.go
jgiannuzzi: good point — I will fix this everywhere whilst I'm at it

dsuhinin (dismissed):
In general I started have a feeling that this test should be in `table test` style. Not a big deal since we are in rush so we can a bit changes it later.

dsuhinin (commented):
Ahh, and it will be nice to add tests for `dir` object.

suprjinx (dismissed):
much clearer now what is a path and what is the rootUri :-)

pkg/database/migrate.go
jgiannuzzi: done in https://github.com/G-Research/fasttrackml/pull/367/commits/499029f421d742b434361ff8fb85ab19e3374485

jgiannuzzi:
> Ahh, and it will be nice to add tests for `dir` object.

already done in both unit and integration tests :)

dsuhinin (dismissed):

suprjinx (approved):

PR #368, opened 2023-09-20
Get artifacts from local or s3 storage

### LINKED ISSUE #357: Implement endpoint to get an artifact

suprjinx:
This PR adds support for streaming an individual artifact object from local or S3 storage.

Fixes #357


pkg/api/mlflow/controller/artifact.go
```diff
@@ -26,3 +28,21 @@ func (c Controller) ListArtifacts(ctx *fiber.Ctx) error {
...
+	if err := ctx.QueryParser(&req); err != nil {
+		return api.NewBadRequestError(err.Error())
+	}
+	log.Debugf("GetArtifact request: %#v", req)
+
+	artifact, err := c.artifactService.GetArtifact(ctx.Context(), &req)
+	if err != nil {
+		return err
+	}
+	defer artifact.Close()
```
dsuhinin, line 44: I don't like this idea to close `reader` object right here. It will be much more reasonable to close it in a place where actual works does. Controller has to just send bytes without knowing about underlying object.

pkg/api/mlflow/service/artifact/service.go
```diff
@@ -58,3 +59,33 @@ func (s Service) ListArtifacts(
...
+	}
+	if run == nil {
+		return nil, api.NewResourceDoesNotExistError("unable to find run '%s'", req.GetRunID())
+	}
+	artifactStorage, err := s.artifactStorageFactory.GetStorage(run.ArtifactURI)
+	if err != nil {
+		return nil, api.NewInternalError("run with id '%s' has unsupported artifact storage", run.ID)
+	}
+
+	artifactReader, err := artifactStorage.GetArtifact(
```
dsuhinin, line 83: will be nice to return here just a prepared bytes. also please change `artifactReader` to be `artifactObject`.

pkg/api/mlflow/service/artifact/service.go
```diff
@@ -58,3 +59,33 @@ func (s Service) ListArtifacts(
...
+		run.ArtifactURI, req.Path,
+	)
+	if err != nil {
+		return nil, api.NewInternalError("error getting artifact URI from storage")
```
dsuhinin, line 87: error message a bit misleading. I think it could be something like -> `error getting artifact object for URI: %s from storage`

pkg/api/mlflow/service/artifact/storage/local_test.go
```diff
@@ -7,8 +7,83 @@ import (
...
+	f, err := os.Create(filepath.Join(defaultArtifactRoot, runArtifactRoot, fileName))
+	assert.Nil(t, err)
+	_, err = f.Write([]byte(fileContent))
+	assert.Nil(t, err)
+
+	// invoke
+	svcConfig := &config.ServiceConfig{
+		DefaultArtifactRoot: defaultArtifactRoot,
+	}
+	svc, err := NewLocal(svcConfig)
```
dsuhinin, line 34: sorry, it isn't a service, it is a `storage`. can you please align names? the same probably about `svcConfig` could be just `config` or `cfg`.

pkg/api/mlflow/service/artifact/storage/local_test.go
```diff
@@ -7,8 +7,83 @@ import (
...
+		os.Remove(f.Name())
+	}()
+
+	// verify
+	assert.NotNil(t, file)
+	p := make([]byte, 20)
+	ln, err := file.Read(p)
+	assert.Nil(t, err)
+
+	assert.Equal(t, fileContent, string(p[:ln]))
```
dsuhinin, line 50: no need to have an empty line here.

pkg/api/mlflow/service/artifact/storage/local_test.go
```diff
@@ -7,8 +7,83 @@ import (
...
+	file, err := svc.GetArtifact(filepath.Join(defaultArtifactRoot, runArtifactRoot), fileName)
+	assert.Nil(t, err)
+	defer func() {
+		file.Close()
+		os.Remove(f.Name())
+	}()
+
+	// verify
+	assert.NotNil(t, file)
+	p := make([]byte, 20)
```
dsuhinin, line 46: can you please change this variables to have more meaning?

pkg/api/mlflow/service/artifact/storage/local_test.go
dsuhinin, line 50: if you just check `fileContent` with what `p` has won't it be the same?

pkg/api/mlflow/service/artifact/storage/s3.go
```diff
@@ -2,6 +2,7 @@ package storage
 
 import (
 	"context"
+	io "io"
```
dsuhinin, line 5: no need for alias.

pkg/api/mlflow/service/artifact/storage/s3.go
```diff
@@ -121,3 +122,23 @@ func (s S3) List(artifactURI, path string) ([]ArtifactObject, error) {
...
+		Bucket: aws.String(bucketName),
+		Key:    aws.String(filepath.Join(prefix, itemPath)),
+	}
+
+	resp, err := s.client.GetObject(context.TODO(), input)
+	if err != nil {
+		return nil, err
+	}
+
+	return resp.Body, nil
```
dsuhinin, line 143: will be nice to read data here and close response immediately.

pkg/api/mlflow/service/artifact/storage/s3.go
```diff
@@ -121,3 +122,23 @@ func (s S3) List(artifactURI, path string) ([]ArtifactObject, error) {
...
+	}
+
+	input := &s3.GetObjectInput{
+		Bucket: aws.String(bucketName),
+		Key:    aws.String(filepath.Join(prefix, itemPath)),
+	}
+
+	resp, err := s.client.GetObject(context.TODO(), input)
+	if err != nil {
+		return nil, err
```
dsuhinin, line 140: can you please wrap errors and add more information.

pkg/api/mlflow/service/artifact/storage/storage.go
```diff
@@ -31,13 +32,13 @@ func (o ArtifactObject) IsDirectory() bool {
 	return o.IsDir
 }
 
-// ArtifactStorageProvider provides and interface to work with particular artifact storage.
+// ArtifactStorageProvider provides an interface to work with artifact storage.
 type ArtifactStorageProvider interface {
-	// List lists all artifact object under provided path.
 	List(artifactURI, path string) ([]ArtifactObject, error)
+	GetArtifact(artifactURI, path string) (io.ReadCloser, error)
```
dsuhinin, line 38: can you please get back comments and add comment for new method?

dsuhinin: also, can you please rename `GetArtifact` to just `Get`?

pkg/api/mlflow/service/artifact/validators.go
```diff
@@ -32,3 +32,26 @@ func ValidateListArtifactsRequest(req *request.ListArtifactsRequest) error {
 	}
 	return nil
 }
+
+// ValidateGetArtifactRequest validates `GET /get-artifact` request.
```
dsuhinin, line 36: please update `GET /get-artifact` according to new route.

tests/integration/golang/helpers/client.go
```diff
@@ -148,3 +148,39 @@ func (c HttpClient) doRequest(httpMethod string, uri string, response interface{
 
 	return nil
 }
+
+// DoGetRequestNoUnmarshalling executes GET request to the uri, returning
+// responses body or error
+func (c HttpClient) DoGetRequestNoUnmarshalling(uri string) ([]byte, error) {
```
dsuhinin, line 154: we have `DoStreamRequest` won't it work?

tests/integration/golang/mlflow/artifact/get_artifact_local_test.go
```diff
@@ -0,0 +1,145 @@
...
+				ArtifactLocation: fmt.Sprintf("%s%s", tt.prefix, experimentArtifactDir),
+			})
+			assert.Nil(s.T(), err)
+
+			// 2. create test run.
+			runID := strings.ReplaceAll(uuid.New().String(), "-", "")
+			runArtifactDir := filepath.Join(experimentArtifactDir, runID, "artifacts")
+			run, err := s.runFixtures.CreateRun(context.Background(), &models.Run{
+				ID:             runID,
+				Status:         models.StatusRunning,
```
dsuhinin, line 100: do we really need to do all of that in cycle?

tests/integration/golang/mlflow/artifact/get_artifact_s3_test.go
```diff
@@ -0,0 +1,207 @@
...
+				LastUpdateTime: sql.NullInt64{
+					Int64: time.Now().UTC().UnixMilli(),
+					Valid: true,
+				},
+				LifecycleStage:   models.LifecycleStageActive,
+				ArtifactLocation: fmt.Sprintf("s3://%s/1", tt.bucket),
+			})
+			assert.Nil(s.T(), err)
+
+			// 2. create test run.
```
dsuhinin, line 99: probably the same question - do we need to do all of that in cycle?

dsuhinin (changes requested):

pkg/api/mlflow/routes.go
```diff
@@ -18,6 +18,7 @@ const (
 // List of `/artifact/*` routes.
 const (
 	ArtifactsListRoute = "/list"
+	ArtifactsGetRoute  = "/get"
```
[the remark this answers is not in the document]
suprjinx: sure

pkg/api/mlflow/controller/artifact.go
```diff
@@ -26,3 +28,21 @@ func (c Controller) ListArtifacts(ctx *fiber.Ctx) error {
...
+
+	_, err = io.CopyBuffer(ctx, artifact, make([]byte, 4096))
```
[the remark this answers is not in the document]
suprjinx: sure

pkg/api/mlflow/controller/artifact.go
suprjinx: i agree this seems not-ideal, but I do need to have the input and output streams in the same place for copy, and this seemed the only way to keep the fiber.Ctx in the controller only. 

What do you think of passing the fiber.Ctx as `Writer` argument down to storage provider?

pkg/api/mlflow/service/artifact/service.go
suprjinx: but what if it's big? I think we should deal in Reader/Writer instead of slice -- see above comment for alternative.

pkg/api/mlflow/service/artifact/storage/storage.go
suprjinx: yes can do

pkg/api/mlflow/service/artifact/validators.go
suprjinx: will do

tests/integration/golang/helpers/client.go
suprjinx: I'll look

tests/integration/golang/mlflow/artifact/get_artifact_local_test.go
suprjinx: not clear what `cycle` is here? Do you mean unnecessary attributes for the test?

pkg/api/mlflow/service/artifact/storage/local_test.go
suprjinx: p will potentially have extra bytes

pkg/api/mlflow/service/artifact/storage/s3.go
suprjinx: i disagree, could be big

pkg/api/mlflow/service/artifact/service.go
suprjinx: done

pkg/api/mlflow/service/artifact/storage/local_test.go
suprjinx: can change but i think the term "service" can apply generally to things which provide a functionality. The config is actually an instance of ServiceConfig so why rename?

pkg/api/mlflow/service/artifact/storage/local_test.go
suprjinx: ok

pkg/api/mlflow/service/artifact/storage/local_test.go
suprjinx: well this is the most important assertion, not quite the same as "err not nil" -- why not set it apart?

pkg/api/mlflow/service/artifact/storage/s3.go
suprjinx: k

tests/integration/golang/mlflow/artifact/get_artifact_s3_test.go
suprjinx: removed unneeded attrs

tests/integration/golang/helpers/client.go
suprjinx: good call, will switch to that

pkg/api/mlflow/controller/artifact.go
dsuhinin: passing it down also not a good idea. probably the same situation when we share implementation. but why we can't copy data into the `buffer`, pass buffer to `controller`. read and write `buffer` data into `fiber` context right in `controller`?

pkg/api/mlflow/service/artifact/service.go
dsuhinin: I think we can try and see. @jgiannuzzi will be nice to have your opinion.

tests/integration/golang/mlflow/artifact/get_artifact_local_test.go
```diff
@@ -0,0 +1,134 @@
...
+			name:   "TestWithFilePrefix",
+			prefix: "file://",
+		},
+		{
+			name:   "TestWithoutPrefix",
+			prefix: "",
+		},
+	}
+
+	for _, tt := range testData {
```
dsuhinin, line 69: I mean that you have 2 test cases: `TestWithFilePrefix` and `TestWithoutPrefix` and for both of them you create `experiment`, `run` etc, so why we can create everything before and only once?

tests/integration/golang/mlflow/artifact/get_artifact_s3_test.go
```diff
@@ -0,0 +1,194 @@
...
+			name:   "TestWithBucket1",
+			bucket: "bucket1",
+		},
+		{
+			name:   "TestWithBucket2",
+			bucket: "bucket2",
+		},
+	}
+
+	for _, tt := range testData {
```
dsuhinin, line 74: and the same question probably here? we can setup `experiment`, `run` and all other data only once? not for each test case?

pkg/api/mlflow/service/artifact/service.go
jgiannuzzi: I agree with @suprjinx that it's best to return an `io.ReadCloser` here so we can stream the artifact's bytes back to the client without making intermediate copies

pkg/api/mlflow/controller/artifact.go
jgiannuzzi: as I mentioned in another comment, this approach seems fine to me — one additional improvement would be to use `SetBodyStreamWriter` so that we send all the right headers to the client and then start streaming the response back (though it's worth checking whether this may already be the case with writing straight to `fiber.Ctx`???)

pkg/api/mlflow/service/artifact/service.go
dsuhinin: ok.

tests/integration/golang/mlflow/artifact/get_artifact_local_test.go
suprjinx: This variable is encoded into the Experiment.ArtifactLocation so it has to be created in the test loop.

pkg/api/mlflow/controller/artifact.go
suprjinx: converted to SetBodyStreamWriter (which, unfortunately, doesn't see to let us handle or return error?)

pkg/api/mlflow/service/artifact/storage/local_test.go
```diff
@@ -7,8 +7,82 @@ import (
...
+	f, err := os.Create(filepath.Join(defaultArtifactRoot, runArtifactRoot, fileName))
+	assert.Nil(t, err)
+	_, err = f.Write([]byte(fileContent))
+	assert.Nil(t, err)
+
+	// invoke
+	config := &config.ServiceConfig{
+		DefaultArtifactRoot: defaultArtifactRoot,
+	}
+	storage, err := NewLocal(config)
```
jgiannuzzi, line 34: you can pass `nil` to `NewLocal` (or an empty config if you prefer) — the default artifact root is not used anymore by the storage providers since my fixes

pkg/api/mlflow/controller/artifact.go
```diff
@@ -26,3 +29,30 @@ func (c Controller) ListArtifacts(ctx *fiber.Ctx) error {
...
+		return api.NewBadRequestError(err.Error())
+	}
+	log.Debugf("GetArtifact request: %#v", req)
+
+	artifact, err := c.artifactService.GetArtifact(ctx.Context(), &req)
+	if err != nil {
+		return err
+	}
+
+	ctx.Set("Content-Type", "application/octet-stream")
```
jgiannuzzi, line 46: please re-read the spec I wrote in #357!
you're missing many headers, and the content-type is wrong.

jgiannuzzi (changes requested):

pkg/api/mlflow/controller/artifact.go
suprjinx: will do

pkg/api/mlflow/service/artifact/storage/local_test.go
suprjinx: will do

pkg/api/mlflow/api/request/artifact.go
```diff
@@ -7,10 +7,25 @@ type ListArtifactsRequest struct {
...
 	return r.RunUUID
 }
+
+// GetRunID returns RunID if available, otherwise RunUUID.
+func (r GetArtifactRequest) GetRunID() string {
+	if r.RunID != "" {
+		return r.RunID
+	}
+	return r.RunUUID
+}
```
dsuhinin, line 31: just a small tiny change - can you please move it to be near original struct `GetArtifactRequest`?

pkg/api/mlflow/controller/artifact.go
```diff
@@ -26,3 +61,53 @@ func (c Controller) ListArtifacts(ctx *fiber.Ctx) error {
...
+	})
+
+	return nil
+}
+
+// getFilename returns the final bit of the path (the filename).
+func getFilename(path string) string {
+	pathParts := strings.Split(path, "/")
+	return pathParts[len(pathParts)-1]
+}
```
dsuhinin, line 99: can you please use golang `path` package for that.

pkg/api/mlflow/controller/artifact.go
```diff
@@ -26,3 +61,53 @@ func (c Controller) ListArtifacts(ctx *fiber.Ctx) error {
...
+	fileExt := fileParts[len(fileParts)-1]
+	if slices.Contains(textTypes, fileExt) {
+		return "text/plain"
+	}
+	mimeType := mime.TypeByExtension("." + fileExt)
+	if mimeType != "" {
+		return mimeType
+	}
+	return "application/octet-stream"
+}
```
dsuhinin, line 113: can you please use golang `path` package?  also please move this functions to `helpers.go` to keep `controller` clean.

pkg/api/mlflow/controller/artifact.go
```diff
@@ -9,6 +16,34 @@ import (
...
+	"py",
+	"py3",
+	"csv",
+	"tsv",
+	"md",
+	"rst",
+	"MLmodel",
+	"mlproject",
+}
+
```
dsuhinin, line 46: please move this also to the `helpers.go`

pkg/api/mlflow/controller/artifact.go
```diff
@@ -26,3 +61,53 @@ func (c Controller) ListArtifacts(ctx *fiber.Ctx) error {
...
+	ctx.Set("Content-Type", getContentType(filename))
+	ctx.Set("Content-Disposition", fmt.Sprintf("attachment; filename=%s", filename))
+	ctx.Set("X-Content-Type-Options", "nosniff")
+	ctx.Context().Response.SetBodyStreamWriter(func(w *bufio.Writer) {
+		defer artifact.Close()
+		bytesWritten, err := io.CopyBuffer(w, artifact, make([]byte, 4096))
+		if err != nil {
+			log.Errorf("error encountered in %s %s: error streaming artifact: %s", ctx.Method(), ctx.Path(), err)
+		}
+		log.Debugf("GetArtifact wrote bytes to output stream: %d", bytesWritten)
```
dsuhinin, line 88: you probably need `else` branch here, because otherwise `error` log message will be logged with `success` log message.

pkg/api/mlflow/controller/artifact_test.go
```diff
@@ -0,0 +1,49 @@
+package controller
+
+import (
+	"github.com/stretchr/testify/assert"
+	"testing"
+)
```
dsuhinin, line 6: please group properly imports.

pkg/api/mlflow/controller/artifact_test.go
```diff
@@ -0,0 +1,49 @@
...
+// Test_getContentType the Test_getContentType private helpers
+func Test_getContentType(t *testing.T) {
+	testCases := []struct {
+		name     string
+		filename string
+		expected string
+	}{
+		{
+			name:     "FromListOfTextTypes",
+			filename: "document.mlproject",
```
dsuhinin, line 25: please move this test into `helpers_test.go`

dsuhinin (changes requested):

pkg/api/mlflow/controller/artifact.go
suprjinx: i'm not sure we should move this unless it is needed by a second controller -- it forces the methods to become public, moves them away from the place they are used, etc

pkg/api/mlflow/controller/artifact.go
suprjinx: will do

pkg/api/mlflow/controller/artifact.go
dsuhinin: but please, this is not about to use it in several places. right now it uses only in one place so it means that we even don't need to move this logic into the function, right? this is a bout organisation in our project and we try to keep such unrelated logic in `helpers.go` throughout the project.

pkg/api/mlflow/controller/artifact.go
suprjinx: yep

pkg/api/mlflow/controller/artifact.go
suprjinx: went ahead and changed to move this along

pkg/api/mlflow/common/helpers.go
```diff
@@ -1,6 +1,62 @@
...
 // GetPointer returns pointer for provided string.
 func GetPointer[T any](str T) *T {
 	return &str
 }
+
+// GetFilename returns the final bit of the path (the filename).
+func GetFilename(fullpath string) string {
+	_, filename := path.Split(fullpath)
+	return filename
+}
```
jgiannuzzi, line 49: why did you reinvent the wheel instead of using `filepath.Base`?

pkg/api/mlflow/common/helpers.go
```diff
@@ -1,6 +1,62 @@
...
+import (
+	"mime"
+	"path"
+	"strings"
+
+	"golang.org/x/exp/slices"
+)
+
+// textTypes used by GetContentType.
+var textTypes = []string{
```
jgiannuzzi, line 12: please add `.` at the beginning of those extensions so we can refer to them without further manipulation to the result of `path.Ext` — this will be consistent with how the Go stdlib does things

pkg/api/mlflow/common/helpers.go
```diff
@@ -1,6 +1,62 @@
 package common
 
+import (
+	"mime"
+	"path"
+	"strings"
+
+	"golang.org/x/exp/slices"
```
jgiannuzzi, line 8: please don't use this — use `slices` instead!

pkg/api/mlflow/dao/repositories/mock_base_repository_provider.go
jgiannuzzi, line 1: why did you check these in????????

please look at your changes before committing and pushing!

pkg/api/mlflow/controller/artifact.go
```diff
@@ -43,16 +45,23 @@ func (c Controller) GetArtifact(ctx *fiber.Ctx) error {
...
 		if err != nil {
-			log.Errorf("error encountered in %s %s: error streaming artifact: %s", ctx.Method(), ctx.Path(), err)
+			log.Errorf(
+				"error encountered in %s %s: error streaming artifact: %s",
+				ctx.Method(),
+				ctx.Path(),
+				err,
+			)
 		}
 		log.Debugf("GetArtifact wrote bytes to output stream: %d", bytesWritten)
```
jgiannuzzi, line 63: please move this line _below_ the flush

jgiannuzzi: and please make it consistent with all of our other calls to `SetBodyStreamWriter`

go.mod
```diff
@@ -27,6 +27,7 @@ require (
 	github.com/spf13/cobra v1.7.0
 	github.com/spf13/viper v1.16.0
 	github.com/stretchr/testify v1.8.4
+	golang.org/x/exp v0.0.0-20220827204233-334a2380cb91
```
jgiannuzzi, line 30: I literally _just_ got rid of this lib in #367...

pkg/api/mlflow/service/artifact/storage/local_test.go
```diff
@@ -7,8 +7,79 @@ import (
...
+	readBuffer := make([]byte, 20)
+	ln, err := file.Read(readBuffer)
+	assert.Nil(t, err)
+	assert.Equal(t, fileContent, string(readBuffer[:ln]))
+}
+
+func TestGetArtifact_Error(t *testing.T) {
+	// setup
+	defaultArtifactRoot := t.TempDir()
+	runArtifactRoot := "/run-artifact-root/"
```
jgiannuzzi, line 52: please get rid of `defaultArtifactRoot` everywhere

pkg/api/mlflow/service/artifact/storage/local_test.go
```diff
@@ -7,8 +7,79 @@ import (
...
+	fileName := "file.txt"
+	fileContent := "artifact content"
+
+	err := os.MkdirAll(filepath.Join(defaultArtifactRoot, runArtifactRoot), os.ModePerm)
+	assert.Nil(t, err)
+
+	f, err := os.Create(filepath.Join(defaultArtifactRoot, runArtifactRoot, fileName))
+	assert.Nil(t, err)
+	_, err = f.Write([]byte(fileContent))
+	assert.Nil(t, err)
```
jgiannuzzi, line 62: why do you even bother creating a file and writing to it here???

pkg/api/mlflow/service/artifact/storage/local_test.go
```diff
@@ -7,8 +7,79 @@ import (
...
+
+	// invoke
+	config := &config.ServiceConfig{
+		DefaultArtifactRoot: defaultArtifactRoot,
+	}
+	storage, err := NewLocal(config)
+	assert.Nil(t, err)
+
+	file, err := storage.Get(filepath.Join(defaultArtifactRoot, runArtifactRoot), "some-other-item")
```
jgiannuzzi, line 71: please add a testcase for getting a directory instead of a file

pkg/api/mlflow/api/request/artifact.go
suprjinx: yep

tests/integration/golang/mlflow/artifact/get_artifact_s3_test.go
```diff
@@ -0,0 +1,194 @@
...
+	experimentFixtures, err := fixtures.NewExperimentFixtures(helpers.GetDatabaseUri())
+	assert.Nil(s.T(), err)
+	s.experimentFixtures = experimentFixtures
+
+	runFixtures, err := fixtures.NewRunFixtures(helpers.GetDatabaseUri())
+	assert.Nil(s.T(), err)
+	s.runFixtures = runFixtures
+}
+
+func (s *GetArtifactS3TestSuite) Test_Ok() {
```
jgiannuzzi, line 55: please also test getting a file in a subdirectory

tests/integration/golang/mlflow/artifact/get_artifact_s3_test.go
```diff
@@ -0,0 +1,194 @@
...
+func (s *GetArtifactS3TestSuite) Test_Error() {
+	defer func() {
+		assert.Nil(s.T(), s.experimentFixtures.UnloadFixtures())
+	}()
+
+	testData := []struct {
+		name    string
+		error   *api.ErrorResponse
+		request *request.GetArtifactRequest
+	}{
```
jgiannuzzi, line 133: please add a testcase where you try to get a valid path that is actually a directory

tests/integration/golang/mlflow/artifact/get_artifact_local_test.go
```diff
@@ -0,0 +1,134 @@
...
+	experimentFixtures, err := fixtures.NewExperimentFixtures(helpers.GetDatabaseUri())
+	assert.Nil(s.T(), err)
+	s.experimentFixtures = experimentFixtures
+
+	runFixtures, err := fixtures.NewRunFixtures(helpers.GetDatabaseUri())
+	assert.Nil(s.T(), err)
+	s.runFixtures = runFixtures
+}
+
+func (s *GetArtifactLocalTestSuite) Test_Ok() {
```
jgiannuzzi, line 50: where is `Test_Error`?

jgiannuzzi (changes requested):

pkg/api/mlflow/service/artifact/storage/local_test.go
suprjinx: is this a supported case? I think the UI only allows preview of leaf artifacts

pkg/api/mlflow/common/helpers.go
suprjinx: will do

pkg/api/mlflow/common/helpers.go
suprjinx: The path at hand is a string so it needs to be parsed either way, right? This seems pretty straightforward.

pkg/api/mlflow/dao/repositories/mock_base_repository_provider.go
suprjinx: restored

pkg/api/mlflow/controller/artifact.go
suprjinx: placed in `func() err` like the others, let me know if that's what you meant

pkg/api/mlflow/service/artifact/storage/local_test.go
jgiannuzzi: but you don't read it at all!
you are trying to read a file named "some-other-item" as this is a an _error_ test

pkg/api/mlflow/common/helpers.go
jgiannuzzi: But why create a helper function `GetFilename` and its corresponding unit test when you can use an existing function from the standard library? This is the opposite of straightforward!

pkg/api/mlflow/service/artifact/storage/local_test.go
suprjinx: it is now an error case

tests/integration/golang/mlflow/artifact/get_artifact_s3_test.go
suprjinx: i don't think s3 really has a directory concept -- it's just a key that won't match anything -- but I will add case to verify

tests/integration/golang/mlflow/artifact/get_artifact_local_test.go
suprjinx: it's the same as `Test_Error` in `get_artifact_s3_test.go`

tests/integration/golang/mlflow/artifact/get_artifact_s3_test.go
suprjinx: check

go.mod
suprjinx: removed

pkg/api/mlflow/common/helpers.go
suprjinx: changed to use filepath.Base at calling site

pkg/api/mlflow/service/artifact/storage/local_test.go
suprjinx: ah okay will remove

pkg/api/mlflow/service/artifact/storage/local_test.go
suprjinx: ok can do

pkg/api/mlflow/service/artifact/storage/local_test.go
```diff
@@ -9,6 +9,67 @@ import (
...
+	storage, err := NewLocal(nil)
+	assert.Nil(t, err)
+
+	file, err := storage.Get(filepath.Join(runArtifactRoot), "some-other-item")
+	assert.NotNil(t, err)
+	defer func() {
+		if file != nil {
+			file.Close()
+		}
+	}()
```
jgiannuzzi, line 61: please remove this and assert the file is nil instead

jgiannuzzi (changes requested):

tests/integration/golang/mlflow/artifact/get_artifact_s3_test.go
```diff
@@ -102,7 +102,19 @@ func (s *GetArtifactS3TestSuite) Test_Ok() {
...
 			assert.Nil(s.T(), err)
 
-			// 4. make actual API call.
+			// upload artifact subdir object to S3
+			putObjReq = &s3.PutObjectInput{
+				Key: aws.String(fmt.Sprintf(
+					"1/%s/artifacts/artifact.subdir/artifact.file",
+					runID),
+				),
+				Body:   strings.NewReader(`content`),
```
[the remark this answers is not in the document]
suprjinx: check

tests/integration/golang/mlflow/artifact/get_artifact_s3_test.go
```diff
@@ -176,6 +231,16 @@ func (s *GetArtifactS3TestSuite) Test_Error() {
...
+		{
+			name: "S3IncompletePath",
+			error: api.NewInternalError(
+				fmt.Sprintf("error getting artifact object for URI: s3:/bucket1/1/%s/artifacts/artifact.subdir", runID),
+			),
+			request: &request.GetArtifactRequest{
+				RunID: runID,
+				Path:  "artifact.subdir",
+			},
+		},
```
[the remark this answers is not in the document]
suprjinx: ok

tests/integration/golang/mlflow/artifact/get_artifact_local_test.go
```diff
@@ -132,3 +133,107 @@ func (s *GetArtifactLocalTestSuite) Test_Ok() {
...
+					"error getting artifact object for URI: %s/%s/artifacts/non-existent-file",
+					experimentArtifactDir,
+					runID,
+				),
+			),
+			request: &request.GetArtifactRequest{
+				RunID: runID,
+				Path:  "non-existent-file",
+			},
+		},
```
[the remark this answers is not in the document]
suprjinx: check

pkg/api/mlflow/service/artifact/storage/local_test.go
```diff
@@ -9,6 +9,67 @@ import (
...
+	assert.Nil(t, err)
+
+	// invoke
+	storage, err := NewLocal(nil)
+	assert.Nil(t, err)
+
+	file, err := storage.Get(runArtifactRoot, fileName)
+	assert.Nil(t, err)
+	defer func() {
+		file.Close()
```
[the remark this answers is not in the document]
suprjinx: i read the file below, so the close() will have to be moved below the read if it's not deferred

pkg/api/mlflow/service/artifact/storage/local_test.go
```diff
@@ -9,6 +9,67 @@ import (
...
+		os.Remove(f.Name())
```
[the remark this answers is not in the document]
suprjinx: check

pkg/api/mlflow/service/artifact/storage/local_test.go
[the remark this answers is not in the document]
jgiannuzzi: good point!

tests/integration/golang/mlflow/artifact/get_artifact_local_test.go
```diff
@@ -132,3 +133,107 @@ func (s *GetArtifactLocalTestSuite) Test_Ok() {
...
+			name:  "IncorrectPathProvidedCase5",
+			error: api.NewInvalidParameterValueError("provided 'path' parameter is invalid"),
+			request: &request.GetArtifactRequest{
+				RunID: "run_id",
+				Path:  "/foo/../bar",
+			},
+		},
+		{
+			name: "NonExistentPathProvided",
+			error: api.NewInternalError(
```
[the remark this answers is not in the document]
suprjinx: Did this for local storage, but I'm not able to identify the error so well from S3. Will keep experimenting.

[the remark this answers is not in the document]
suprjinx: added 404 detection for S3 case as well

pkg/api/mlflow/controller/artifact.go
```diff
@@ -45,23 +47,33 @@ func (c Controller) GetArtifact(ctx *fiber.Ctx) error {
...
-		if err != nil {
+
+		start := time.Now()
+		if err := func() error {
+			bytesWritten, err := io.CopyBuffer(w, artifact, make([]byte, 4096))
+			if err != nil {
+				return fmt.Errorf("unable to copy artifact Reader to output stream: %w", err)
+			}
+			if err := w.Flush(); err != nil {
+				return fmt.Errorf("GetArtifact error encountered flushing stream: %w", err)
```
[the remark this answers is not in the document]
suprjinx: check

pkg/api/mlflow/service/artifact/storage/local_test.go
```diff
@@ -48,36 +42,32 @@ func TestGetArtifact_Ok(t *testing.T) {
 
 func TestGetArtifact_Error(t *testing.T) {
 	// setup
-	defaultArtifactRoot := t.TempDir()
-	runArtifactRoot := "/run-artifact-root/"
-	fileName := "file.txt"
-	fileContent := "artifact content"
+	runArtifactRoot := t.TempDir()
+	subdir := "subdir/"
```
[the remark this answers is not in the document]
suprjinx: done

pkg/api/mlflow/service/artifact/storage/local_test.go
```diff
@@ -48,36 +42,32 @@ func TestGetArtifact_Ok(t *testing.T) {
...
 	// invoke
-	config := &config.ServiceConfig{
-		DefaultArtifactRoot: defaultArtifactRoot,
-	}
-	storage, err := NewLocal(config)
+	storage, err := NewLocal(nil)
 	assert.Nil(t, err)
 
-	file, err := storage.Get(filepath.Join(defaultArtifactRoot, runArtifactRoot), "some-other-item")
+	file, err := storage.Get(filepath.Join(runArtifactRoot), "some-other-item")
```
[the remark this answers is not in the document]
suprjinx: done

pkg/api/mlflow/service/artifact/storage/local_test.go
```diff
@@ -52,13 +49,9 @@ func TestGetArtifact_Error(t *testing.T) {
 	storage, err := NewLocal(nil)
 	assert.Nil(t, err)
 
-	file, err := storage.Get(filepath.Join(runArtifactRoot), "some-other-item")
+	file, err := storage.Get(filepath.Join(runArtifactRoot), "non-existent-file")
```
jgiannuzzi, line 52: could you please remove the useless call to `filepath.Join`, as indicated in my previous comment?

pkg/api/mlflow/service/artifact/service.go
```diff
@@ -85,10 +87,20 @@ func (s Service) GetArtifact(
...
 			"error getting artifact object for URI: %s",
 			filepath.Join(run.ArtifactURI, req.Path),
 		)
+		statError := regexp.MustCompile(storage.PathError)
+		isDirError := regexp.MustCompile(storage.IsDirError)
+		switch {
+		case statError.MatchString(err.Error()):
+			return nil, api.NewResourceDoesNotExistError(msg)
+		case isDirError.MatchString(err.Error()):
+			return nil, api.NewInvalidParameterValueError(msg)
```
jgiannuzzi, line 100: trying to get an existing directory should result in a `ResourceDoesNotExistError` too, like it already does in the S3 storage provider.

pkg/api/mlflow/service/artifact/service.go
```diff
@@ -85,10 +87,20 @@ func (s Service) GetArtifact(
...
+		default:
+			return nil, api.NewInternalError(msg)
+		}
```
jgiannuzzi, line 103: why do you use this weird regex matching instead of the standard `errors.Is`, or even `eris.Is`?

jgiannuzzi: also, if `os.Stat` or `os.Open` failed with a different error than `fs.ErrNotExist`, it _is_ an internal error, not a `ResourceDoesNotExistError`!

pkg/api/mlflow/service/artifact/storage/s3.go
```diff
@@ -137,7 +138,11 @@ func (s S3) Get(artifactURI, itemPath string) (io.ReadCloser, error) {
 
 	resp, err := s.client.GetObject(context.TODO(), input)
 	if err != nil {
-		return nil, eris.Wrap(err, "error fetching object from s3")
+		notFoundRegexp := regexp.MustCompile("StatusCode.+404")
+		if notFoundRegexp.MatchString(err.Error()) {
+			return nil, eris.Wrap(err, PathError)
+		}
+		return nil, eris.Wrap(err, "unable to get object")
```
jgiannuzzi, line 145: please handle errors correctly — have a look at the [AWS SDK v2 docs](https://aws.github.io/aws-sdk-go-v2/docs/handling-errors/#api-error-responses) and check for the `*types.NoSuchKey` error

jgiannuzzi (changes requested):

pkg/api/mlflow/service/artifact/service.go
suprjinx: i was thinking it does "exist" but we're requesting something we can't return in this case -- bad request. with S3, on the other hand, it does not actually exist.

pkg/api/mlflow/service/artifact/storage/local_test.go
suprjinx: check

pkg/api/mlflow/service/artifact/service.go
jgiannuzzi: no artifact exists on that path — so 404
the reason I asked you to handle it is that if you do `os.Open` on a directory, it will actually work and give you a io.ReadCloser back, but will then of course fail with `EISDIR` later

pkg/api/mlflow/service/artifact/storage/s3.go
jgiannuzzi: I literally tried this locally with minio before typing the comment above

pkg/api/mlflow/service/artifact/service.go
suprjinx: I can switch to eris.Is()

pkg/api/mlflow/service/artifact/storage/storage.go
```diff
@@ -31,13 +32,15 @@ func (o ArtifactObject) IsDirectory() bool {
 	return o.IsDir
 }
 
-// ArtifactStorageProvider provides and interface to work with particular artifact storage.
+// ArtifactStorageProvider provides an interface to work with artifact storage.
 type ArtifactStorageProvider interface {
+	// Get will return a io.ReadCloser for specific artifact.
+	Get(artifactURI, path string) (io.ReadCloser, error)
```
jgiannuzzi, line 38: for a future PR: all `ArtifactStorageProvider` methods should take a `context.Context`

tests/integration/golang/mlflow/artifact/get_artifact_s3_test.go
```diff
@@ -0,0 +1,269 @@
...
+			_, err = s.s3Client.PutObject(context.Background(), putObjReq)
+			assert.Nil(s.T(), err)
+
+			// upload artifact subdir object to S3
+			putObjReq = &s3.PutObjectInput{
+				Key: aws.String(fmt.Sprintf(
+					"1/%s/artifacts/artifact.subdir/artifact.file",
+					runID),
+				),
+				Body:   strings.NewReader(`subdir-object-content`),
```
jgiannuzzi, line 111: for future PR: let's get rid of those backtick strings here and in the ListArtifacts tests

jgiannuzzi (dismissed):
we eventually got there! 🎉

pkg/api/mlflow/common/helpers_test.go
```diff
@@ -0,0 +1,42 @@
...
+
+// TestGetContentType the TestGetContentType private helpers
+func TestGetContentType(t *testing.T) {
+	testCases := []struct {
+		name     string
+		filename string
+		expected string
+	}{
+		{
+			name:     "FromListOfTextTypes",
```
dsuhinin, line 17: it will be of course nice to text all the extensions if it possible.

pkg/api/mlflow/common/helpers_test.go
```diff
@@ -0,0 +1,42 @@
+package common
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+// TestGetContentType the TestGetContentType private helpers
```
dsuhinin, line 9: I think we don't need comments in tests since the name of test give the context.

pkg/api/mlflow/controller/artifact.go
```diff
@@ -26,3 +33,47 @@ func (c Controller) ListArtifacts(ctx *fiber.Ctx) error {
...
+			return nil
+		}(); err != nil {
+			log.Errorf(
+				"error encountered in %s %s: error streaming artifact: %s",
+				ctx.Method(),
+				ctx.Path(),
+				err,
+			)
+		}
+		log.Infof("body - %s %s %s", time.Since(start), ctx.Method(), ctx.Path())
```
dsuhinin, line 76: can you please add a bit more information here like:
```
time spent: %s, method: %s, route: %s
```

pkg/api/mlflow/controller/artifact.go
```diff
@@ -26,3 +33,47 @@ func (c Controller) ListArtifacts(ctx *fiber.Ctx) error {
...
+	ctx.Set("Content-Disposition", fmt.Sprintf("attachment; filename=%s", filename))
+	ctx.Set("X-Content-Type-Options", "nosniff")
+	ctx.Context().Response.SetBodyStreamWriter(func(w *bufio.Writer) {
+		defer artifact.Close()
+
+		start := time.Now()
+		if err := func() error {
+			bytesWritten, err := io.CopyBuffer(w, artifact, make([]byte, 4096))
+			if err != nil {
+				return fmt.Errorf("unable to copy artifact Reader to output stream: %w", err)
```
dsuhinin, line 61: can we try to use next format of error message -> `error doing something`. in that case it is much much easier when you search over the log, you always know that everything related to error star with `error` and it will be nice to still use `eris` since it add more information into `error` context.

pkg/api/mlflow/controller/artifact.go
```diff
@@ -26,3 +33,47 @@ func (c Controller) ListArtifacts(ctx *fiber.Ctx) error {
...
+			}
+			if err := w.Flush(); err != nil {
+				return fmt.Errorf("unable to flush output stream: %w", err)
```
dsuhinin, line 64: and here please as well.

pkg/api/mlflow/service/artifact/service.go
```diff
@@ -58,3 +63,40 @@ func (s Service) ListArtifacts(
...
+		run.ArtifactURI, req.Path,
+	)
+	if err != nil {
+		msg := fmt.Sprintf(
+			"error getting artifact object for URI: %s",
+			filepath.Join(run.ArtifactURI, req.Path),
+		)
+		if errors.Is(err, fs.ErrNotExist) {
+			return nil, api.NewResourceDoesNotExistError(msg)
+		} else {
```
dsuhinin, line 97: you probably don't need this `else` statement. just return `api.NewInternalError`. also it will be still nice to track original error here, since it has a lot of information about error, so please process it.

pkg/api/mlflow/service/artifact/storage/local_test.go
```diff
@@ -9,6 +9,60 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
+// TestGetArtifact for the local storage implementation.
```
dsuhinin, line 12: no need for comments here. for other tests we have no comments - which is fine, so lets have everything consistent.

pkg/api/mlflow/service/artifact/storage/s3.go
```diff
@@ -121,3 +125,27 @@ func (s S3) List(artifactURI, path string) ([]ArtifactObject, error) {
...
+	}
+
+	input := &s3.GetObjectInput{
+		Bucket: aws.String(bucketName),
+		Key:    aws.String(filepath.Join(prefix, itemPath)),
+	}
+
+	resp, err := s.client.GetObject(context.TODO(), input)
+	if err != nil {
+		var s3NoSuchKey *types.NoSuchKey
```
dsuhinin, line 143: you have a pointer here and the take address?

pkg/api/mlflow/service/artifact/storage/s3.go
```diff
@@ -121,3 +125,27 @@ func (s S3) List(artifactURI, path string) ([]ArtifactObject, error) {
...
+		if errors.As(err, &s3NoSuchKey) {
```
dsuhinin, line 144: maybe `is` will work better/shorter? you just need to make sure that err has type of `type.NoSuchKey` . you don't need then to use this error, so maybe `is`?

pkg/api/mlflow/service/artifact/storage/storage.go
```diff
@@ -31,13 +32,15 @@ func (o ArtifactObject) IsDirectory() bool {
 	return o.IsDir
 }
 
-// ArtifactStorageProvider provides and interface to work with particular artifact storage.
+// ArtifactStorageProvider provides an interface to work with artifact storage.
 type ArtifactStorageProvider interface {
+	// Get will return a io.ReadCloser for specific artifact.
```
dsuhinin, line 37: can you please change `comment` to be like `Get returns a io.ReadCloser for specific artifact.`

pkg/api/mlflow/service/artifact/validators_test.go
```diff
@@ -139,3 +139,156 @@ func TestValidateListArtifactsRequest_Error(t *testing.T) {
...
 }
+
+func TestValidateGetArtifactRequest_Ok(t *testing.T) {
+	testData := []struct {
+		name    string
+		request request.GetArtifactRequest
+	}{
+		{
+			name: "NotEmptyPathCase1",
+			request: request.GetArtifactRequest{
```
dsuhinin, line 150: can you please add another case with `RunUUID`.

pkg/api/mlflow/service/artifact/validators_test.go
```diff
@@ -139,3 +139,156 @@ func TestValidateListArtifactsRequest_Error(t *testing.T) {
 		})
 	}
 }
+
+func TestValidateGetArtifactRequest_Ok(t *testing.T) {
+	testData := []struct {
```
dsuhinin, line 144: sometimes we use `testCases` sometimes `testData`, can you please unity it? for me `tests` just works really well.

pkg/api/mlflow/common/helpers_test.go
```diff
@@ -0,0 +1,42 @@
...
+func TestGetContentType(t *testing.T) {
+	testCases := []struct {
```
dsuhinin, line 11: sometimes we use `testCases` sometimes `testData`, can you please unity it? for me `tests` just works really well.

pkg/api/mlflow/service/artifact/validators_test.go
```diff
@@ -139,3 +139,156 @@ func TestValidateListArtifactsRequest_Error(t *testing.T) {
...
+
+	for _, tt := range testData {
+		t.Run(tt.name, func(t *testing.T) {
+			assert.Nil(t, ValidateGetArtifactRequest(&tt.request))
+		})
+	}
+}
+
+func TestValidateGetArtifactRequest_Error(t *testing.T) {
+	testData := []struct {
```
dsuhinin, line 228: sometimes we use `testCases` sometimes `testData`, can you please unity it? for me `tests` just works really well.

tests/integration/golang/mlflow/artifact/get_artifact_local_test.go
```diff
@@ -0,0 +1,256 @@
...
+	assert.Nil(s.T(), err)
+	s.runFixtures = runFixtures
+}
+
+func (s *GetArtifactLocalTestSuite) Test_Ok() {
+	defer func() {
+		assert.Nil(s.T(), s.experimentFixtures.UnloadFixtures())
+	}()
+
+	testData := []struct {
```
dsuhinin, line 56: sometimes we use `testCases` sometimes `testData`, can you please unity it? for me `tests` just works really well.

tests/integration/golang/mlflow/artifact/get_artifact_local_test.go
```diff
@@ -0,0 +1,256 @@
...
+		ExperimentID:   *experiment.ID,
+		ArtifactURI:    runArtifactDir,
+		LifecycleStage: models.LifecycleStageActive,
+	})
+	assert.Nil(s.T(), err)
+
+	err = os.MkdirAll(filepath.Join(runArtifactDir, "subdir"), fs.ModePerm)
+	assert.Nil(s.T(), err)
+
+	testData := []struct {
```
dsuhinin, line 163: sometimes we use `testCases` sometimes `testData`, can you please unity it? for me `tests` just works really well.

tests/integration/golang/mlflow/artifact/get_artifact_s3_test.go
```diff
@@ -0,0 +1,269 @@
...
+	assert.Nil(s.T(), err)
+	s.runFixtures = runFixtures
+}
+
+func (s *GetArtifactS3TestSuite) Test_Ok() {
+	defer func() {
+		assert.Nil(s.T(), s.experimentFixtures.UnloadFixtures())
+	}()
+
+	testData := []struct {
```
dsuhinin, line 60: sometimes we use `testCases` sometimes `testData`, can you please unity it? for me `tests` just works really well.

tests/integration/golang/mlflow/artifact/get_artifact_s3_test.go
```diff
@@ -0,0 +1,269 @@
...
+	// upload artifact subdir object to S3
+	putObjReq := &s3.PutObjectInput{
+		Key:    aws.String(fmt.Sprintf("1/%s/artifacts/artifact.subdir/artifact.file", runID)),
+		Body:   strings.NewReader(`content`),
+		Bucket: aws.String("bucket1"),
+	}
+	_, err = s.s3Client.PutObject(context.Background(), putObjReq)
+	assert.Nil(s.T(), err)
+
+	testData := []struct {
```
dsuhinin, line 184: sometimes we use `testCases` sometimes `testData`, can you please unity it? for me `tests` just works really well.

tests/integration/golang/mlflow/artifact/get_artifact_s3_test.go
```diff
@@ -0,0 +1,269 @@
...
+		ExperimentID:   *experiment.ID,
+		ArtifactURI:    fmt.Sprintf("%s/%s/artifacts", experiment.ArtifactLocation, runID),
+		LifecycleStage: models.LifecycleStageActive,
+	})
+	assert.Nil(s.T(), err)
+
+	// upload artifact subdir object to S3
+	putObjReq := &s3.PutObjectInput{
+		Key:    aws.String(fmt.Sprintf("1/%s/artifacts/artifact.subdir/artifact.file", runID)),
+		Body:   strings.NewReader(`content`),
```
dsuhinin, line 178: lets use `""` instead for consistency?

pkg/api/mlflow/controller/artifact.go
dsuhinin, line 64: can you please create `error` via `eris`.

pkg/api/mlflow/service/artifact/service.go
suprjinx: @jgiannuzzi and I discussed this, it probably needs a refactoring of the API errors to wrap an error, which is beyond the scope of this PR.

pkg/api/mlflow/service/artifact/storage/s3.go
suprjinx: yes that is the example in documentation from aws sdk

pkg/api/mlflow/service/artifact/storage/s3.go
suprjinx: following example from the aws sdk docs, `is` does not work

pkg/api/mlflow/controller/artifact.go
suprjinx: sure

tests/integration/golang/mlflow/artifact/get_artifact_s3_test.go
suprjinx: sure

tests/integration/golang/mlflow/artifact/get_artifact_s3_test.go
suprjinx: sure

tests/integration/golang/mlflow/artifact/get_artifact_s3_test.go
suprjinx: sure

tests/integration/golang/mlflow/artifact/get_artifact_local_test.go
suprjinx: sure

pkg/api/mlflow/service/artifact/storage/local_test.go
suprjinx: ok

pkg/api/mlflow/service/artifact/storage/storage.go
suprjinx: ok

pkg/api/mlflow/service/artifact/validators_test.go
suprjinx: added RunUUIDAndNotEmptyPath

pkg/api/mlflow/service/artifact/validators_test.go
suprjinx: done

pkg/api/mlflow/common/helpers_test.go
suprjinx: done

pkg/api/mlflow/service/artifact/validators_test.go
suprjinx: done

tests/integration/golang/mlflow/artifact/get_artifact_local_test.go
suprjinx: done

pkg/api/mlflow/controller/artifact.go
suprjinx: done

pkg/api/mlflow/controller/artifact.go
suprjinx: can do

pkg/api/mlflow/controller/artifact.go
suprjinx: this was following example @jgiannuzzi requested

pkg/api/mlflow/common/helpers_test.go
suprjinx: sample that shows code is working as expected is correct, yes?

pkg/api/mlflow/common/helpers_test.go
suprjinx: removing

pkg/api/mlflow/controller/artifact.go
suprjinx: but of course, it is obvious an error message pertains to error

pkg/api/mlflow/service/artifact/service.go
dsuhinin: ok, what about `else`?

dsuhinin (approved):
Just one small moment related to `else` statement.

jgiannuzzi (approved):

pkg/api/mlflow/service/artifact/storage/s3.go
jgiannuzzi: I have tried as well but haven't found a way to get this to work with a regular `errors.Is`. It seems to be because the error is a pointer _and_ does not implement `Is()`.

pkg/api/mlflow/service/artifact/storage/storage.go
jgiannuzzi: done in #405

tests/integration/golang/mlflow/artifact/get_artifact_s3_test.go
jgiannuzzi: done in #405

PR #378, opened 2023-09-21
Support multiple artifact storage backends

dsuhinin:


dsuhinin:
> Just scanned it very quickly and it looks to me like we will create a new storage provider on every request to list artifacts. For the S3 provider, I believe this would mean recreating a client (and thus connections) every time.
> 
> I think we should instead register each storage provider with a URL scheme and instantiate them (lazily?) in the "factory" so they can be returned based on the scheme of the requested URI. So a proxy instead of a factory I guess.

emmm, I understand what you mean. in general even now there is no heavy lifting but let me think.

pkg/api/mlflow/config/config_test.go
```diff
@@ -17,19 +17,19 @@ func TestServiceConfig_Validate_Ok(t *testing.T) {
 		{
 			name: "ArtifactRootHasS3Prefix",
 			providedConfig: &ServiceConfig{
-				ArtifactRoot: "s3://bucket_name",
+				DefaultArtifactRoot: "s3://bucket_name",
```
jgiannuzzi, line 20: Please update all the test case names as well

pkg/cmd/import.go
```diff
@@ -74,7 +74,7 @@ func init() {
 
 	ImportCmd.Flags().StringP("input-database-uri", "i", "", "Input Database URI (eg., sqlite://fasttrackml.db)")
 	ImportCmd.Flags().StringP("output-database-uri", "o", "", "Output Database URI (eg., postgres://user:psw@postgres:5432)")
-	ImportCmd.Flags().StringP("artifact-root", "a", "./artifacts", "Artifact Root")
+	ImportCmd.Flags().StringP("default-artifact-root", "a", "./artifacts", "Artifact Root")
```
jgiannuzzi, line 77: please also update the lines above, most importantly line 44!

pkg/api/mlflow/config/config.go
```diff
@@ -34,8 +34,8 @@ func NewServiceConfig() *ServiceConfig {
 		ListenAddress:         viper.GetString("listen-address"),
 		AuthUsername:          viper.GetString("auth-username"),
 		AuthPassword:          viper.GetString("auth-password"),
-		ArtifactRoot:          viper.GetString("artifact-root"),
 		S3EndpointURI:         viper.GetString("s3-endpoint-uri"),
+		DefaultArtifactRoot:   viper.GetString("default-artifact-root"),
```
jgiannuzzi, line 38: if you move it down, maybe sort the list?

pkg/cmd/server.go
```diff
@@ -208,7 +207,7 @@ func init() {
 	RootCmd.AddCommand(ServerCmd)
 
 	ServerCmd.Flags().StringP("listen-address", "a", "localhost:5000", "Address (host:post) to listen to")
-	ServerCmd.Flags().String("artifact-root", "./artifacts", "Artifact root")
+	ServerCmd.Flags().String("default-artifact-root", "./artifacts", "Artifact root")
```
jgiannuzzi, line 210:
```suggestion
	ServerCmd.Flags().String("default-artifact-root", "./artifacts", "Default artifact root")
```

jgiannuzzi (changes requested):
Please also update `db_factory.go`, `db_instance.go`, and `import_test.go`

jgiannuzzi (commented):
Please also update `db_factory.go`, `db_instance.go`, and `import_test.go`

pkg/api/mlflow/config/config_test.go
dsuhinin: ahh, shit. yes.

pkg/api/mlflow/config/config.go
dsuhinin: I though about that too =) but since it is a bit related to `s3-endpoint`, decided to keep it here.

dsuhinin: but ok, let me move it.

pkg/api/mlflow/config/config.go
```diff
@@ -35,11 +35,11 @@ func NewServiceConfig() *ServiceConfig {
 		AuthUsername:          viper.GetString("auth-username"),
 		AuthPassword:          viper.GetString("auth-password"),
 		S3EndpointURI:         viper.GetString("s3-endpoint-uri"),
-		DefaultArtifactRoot:   viper.GetString("default-artifact-root"),
 		DatabaseURI:           viper.GetString("database-uri"),
 		DatabaseReset:         viper.GetBool("database-reset"),
 		DatabasePoolMax:       viper.GetInt("database-pool-max"),
 		DatabaseMigrate:       viper.GetBool("database-migrate"),
+		DefaultArtifactRoot:   viper.GetString("default-artifact-root"),
```
jgiannuzzi, line 42: This position makes even less sense 😂

dsuhinin: correct.

dsuhinin: lets keep it like before.

jgiannuzzi (approved):

PR #381, opened 2023-09-22
Improve setup for integration tests

jgiannuzzi:
- Dramatically improve integration tests speed and setup complexity 
  - move docker-compose.yml with the integration tests
  - use proper healthchecks to order service startup
  - cleanup shared /tmp volume at service startup
  - use Go module and build cache to dramatically speedup running the tests
  - share the Go cache between the service and the tests to speedup even more
  - set env var defaults in docker-compose.yml, where they belong, instead of the
    Makefile
- Integration tests run against all artifact backends at the same time
- Run integration tests against encrypted SQLite database too
- Fix import test linting issues
- Prevent integration tests from creating cruft in the source directory


tests/integration/golang/database/import_test.go
```diff
@@ -79,6 +79,7 @@ func (s *ImportTestSuite) SetupTest() {
 	s.runs = runs
 
 	appFixtures, err := fixtures.NewAppFixtures(helpers.GetInputDatabaseUri())
+	assert.Nil(s.T(), err)
```
dsuhinin, line 82: all of this fixes exists in `namespace` branch also.

dsuhinin (approved):

tests/integration/docker-compose.yml
```diff
@@ -0,0 +1,78 @@
...
+    depends_on:
+      minio:
+        condition: service_healthy
+      postgres:
+        condition: service_healthy
+      service:
+        condition: service_healthy
+    environment:
+      GOCACHE: /cache/go-build
+      GOMODCACHE: /cache/go-mod
```
dsuhinin, line 69: I didn't know that. hm. so we download deps only once and then just re use them, right as I can ses?

jgiannuzzi: correct — and we also don't rebuild everything every time, but instead let Go figure out what needs to be rebuilt, which is bringing the most important speedup, especially in the local dev loop!

dsuhinin: you mean changed tests. the same approach when you just run `go test ./...` to run unit tests. so if something changed, go will only re run this tests?

jgiannuzzi: correct

jgiannuzzi: which reminds me that we should add `-count=1` to the integration tests so their results don't get cached

jgiannuzzi: fixed in https://github.com/G-Research/fasttrackml/pull/382

PR #399, opened 2023-09-27
Address namespace PR comments.

dsuhinin:


pkg/api/aim/dashboards.go
```diff
@@ -106,10 +111,15 @@ func GetDashboard(c *fiber.Ctx) error {
 		},
 	}
 	if err := database.DB.
+		InnerJoins(
+			"App",
+			database.DB.Select(
```
dave-gantenbein, line 116: Can we DRY this up?

jgiannuzzi: good point

jgiannuzzi (commented):
could you please apply the same `IntterJoins` + `Select` logic to all the other places where we now join on experiments from the namespace?

dsuhinin:
> could you please apply the same `InnerJoins` + `Select` logic to all the other places where we now join on experiments from the namespace?

applied `inner` join in other places.

pkg/api/aim/dashboards.go
dsuhinin: Im not sure that this is good point since it is a `sql` builder so I always want to see expressions, values, etc to be in the same place. I can't see any reason to hide into the functions to save just a few lines and have less readability?

jgiannuzzi: I see your point, but if we call this `FilterQueryByNamespace(tx *gorm.DB, parent string, fields []string, nsID string) *gorm.DB`, we could make the whole seriously more readable. It would also avoid the unoptimised queries we have below where we select all the fields instead of just the ones we care about. And this helper can be used for runs too!

pkg/api/mlflow/dao/repositories/metric.go
```diff
@@ -154,7 +154,7 @@ func (r MetricRepository) GetMetricHistories(
 		query := r.db.WithContext(ctx).Model(
 			&database.Run{},
 		).Joins(
-			"LEFT JOIN experiments ON experiments.experiment_id = runs.experiment_id",
+			"INNER JOIN experiments ON experiments.experiment_id = runs.experiment_id",
 		).Where(
 			"experiments.namespace_id = ?", namespaceID,
```
jgiannuzzi, line 159: this should be replaced by the same kind of inner join we do for apps, so that we filter whilst joining instead of joining everything and then having the where clause applied — if we go for the helper described in https://github.com/G-Research/fasttrackml/pull/399/files#r1358038830, this can all be made simpler and more readable IMHO

pkg/api/aim/dashboards.go
dsuhinin: ok. let me try.

dsuhinin: ok, I think that this approach won't work if we have for example next joining:
```
join `metrics` with `runs`.
and then
join `runs` with `experiments`
```
also, since we have no standardised approach to build our queries:
in mlflow we mostly use `common` gorm functions:
```
.Where(
		"runs.run_uuid IN ?", runIDs,
	).Joins(
		"LEFT JOIN runs ON runs.run_uuid = metrics.run_uuid",
	).Joins(
		"INNER JOIN experiments ON experiments.experiment_id = runs.experiment_id",
	).Where(
		"key = ?", key,
	).Where(
		"experiments.namespace_id = ?", namespaceID,
	).Order(
		"metrics.run_uuid",
	).Order(
		"metrics.timestamp",
	).Order(
		"metrics.step",
	).Order(
		"metrics.value",
	)
```
but in `aim` we use a bit more `object` oriented apporach:
```
tx := database.DB.
		Joins(
			"Experiment",
			database.DB.Where(&models.Experiment{NamespaceID: ns.ID}),
		).
		Preload("Params").
		Preload("Tags")
```
it is a bit tricky to unify all of that. So I propose to firstly refactor `aim` part and and use one approach to build `queires` everywhere and then check what we can improve.

jgiannuzzi: ok sounds good to me — but please change the `mlflow` inner joins to filter on namespace _in the join_

dsuhinin: should be done everywhere.

PR #407, opened 2023-09-29
Features/namespaces

suprjinx:
PR for comparing `features/namespaces` to `main`


pkg/api/aim/apps.go
```diff
@@ -2,18 +2,29 @@ package aim
...
 	"github.com/G-Research/fasttrackml/pkg/database"
 )
 
 func GetApps(c *fiber.Ctx) error {
+	ns, err := namespace.GetNamespaceFromContext(c.Context())
+	if err != nil {
+		return api.NewInternalError("error getting namespace from context")
+	}
+	log.Debugf("getApps namespace: %s", ns.Code)
+
```
suprjinx, line 23: it seems like we could do this "extract from context" in a middleware or something so it is not repeated in every single controller function

pkg/api/mlflow/service/experiment/service.go
```diff
@@ -231,12 +248,16 @@ func (s Service) SetExperimentTag(ctx context.Context, req *request.SetExperimen
...
-	ctx context.Context, req *request.SearchExperimentsRequest,
+	ctx context.Context, ns *models.Namespace, req *request.SearchExperimentsRequest,
 ) ([]models.Experiment, int, int, error) {
 	if err := ValidateSearchExperimentsRequest(req); err != nil {
 		return nil, 0, 0, err
 	}
 
+	query := database.DB.Where(
+		"experiments.namespace_id = ?", ns.ID,
+	)
```
suprjinx, line 259: i like this approach, where we separate the namespace narrowing from the rest of the query

pkg/api/mlflow/service/experiment/service_test.go
```diff
@@ -90,7 +102,7 @@ func TestService_CreateExperiment_Error(t *testing.T) {
 			service: func() *Service {
 				experimentRepository := repositories.MockExperimentRepositoryProvider{}
 				experimentRepository.On(
-					"GetByName", context.TODO(), "name",
+					"GetByNamespaceIDAndName", context.TODO(), ns.ID, "name",
```
suprjinx, line 105: should we use `context.Background()` in cases like this where there is nothing more to do?

pkg/api/mlflow/service/metric/service.go
```diff
@@ -12,24 +12,37 @@ import (
 
 // Service provides service layer to work with `metric` business logic.
 type Service struct {
+	runRepository    repositories.RunRepositoryProvider
 	metricRepository repositories.MetricRepositoryProvider
```
suprjinx, line 16: alpha order?

pkg/api/mlflow/service/metric/service.go
```diff
@@ -12,24 +12,37 @@ import (
...
+) *Service {
 	return &Service{
+		runRepository:    runRepository,
 		metricRepository: metricRepository,
 	}
 }
 
 func (s Service) GetMetricHistory(
-	ctx context.Context, req *request.GetMetricHistoryRequest,
+	ctx context.Context, namespace *models.Namespace, req *request.GetMetricHistoryRequest,
```
suprjinx, line 31: I think the argument types feel a little mismatched -- one is a `model`, one a `request`. On the one hand, it seems like namespace could just be an attribute on the GetMetricHistoryRequest, but I also don't quite think the controller-level request should be used in the service (similarly, the controller-level api errors feel like the wrong level).

pkg/api/aim/apps.go
suprjinx: To clarify, why even put it in the context if we extract it right away in the controller function? Couldn't we just as easily call a function here with `fiber.Ctx.Params("ns")` which returns the `ns.ID` or `Namespace` struct?

pkg/api/mlflow/service/artifact/service_test.go
```diff
@@ -15,55 +15,63 @@ import (
 )
 
 func TestService_ListArtifacts_Ok(t *testing.T) {
-	artifactStorage := storage.MockProvider{}
+	artifactStorage := storage.MockArtifactStorageProvider{}
 	artifactStorage.On(
-		"List", "/artifact/uri", "",
+		"List", context.TODO(), "/artifact/uri", "",
```
suprjinx, line 20: I'd change to `context.Background()` for tests, since there is no intention to do anything else.

pkg/api/mlflow/service/artifact/service_test.go
```diff
@@ -15,55 +15,63 @@ import (
...
+		uint(1),
 		"id",
 	).Return(&models.Run{
 		ID:          "id",
 		ArtifactURI: "/artifact/uri",
 	}, nil)
 
 	// call service under testing.
-	service := NewService(&artifactStorage, &runRepository)
+	service := NewService(&runRepository, &artifactStorageFactory)
```
suprjinx, line 54: why do these change to unalphabetized?

PR #417, opened 2023-10-03
Add support for negative numbers in Aim query syntax

### LINKED ISSUE #416: Negative numbers are not supported in Aim query syntax
The query syntax in Aim UI does not accept negative numbers: `run.metrics['my_metric'].last < -1` gives `Error:Query syntax error at line (1, 35)`

jgiannuzzi:
Fixes #416.

Also reworked the query unit tests to be more readable and more focused on the actual query parsing.

### DIFF
--- a/pkg/api/aim/query/query.go
+++ b/pkg/api/aim/query/query.go
@@ -733,13 +733,23 @@ func (pq *parsedQuery) parseUnaryOp(node *ast.UnaryOp) (any, error) {
 	if err != nil {
 		return nil, err
 	}
-	c, ok := e.(clause.Expression)
-	if !ok {
-		return nil, fmt.Errorf("not a valid SQL expression: %#v", e)
-	}
 	switch node.Op {
+	case ast.USub:
+		switch e := e.(type) {
+		case int:
+			return -e, nil
+		case float64:
+			return -e, nil
+		default:
+			return nil, fmt.Errorf("unsupported unary operation %q on %T", node.Op, e)
+		}
 	case ast.Not:
-		return clause.Not(c), nil
+		switch e := e.(type) {
+		case clause.Expression:
+			return clause.Not(e), nil
+		default:
+			return nil, fmt.Errorf("unsupported unary operation %q on %T", node.Op, e)
+		}
 	default:
 		return nil, fmt.Errorf("unsupported unary operation %q", node.Op)
 	}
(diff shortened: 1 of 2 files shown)


pkg/api/aim/query/query_test.go
```diff
@@ -42,47 +42,95 @@ func (s *QueryTestSuite) TestPostgresDialector_Ok() {
 		expectedVars []interface{}
 	}{
 		{
-			name:         "TestRunNameWithoutFunction",
-			query:        `(run.name == 'run')`,
-			expectedSQL:  `SELECT * FROM "runs" WHERE ("runs"."name" = $1 AND "runs"."lifecycle_stage" <> $2) ORDER BY "runs"."run_uuid" LIMIT 1`,
+			name:  "TestRunNameWithoutFunction",
+			query: `(run.name == 'run')`,
+			expectedSQL: `SELECT "run_uuid" FROM "runs" ` +
```
dsuhinin, line 47: Jonathan it is hard to maintain all this `+` Why do we need them? We already use `backticks`.

pkg/api/aim/query/query.go
```diff
@@ -733,13 +733,23 @@ func (pq *parsedQuery) parseUnaryOp(node *ast.UnaryOp) (any, error) {
...
-	}
 	switch node.Op {
+	case ast.USub:
+		switch e := e.(type) {
+		case int:
+			return -e, nil
+		case float64:
+			return -e, nil
+		default:
+			return nil, fmt.Errorf("unsupported unary operation %q on %T", node.Op, e)
```
dsuhinin, line 744: maybe `unsupported` type?

dsuhinin (commented):
few comments.

pkg/api/aim/query/query_test.go
jgiannuzzi: the idea was to make this more readable, and to avoid lines that are too long

dsuhinin: correct. but `backticks` allows you to use new lines.

pkg/api/aim/query/query.go
jgiannuzzi: good point!
addressed in 5005490d6375e9d6856325dfb12d7b7cd29cd866

pkg/api/aim/query/query_test.go
jgiannuzzi: yes but we don't want new lines to be part of the string!

dsuhinin (approved):

PR #418, opened 2023-10-03
Fix issue with searching for metric names via regex

### LINKED ISSUE #414: Searching metric names with regex sometimes yields an error
No minimal reproducible example yet. Only happens with SQLite backend. Error message is  `argument must be BLOB or TEXT`.

jgiannuzzi:
Fixes #414.

There are also a few minor changes:
- an unnecessary call to `fmt.Sprintf` was removed
- SQL expressions have been made uppercase for consistency
- a few linting issues have been "fixed" by ignoring them as we stick to the Gorm conventions

I didn't add test cases to the integration tests to avoid merge conflicts with `features/namespaces`, but we should definitely cover this — see https://github.com/G-Research/fasttrackml/issues/414#issuecomment-1744744382 for how to reproduce.

### DIFF
--- a/pkg/api/aim/query/clause.go
+++ b/pkg/api/aim/query/clause.go
@@ -16,8 +16,10 @@ func (regexp Regexp) Build(builder clause.Builder) {
 	builder.WriteQuoted(regexp.Column)
 	switch regexp.Dialector {
 	case postgres.Dialector{}.Name():
+		// #nosec G104
 		builder.WriteString(" ~ ")
 	default:
+		// #nosec G104
 		builder.WriteString(" REGEXP ")
 	}
 	builder.AddVar(builder, regexp.Value)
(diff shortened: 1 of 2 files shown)


pkg/api/aim/query/clause.go
```diff
@@ -15,24 +14,41 @@ type Regexp struct {
...
+		builder.WriteString(" NOT REGEXP ")
 	}
 	builder.AddVar(builder, regexp.Value)
 }
+
+func (regexp Regexp) writeColumn(builder clause.Builder) {
+	switch regexp.Dialector {
+	case sqlite.Dialector{}.Name():
+		// #nosec G104
+		builder.WriteString("IFNULL(")
```
dsuhinin, line 47: so you casting `null` to be an empty string and only then apply `regexp`. ok.

dsuhinin (approved):

pkg/api/aim/query/clause.go
jgiannuzzi: correct

PR #420, opened 2023-10-03
code normalising according to linter remarks

dsuhinin:
What's inside the box:
- fixed all the linter remarks, so now when we run `golangci-lint run` we probably should see nothing.
- we are ready to integrate `golangci-lint run` into build pipeline.


pkg/api/aim/query/clause.go
```diff
@@ -18,8 +18,10 @@ func (regexp Regexp) Build(builder clause.Builder) {
 	builder.WriteQuoted(regexp.Column)
 	switch regexp.Dialector {
 	case postgres.Dialector{}.Name():
+		//nolint
```
jgiannuzzi, line 21: I have fixed in a more targeted way in #418

pkg/api/aim/encoding/decoder.go
```diff
@@ -43,6 +43,7 @@ func (d *reader) readField() ([]byte, error) {
 }
 
 // Decode decodes input stream of data into map[string]interface{}.
+// nolint:gocyclo
```
jgiannuzzi, line 46: we should probably leave a TODO on the `gocyclo` ones to properly fix them — the linter is right that many of them are way too complex

dsuhinin: yes yes. I've decided to also added this moment. I have other portion of changes where you will see such a TODO's

.golangci.yml
```diff
@@ -139,7 +144,7 @@ issues:
   # it can be disabled by `exclude-use-default: false`. To creators all
   # excluded by default patterns execute `golangci-lint run --help`
   exclude:
-    - abcdef
+    - declaration of "err" shadows declaration at line
```
dsuhinin, line 147: to avoid message related to `err`.

pkg/api/aim/query/query.go
```diff
@@ -408,6 +410,8 @@ func (pq *parsedQuery) parseList(node *ast.List) (any, error) {
 	return list, nil
 }
 
+// nolint:gocyclo
+// TODO:get back and ifx `gocyclo` problem.
```
suprjinx, line 414: ifx -> fix

suprjinx (dismissed):
Tough task -- thanks!

pkg/api/aim/query/query.go

jgiannuzzi (approved):
LGTM for ci.yml and .golangci.yml — I trust that the the rest is fine!

jgiannuzzi (approved):

PR #425, opened 2023-10-08
Add target for testing application with k6 to Makefile. Closes #260

### LINKED ISSUE #260: Create benchmark/performance tests for APIs
It would be good to have a performance test that we could track over multiple CI runs, in order to see if any particular changes affect speed. Would be good to run the same benchmarks against MLFlow tracking server, too.

Perhaps start with the logging API but also the UI-support endpoints.

asuzukosi:
Close #260 

Co-authored-by: Jonathan Gianuzzi @jgiannuzzi 

Add command to Makefile to perform k6 performance tests on FasttrackML application
The tests are run with the following command 
`make test-k6-performance`
from the root directory.

The command will restart the fml-service and then run the k6 performance tests which is located at `/tests/performance/k6_perf.js`
After the tests are complete it will clean all the docker images

### DIFF
--- a/pkg/api/aim/runs.go
+++ b/pkg/api/aim/runs.go
@@ -252,7 +252,7 @@ func GetRunsActive(c *fiber.Ctx) error {
 	}
 
 	var runs []database.Run
-	if tx := database.DB.
+	if err := database.DB.
 		Where("status = ?", database.StatusRunning).
 		InnerJoins(
 			"Experiment",
@@ -263,8 +263,10 @@ func GetRunsActive(c *fiber.Ctx) error {
 			),
 		).
 		Preload("LatestMetrics").
-		Find(&runs); tx.Error != nil {
-		return fmt.Errorf("error retrieving active runs: %w", tx.Error)
+		Limit(50).
+		Order("start_time DESC").
+		Find(&runs).Error; err != nil {
+		return fmt.Errorf("error retrieving active runs: %w", err)
 	}
 
 	c.Set("Content-Type", "application/octet-stream")


Makefile
```diff
@@ -147,6 +147,15 @@ test-python-integration-aim: build ## run the Aim python integration tests.
 	@echo ">>> Running Aim python integration tests."
 	tests/integration/python/aim/test.sh
 
+PHONY: test-k6-performance
+test-k6-performance: service-restart
+	@echo ">>> Running k6 performance tests."
+	@COMPOSE_FILE=$(COMPOSE_FILE) COMPOSE_PROJECT_NAME=$(COMPOSE_PROJECT_NAME) \
+	    docker-compose run performance-tests
+	@COMPOSE_FILE=$(COMPOSE_FILE) COMPOSE_PROJECT_NAME=$(COMPOSE_PROJECT_NAME) \
```
suprjinx, line 155: i think the "down" step isn't needed here (previously only in the `service-clean` target)

tests/integration/docker-compose.yml
```diff
@@ -73,6 +74,31 @@ services:
...
+        condition: service_healthy
+      postgres:
+        condition: service_healthy
+      service:
+        condition: service_healthy
+    environment:
+      GOCACHE: /cache/go-build
+      GOMODCACHE: /cache/go-mod
+      FML_SERVICE_URI: http://service:5000
+      FML_DATABASE_URI: ${FML_DATABASE_URI:-postgres://postgres:postgres@postgres/postgres}
```
suprjinx, line 96: Presumably for this test don't really need these environment vars -- except for FML_SERVICE_URI if used in the js file

Makefile
asuzukosi: Oh okay, I set it there to remove the containers once the process is complete, but that's fine, I will remove it

tests/integration/docker-compose.yml
asuzukosi: Thanks, removing them now.

tests/performance/k6_perf.js
```diff
@@ -0,0 +1,84 @@
+import http from 'k6/http';
+
+export default function () {
+  const base_url = 'http://service:5000/api/2.0/mlflow/';
```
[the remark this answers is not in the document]
asuzukosi: Yeah, would fix that now

PR #431, opened 2023-10-09
VSCode settings improvements

### LINKED ISSUE #491: VSCode settings to run integration tests
adjust VSCode settings so integration tests can run with devcontainer

suprjinx:
changes to support integration tests in vscode devcontainer and better launch server experience.

Fixes #491


.vscode/launch.json
```diff
@@ -31,24 +31,24 @@
             ],
         },
         {
-            "name": "Launch server (sqlite file)",
+            "name": "Launch server (sqlite file) - for integration tests",
```
jgiannuzzi, line 34: please don't bork this one but create another one instead — otherwise we won't have any developing against sqlite and restarting the dev container

.devcontainer/docker-compose.yml
```diff
@@ -76,5 +76,6 @@ services:
       MINIO_ROOT_USER: user
       MINIO_ROOT_PASSWORD: password
       MINIO_CONSOLE_ADDRESS: ":9001"
-    command: server /data
+    entrypoint: sh
+    command: -c 'mkdir -p /data/{bucket1,bucket2,bucket3} && exec minio server /data'
```
jgiannuzzi, line 80: please use the `postStartCommand` like we already do to create the `fasttrackml` bucket

suprjinx: this offers 1) max compatibility with other dev tooling and 2) matches what we do in `tests/integration/docker-compose.yaml'

.vscode/launch.json
suprjinx: will do

.devcontainer/docker-compose.yml
jgiannuzzi: but it also makes us totally inconsistent

jgiannuzzi: also, to address your points:
1) we do not _need_ to keep compatibility with other tooling than what we've described in the dev guide — `postStartCommand` is part of the dev container standard, and anyone using a tool that implements this standard will have their buckets created
2) this is actually really badly implemented and sometimes results in weird things in the integration tests — these buckets should be considered fixtures and should be created and removed by the integration tests

jgiannuzzi: our goal should actually be to address 2) and make this better

suprjinx: for 1) I just think putting it in docker-compose is compatible *both* with devcontainer and plain docker, whereas `postStartCommand` *only* works for devcontainer -- when something can be either more compatible or less with no impact to the devcontainer, why insist on less?

for 2) do you want to just remove this change here and add it to integration tests setup instead?

jgiannuzzi: 1) because it's not consistent (one bucket is created one way, and another is created another way) and it's a hack
2) yes please, it would make much more sense

jgiannuzzi: It is _not_ proprietary! It's part of an [open spec](https://containers.dev) implemented by a growing number of tools. It's not more or less open than the "standard" docker compose spec!

suprjinx: moved to test setup

tests/integration/golang/mlflow/artifact/list_s3_test.go
jgiannuzzi, line 1: great stuff!
we should also delete the buckets when the test is done

tests/integration/golang/mlflow/artifact/list_s3_test.go
```diff
@@ -41,9 +42,20 @@ func (s *ListArtifactS3TestSuite) SetupTest() {
...
 	assert.Nil(s.T(), err)
 
+	createBuckets(s3Client)
 	s.s3Client = s3Client
 }
 
+func createBuckets(s3Client *s3.Client) {
+	for _, bucket := range []string{"bucket1", "bucket2", "bucket3"} {
+		// ignore error if bucket already exists.
+		_, _ = s3Client.CreateBucket(context.Background(), &s3.CreateBucketInput{
```
jgiannuzzi, line 52: once we delete the buckets during the tear down, we can actually capture the error and have a more robust setup

tests/integration/golang/mlflow/artifact/list_s3_test.go
```diff
@@ -41,9 +42,20 @@ func (s *ListArtifactS3TestSuite) SetupTest() {
 	s3Client, err := helpers.NewS3Client(helpers.GetS3EndpointUri())
 	assert.Nil(s.T(), err)
 
+	createBuckets(s3Client)
```
dsuhinin, line 45: maybe we can put all of that into `BaseTestSuite` so we can:
- create `s3` client in centralised place.
- setup `buckets` in centralised place.
- we can use `TearDownSuite` to clean something we need.

tests/integration/golang/mlflow/artifact/list_s3_test.go
suprjinx: done!

tests/integration/golang/mlflow/artifact/list_s3_test.go
suprjinx: done!

tests/integration/golang/mlflow/artifact/list_s3_test.go
suprjinx: i moved to tests/integration/golang/helpers/s3.go

tests/integration/golang/helpers/s3.go
```diff
@@ -31,3 +35,51 @@ func NewS3Client(endpoint string) (*s3.Client, error) {
...
 }
+
+func CreateBuckets(s3Client *s3.Client) error {
+	for _, bucket := range testBuckets {
+		_, err := s3Client.CreateBucket(context.Background(), &s3.CreateBucketInput{
+			Bucket:                    aws.String(bucket),
+			CreateBucketConfiguration: &types.CreateBucketConfiguration{},
+		})
+		if err != nil {
+			return fmt.Errorf("failed to create bucket '%s': %v", bucket, err)
```
jgiannuzzi, line 46: please either use `%w` to wrap the error, or use `eris.Wrapf`

same remark further in this file

jgiannuzzi: actually, please use `eris` for consistency

tests/integration/golang/mlflow/artifact/get_artifact_s3_test.go
```diff
@@ -39,9 +39,16 @@ func (s *GetArtifactS3TestSuite) SetupTest() {
 	s3Client, err := helpers.NewS3Client(helpers.GetS3EndpointUri())
 	assert.Nil(s.T(), err)
 
+	//nolint:errcheck,gosec
+	helpers.CreateBuckets(s3Client)
```
jgiannuzzi, line 43: you're not checking the error (and the linter tells you about it before you silence it!)

tests/integration/golang/helpers/s3.go
```diff
@@ -31,3 +35,51 @@ func NewS3Client(endpoint string) (*s3.Client, error) {
...
+		// Delete all objects in the bucket
+		listObjectsOutput, err := s3Client.ListObjectsV2(context.Background(), &s3.ListObjectsV2Input{
+			Bucket: aws.String(bucket),
+		})
+		if err != nil {
+			return fmt.Errorf("failed to list objects in bucket '%s': %v", bucket, err)
+		}
+		var objectIds []types.ObjectIdentifier
+		for _, object := range listObjectsOutput.Contents {
+			objectIds = append(objectIds, types.ObjectIdentifier{Key: aws.String(*object.Key)})
```
jgiannuzzi, line 63: this call to `aws.String` is totally unnecessary

tests/integration/golang/helpers/s3.go
```diff
@@ -31,3 +35,51 @@ func NewS3Client(endpoint string) (*s3.Client, error) {
...
+			return fmt.Errorf("failed to create bucket '%s': %v", bucket, err)
+		}
+	}
+	return nil
+}
+
+func RemoveBuckets(s3Client *s3.Client) error {
+	for _, bucket := range testBuckets {
+		// Delete all objects in the bucket
+		listObjectsOutput, err := s3Client.ListObjectsV2(context.Background(), &s3.ListObjectsV2Input{
```
jgiannuzzi, line 55: please use pagination

tests/integration/golang/helpers/s3.go
```diff
@@ -2,13 +2,17 @@ package helpers
...
+	"fmt"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
 	awsConfig "github.com/aws/aws-sdk-go-v2/config"
 	"github.com/aws/aws-sdk-go-v2/service/s3"
+	"github.com/aws/aws-sdk-go-v2/service/s3/types"
 	"github.com/rotisserie/eris"
 )
 
+var testBuckets = []string{"bucket1", "bucket2", "bucket3"}
```
jgiannuzzi, line 14: I know this was copied from the docker compose file, but bucket3 is not used anywhere, so let's retire it

tests/integration/golang/mlflow/artifact/get_artifact_s3_test.go
suprjinx: this error is not worth checking -- if it can't create because bucket already exists the test will pass, if it fails for some other reason the PutObject for the bucket will fail the test later

tests/integration/golang/helpers/s3.go
suprjinx: for our integration test with 2 objects in each bucket?

tests/integration/golang/helpers/s3.go
suprjinx: 10-4

tests/integration/golang/helpers/s3.go
suprjinx: i think the sdk docs showed this usage

tests/integration/golang/helpers/s3.go
jgiannuzzi: Arguably our integration tests should have an additional test case to validate listing large amounts of artifacts that don't fit in a single page, so yes

tests/integration/golang/mlflow/artifact/get_artifact_s3_test.go
jgiannuzzi: What if it can't create the bucket, full stop? Then we'll need to hunt down the actual error. It's literally just adding an assert instead of a linter comment, what is there to win by not checking the error?

tests/integration/golang/helpers/s3.go
jgiannuzzi: Please also have a look at https://github.com/awsdocs/aws-doc-sdk-examples/blob/c2edcff1365d4b454b51075d632a1be844dd3e47/go/example_code/s3/s3_delete_bucket.go#L71 which shows that deleting a bucket is asynchronous and we thus need to check for completion.

tests/integration/golang/mlflow/artifact/get_artifact_s3_test.go
suprjinx: it's a conditional assert -- sometimes this is okay (already exists)

tests/integration/golang/helpers/s3.go
suprjinx: I don't think local Minio for integration tests, with with 1 item per bucket, requires this kind of bullet-proof approach

tests/integration/golang/helpers/s3.go
suprjinx: that seems like a premature optimization for bridge we have not crossed, but I will make the change

tests/integration/golang/helpers/s3.go
suprjinx: will do

tests/integration/golang/helpers/s3.go
suprjinx: AFAICT the `s3manager` used in the SDK examples you linked is not available in our V2 s3 packages, so I'd suggest we stick with "list-then-delete-objects" approach here using just the s3 client. 
```gwilson@macbook ~/P/f/t/i/g/m/artifact (integration-tests-devcontainer) [1]> go get github.com/aws/aws-sdk-go-v2/service/s3/s3manager
go: module github.com/aws/aws-sdk-go-v2/service/s3@upgrade found (v1.40.2), but does not contain package github.com/aws/aws-sdk-go-v2/service/s3/s3manager
```
I can add pagination on list and wait for delete.

suprjinx: done

tests/integration/golang/helpers/s3.go
suprjinx: done

tests/integration/golang/helpers/s3.go
suprjinx: done

tests/integration/golang/mlflow/artifact/get_artifact_s3_test.go
suprjinx: well it may fail at first until the teardown succeeds, but I think simple assert will work after that.

tests/integration/golang/helpers/s3.go
```diff
@@ -31,3 +35,71 @@ func NewS3Client(endpoint string) (*s3.Client, error) {
...
+	}); err != nil {
+		return eris.Wrapf(err, "failed to delete bucket '%s'", bucket)
+	}
+	waiter := s3.NewBucketNotExistsWaiter(s3Client)
+	if err := waiter.Wait(
+		context.Background(),
+		&s3.HeadBucketInput{
+			Bucket: aws.String(bucket),
+		},
+		time.Second,
```
jgiannuzzi, line 100: We should wait for longer than a second here, to avoid tests failing because of some disk/cpu/memory performance issue on a GitHub Actions hosted runner.
I'd suggest 10 seconds to be on the safe side, as in most cases this will be close to instant.

jgiannuzzi (commented):
This is looking great @suprjinx 🤩
Just one more tiny comment.

tests/integration/golang/helpers/s3.go
suprjinx: will do

jgiannuzzi (approved):

PR #432, opened 2023-10-10
Search syntax documentation

### LINKED ISSUE #327: Documentation of search syntax

fabiovincenzi:
fixes #327


docs/search_syntax.md
```diff
@@ -0,0 +1,228 @@
...
+| ```name```  | Run name | ```string```|
+| ```hash``` | Run hash | ```string```|
+| ```experiment``` | Experiment name | ```string```|
+| ```tags``` | List of run tags |```dictionary```|
+| ```archived``` | True if run is archived, otherwise False | ```boolean```|
+| ```active``` | True if run is active(in progress), otherwise False | ```boolean``` |
+| ```duration``` | Run duration in seconds | ```numeric```|
+| ```created_at``` | Run creation datetime | ```numeric``` |
+| ```finalized_at``` | Run end datetime | ```numeric``` |
+| ```metrics``` | Set of run metrics | ```dictionary``` |
```
suprjinx, line 20: can these be given table-like format in the markdown source, too? ie, lining up the columns

docs/search_syntax.md
```diff
@@ -0,0 +1,228 @@
...
+
+Select only the runs where the archived attribute is not true
+```python
+run.archived != True
+```
+
+### Run parameters
+Run parameters could be accessed via attributes.
+```python
+run.hparams.param1 >= 64
```
dsuhinin, line 135: typo? `hparams`? should be just `params`.

docs/search_syntax.md
```diff
@@ -0,0 +1,228 @@
+# Search Syntax Documentation
+
+## Introduction
```
dsuhinin, line 3: what do you think about to have a `glossary` with links so you can easily jump to any section? not a requirement but just a thought if I want to quickly check special section.

docs/search_syntax.md
```diff
@@ -0,0 +1,228 @@
...
+```python
+metric.last <= 1.1
+```
+
+### Filter Metrics by run
+You can also filter the metrics by combining  metric attributes with run attributes.
+
+Showing the metrics with the last value greater than 6 belonging to a run with the name that starts with marvelous.
+
+![FastTrackML Metric List filter by metric and run attributes](images/metrics_filter_by_run.png)
```
dsuhinin, line 228: what do you think to a like a complex query with different operators and cases?

dsuhinin (commented):
What do you think about to add `search_` prefix to each `png` file so it will be much easy to navigate and understand to which `doc` this images belong?

docs/search_syntax.md
fabiovincenzi: Yes! I'll add it

docs/search_syntax.md
fabiovincenzi: Yep

docs/search_syntax.md
fabiovincenzi: Actually should be just param1 I'll fix that

suprjinx (approved):
looks great

docs/search_syntax.md
```diff
@@ -0,0 +1,281 @@
...
+| ---------------- | ------------- |
+| ```name```       | ```string```  |
+| ```last```       | ```numeric``` |
+| ```last_step```  | ```numeric``` |
+| ```first_step``` | ```numeric``` |
+
+## Operations
+
+### String operations
+For the ```string``` attributes you can use the following comparing operator:
```
jgiannuzzi, line 56:
```suggestion
For the ```string``` attributes you can use the following comparison operator:
```

(there are multiples places where this comment applies)

docs/search_syntax.md
```diff
@@ -0,0 +1,281 @@
...
+For the ```numeric``` attributes you can use the following comparing operator:
+- ``` == ```
+- ``` != ```
+- ``` > ```
+- ``` >= ```
+- ``` < ```
+- ``` <= ```
+
+### Boolean operations
+For the ```boolean``` attributes you can use the following comparing operator:
```
jgiannuzzi, line 73: Please also document that having no operator or a unary operation also works for booleans, e.g. `run.archived` is the same as `run.archived == True`, and `!run.archived` is the same as `run.archived == False`

docs/search_syntax.md
```diff
@@ -0,0 +1,281 @@
...
+```
+
+Select only the runs where the name ends with "Run1"
+```python
+run.name.endswith('Run1')
+```
+
+### Example with ```run.duration``` (numeric)
+
+Select only the runs where the duration is exactly 111111111
```
jgiannuzzi, line 116: Use something more realistic, like 60

docs/search_syntax.md
```diff
@@ -0,0 +1,281 @@
...
+run.archived == True
+```
+
+Select only the runs where the archived attribute is not true
+```python
+run.archived != True
+```
+
+### Run parameters
+Run parameters could be accessed via attributes.
```
jgiannuzzi, line 160:
```suggestion
Run parameters can be accessed via attributes.
```

docs/search_syntax.md
```diff
@@ -0,0 +1,281 @@
...
+![FastTrackML Run List, param filter](images/search_runs_param_filter.png)
+
+### Filtering Runs with Unset Attributes
```
jgiannuzzi, line 163:
```suggestion
### Filtering Runs with Unset Parameters
```

(there are multiples places where this comment applies)

docs/search_syntax.md
```diff
@@ -0,0 +1,281 @@
...
+
+Showing all the runs
+![FastTrackML Run List](images/search_runs.png)
+
+Showing only the runs where param1 is not set
+![FastTrackML Run List of not set param](images/search_runs_none_param.png)
+
+### Filter Runs using Regular Expressions
+- ``` .match() ```
+- ``` .search() ```
```
jgiannuzzi, line 180:
```suggestion
- ``` re.match() ```
- ``` re.search() ```
```

docs/search_syntax.md
```diff
@@ -0,0 +1,281 @@
...
+Search looks for a pattern anywhere in the string.
+![FastTrackML Run filter using regular expression match](images/search_runs_regular_expression_search.png)
+
+### Complex query for run search
+The query selects the runs that meet the following conditions:
+
+
+- run.archived can be either True or False.
+- The duration of run must be greater than 0.
+- The run has to contain a metric named 'TestMetric' and is value of last  must be greater than 2.5.
```
jgiannuzzi, line 194:
```suggestion
- The run has to contain a metric named 'TestMetric' and its last recorded value must be greater than 2.5.
```

docs/search_syntax.md
```diff
@@ -0,0 +1,281 @@
+# Search Syntax Documentation
+## Introduction
+
+This document provides an overview of the search syntax that users can utilize to filter metrics and runs.
+
```
jgiannuzzi, line 5: We should start by introducing that the syntax is essentially Python "so it should feel pretty intuitive for most users", and then explain that we'll dive into more details about supported operations below.

docs/search_syntax.md
```diff
@@ -0,0 +1,281 @@
...
+  - [Complex query for run search](#complex-query-for-run-search)
+- [Search metrics examples](#search-metrics-examples)
+  - [Example with metric.name (string)](#example-with-metricname-string)
+  - [Example with metric.last (numeric)](#example-with-metriclast-numeric)
+  - [Filter Metrics by run](#filter-metrics-by-run)
+  - [Complex query for metric search](#complex-query-for-metric-search)
+
+
+## Search Runs
+You can filter the runs using the following run attributes:
```
jgiannuzzi, line 30: We should make it clear that these are attributes for the `run` object

docs/search_syntax.md
```diff
@@ -0,0 +1,281 @@
...
+| ```tags```         | List of run tags                                    | ```dictionary``` |
+| ```archived```     | True if run is archived, otherwise False            | ```boolean```    |
+| ```active```       | True if run is active(in progress), otherwise False | ```boolean```    |
+| ```duration```     | Run duration in seconds                             | ```numeric```    |
+| ```created_at```   | Run creation datetime                               | ```numeric```    |
+| ```finalized_at``` | Run end datetime                                    | ```numeric```    |
+| ```metrics```      | Set of run metrics                                  | ```dictionary``` |
+
+## Search Metrics
+You can filter the metrics using the following metric attributes:
```
jgiannuzzi, line 45: We should make it clear that these are attributes for the `metric` object

jgiannuzzi (commented):
Great stuff!

PR #443, opened 2023-10-12
Search syntax documentation

### LINKED ISSUE #327: Documentation of search syntax

fabiovincenzi:
fixes #327


suprjinx (approved):
LGTM

docs/search_syntax.md
```diff
@@ -0,0 +1,294 @@
...
+```
+This expression will return runs for which the specified attribute is not defined.
+
+Showing all the runs
+![FastTrackML Run List](images/search_runs.png)
+
+Showing only the runs where param1 is not set
+![FastTrackML Run List of not set param](images/search_runs_none_param.png)
+
+### Filter Runs using Regular Expressions
```
jgiannuzzi, line 191: regex is not specific to runs, it works for any string (e.g. `metric.name`) — it may unclear to users that it can work anywhere, maybe we should add regex ops to the string ops at the top and add an example of filtering metrics using regex?

docs/search_syntax.md
```diff
@@ -0,0 +1,294 @@
...
+| ```metric.last_step```  | ```numeric``` |
+| ```metric.first_step``` | ```numeric``` |
+
+## Operations
+
+### String operations
+For the ```string``` attributes you can use the following comparison operator:
+- ``` == ```
+- ``` != ```
+- ``` in ```
```
jgiannuzzi, line 60: `in` can also be used for checking whether something is in a list, e.g.
```python
run.experiment in ["my-first-experiment", "my-second-experiment"]
```

docs/search_syntax.md
```diff
@@ -0,0 +1,294 @@
...
+
+Select only the runs where the duration is greater or equal to 60
+```python
+run.duration >= 60
+```
+
+Select only the runs where the duration is less than 60
+```python
+run.duration < 60
+```
```
jgiannuzzi, line 154: we shouldn't list _all_ the numerical comparison operators here — let's just give 2 or 3 examples, like `==`, `>=`, and `<`

docs/search_syntax.md
```diff
@@ -0,0 +1,294 @@
...
+
+Select only the metrics where the name starts with "Test"
+```python
+metric.name.startswith('Test')
+```
+
+Select only the metrics where the name ends with "Metric1"
+```python
+metric.name.endswith('Metric1')
+```
```
jgiannuzzi, line 240: we should list _all_ the string comparison operators here, especially as we gave them for runs too — maybe we just keep `==`, `startswith`, and `re.search`

docs/search_syntax.md
jgiannuzzi: same remark for the `metric.last` example

docs/search_syntax.md
fabiovincenzi: Actually regex are not working at the moment with metric search

jgiannuzzi (approved):
great stuff, thanks @fabiovincenzi 👍

PR #451, opened 2023-10-13
Use default experiment per namespace

### LINKED ISSUE #459: Interpret default experiment per namespace
When a tracking uri is set for a namespace, but no experiment is given, the MLFlow client provides experiment ID "0". We should interpret this as "default for the namespace".

Additionally, when reporting experiments back to the client, the actual default experiment ID should be replaced with "0", so they are consistent.

suprjinx:
Fixes #459 

When a namespace is provided but an experiment is not provided when creating a run, mlflow client supplies experiment ID 0, which belongs only to the default namespace. Use the default experiment ID of that namespace instead.


pkg/api/mlflow/service/run/service.go
```diff
@@ -78,9 +78,18 @@ func (s Service) CreateRun(
...
+	// experimentID 0 should be interpreted as the default for the namespace
+	if experimentID == 0 {
+		experimentID = int64(*ns.DefaultExperimentID)
+	}
+
 	experiment, err := s.experimentRepository.GetByNamespaceIDAndExperimentID(ctx, ns.ID, int32(experimentID))
 	if err != nil {
-		return nil, api.NewResourceDoesNotExistError("unable to find experiment with id '%s': %s", req.ExperimentID, err)
+		return nil, api.NewResourceDoesNotExistError(
+			"unable to find experiment for namespace with id '%d': %s",
```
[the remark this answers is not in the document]
suprjinx: well we can reword but I thought it was important to indicate the experiment may in fact exist but not in the namespace

pkg/ui/admin/controller/namespaces.go
```diff
@@ -111,7 +111,7 @@ func (c Controller) renderIndex(ctx *fiber.Ctx, msg string) error {
 		return ctx.Render("namespaces/index", fiber.Map{
 			"Namespaces": namespaces,
 			"Status":     StatusError,
-			"Message":    err.Error(),
+			"Message":    common.ErrorMessageForUI("namespace", err.Error()),
```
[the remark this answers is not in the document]
suprjinx: the error messages are pretty ugly and are always of the same type, APIError

tests/integration/golang/helpers/client.go
```diff
@@ -79,6 +79,12 @@ func (c *HttpClient) WithRequest(request any) *HttpClient {
 	return c
 }
 
+// WithBasePath sets base path.
+func (c *HttpClient) WithBasePath(basePath string) *HttpClient {
```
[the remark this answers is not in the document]
suprjinx: will do

pkg/api/mlflow/service/run/service.go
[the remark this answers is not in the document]
suprjinx: reworded for clarity

pkg/api/mlflow/service/run/service.go
```diff
@@ -78,9 +78,19 @@ func (s Service) CreateRun(
...
+	}
+
 	experiment, err := s.experimentRepository.GetByNamespaceIDAndExperimentID(ctx, ns.ID, int32(experimentID))
 	if err != nil {
-		return nil, api.NewResourceDoesNotExistError("unable to find experiment with id '%s': %s", req.ExperimentID, err)
+		return nil, api.NewResourceDoesNotExistError(
+			"unable to find experiment '%d' for namespace '%s': %s",
+			experimentID,
+			ns.Code,
+			err,
```
jgiannuzzi, line 92: I would actually argue against changing this error message, unless we change this for _everything_

pkg/ui/admin/controller/namespaces.go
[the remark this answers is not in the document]
jgiannuzzi: but they are (or at least should be!) wrapping the underlying `pkg/api/mlflow/api.ErrorResponse` error, which has an `ErrorCode` field that you actually want to look at (or at least that's what I think)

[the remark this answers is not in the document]
suprjinx: [okay new story for that](https://github.com/G-Research/fasttrackml/issues/517)

jgiannuzzi (commented):
what about SearchRuns and other methods that take an experiment ID?

suprjinx:
> what about SearchRuns and other methods that take an experiment ID?

I'll take a look -- i was only trying to get logging to work with a NS

suprjinx:
> what about SearchRuns and other methods that take an experiment ID?

It looks like MLflow is providing the actual experiment IDs here since a previous api call fetched

jgiannuzzi:
if you try `mlflow.search_runs()` it will search for runs in experiment_id `"0"`

tests/integration/golang/helpers/client.go
```diff
@@ -109,12 +117,18 @@ func (c *HttpClient) DoRequest(uri string, values ...any) error {
 		requestBody = bytes.NewBuffer(data)
 	}
 
-	// 2. build actual URL.
-	u, err := url.Parse(fmt.Sprintf("%s%s%s", c.baseURL, c.basePath, fmt.Sprintf(uri, values...)))
+	// 2. build path with namespace.
+	path := c.basePath
```
dsuhinin, line 121: it is a bit strange moment. since now we have more logic related to `url` contractions, won't it be reasonable to do like this:
```
path = []string{c.baseURL}
if c.namespace != "" {
               path = append(path, "ns",  c.namespace, c.basePath)
 	}
uri, err := url.Joins(append(path, fmt.Sprintf(uri, values...)))
```
so I mean to construct final `url` like we do everywhere?

pkg/api/mlflow/service/run/converters.go
```diff
@@ -0,0 +1,24 @@
...
+
+import (
+	"fmt"
+
+	"github.com/G-Research/fasttrackml/pkg/api/mlflow/api/request"
+	"github.com/G-Research/fasttrackml/pkg/api/mlflow/dao/models"
+)
+
+// adjustSearchRunsRequestForNamespace preprocesses the SearchRunRequest for the given namespace.
+func adjustSearchRunsRequestForNamespace(ns *models.Namespace, srr *request.SearchRunsRequest) {
```
dsuhinin, line 11: maybe `normalise` more meaningful name?

pkg/api/mlflow/service/run/service.go
```diff
@@ -73,6 +73,7 @@ func NewService(
 func (s Service) CreateRun(
 	ctx context.Context, ns *models.Namespace, req *request.CreateRunRequest,
 ) (*models.Run, error) {
+	adjustCreateRunRequestForNamespace(ns, req)
```
dsuhinin, line 76: Im not sure about this function at all. we already have `ConvertCreateRunRequestToDBModel` which converts actual input `request` into `db` model, so for me it looks reasonable to put logic into `ConvertCreateRunRequestToDBModel` instead of have another function.

pkg/api/mlflow/service/run/service.go
```diff
@@ -156,6 +157,7 @@ func (s Service) GetRun(
 func (s Service) SearchRuns(
 	ctx context.Context, namespace *models.Namespace, req *request.SearchRunsRequest,
 ) ([]models.Run, int, int, error) {
+	adjustSearchRunsRequestForNamespace(namespace, req)
```
dsuhinin, line 160: since we have normal `builder` functions which will convert `req` into `db` query, here it looks temporary reasonable to have it.

dsuhinin: maybe it is reasonable to firstly validate request and then do all the adjustments?

tests/integration/golang/mlflow/run/create_with_namespace_test.go
```diff
@@ -0,0 +1,192 @@
...
+		{
+			name:      "CreateRunWithNotExistingNamespaceAndExistingExperimentID",
+			namespace: "custom-ns1",
+			request: request.CreateRunRequest{
+				ExperimentID: fmt.Sprintf("%d", *experiment.ID),
+			},
+			error: api.NewResourceDoesNotExistError(
+				`unable to find namespace with code: custom-ns1`,
+			),
+		},
```
dsuhinin: can you please remove this test case from here? Ill add it in scope of my tests related to `namespace` integration tests.

dsuhinin (changes requested):

pkg/api/mlflow/service/run/converters.go
suprjinx: not clear what you mean, can you suggest a name?

pkg/api/mlflow/service/run/service.go
suprjinx: I thought about this and it's a good suggestion, but I thought it would be best to make the step very explicit with a separate named method.

pkg/api/mlflow/service/run/converters_test.go
```diff
@@ -0,0 +1,93 @@
...
+	"github.com/G-Research/fasttrackml/pkg/api/mlflow/common"
+	"github.com/G-Research/fasttrackml/pkg/api/mlflow/dao/models"
+)
+
+func Test_adjustSearchRunsRequestForNamespace_Ok(t *testing.T) {
+	testData := []struct {
+		name           string
+		ns             *models.Namespace
+		srr            *request.SearchRunsRequest
+		expectedResult *request.SearchRunsRequest
```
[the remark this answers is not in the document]
suprjinx: well it's the expected result of the function, right? given inputs, we expect this result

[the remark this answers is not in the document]
suprjinx: renamed to `inputRequest` and `resultRequest`

pkg/api/mlflow/service/run/converters.go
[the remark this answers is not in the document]
suprjinx: done

pkg/api/mlflow/service/run/converters.go
```diff
@@ -0,0 +1,24 @@
...
+	for i, expID := range srr.ExperimentIDs {
+		if expID == "0" {
+			srr.ExperimentIDs[i] = fmt.Sprintf("%d", *ns.DefaultExperimentID)
+		}
+	}
+}
+
+// adjustCreateRunRequestForNamespace preprocesses the CreateRunRequest for the given namespace.
+func adjustCreateRunRequestForNamespace(ns *models.Namespace, crr *request.CreateRunRequest) {
```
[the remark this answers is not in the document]
suprjinx: done

pkg/api/mlflow/service/run/converters_test.go
```diff
@@ -0,0 +1,93 @@
...
+		})
+	}
+}
+
+func Test_adjustCreateRunRequestForNamespace_Ok(t *testing.T) {
+	testData := []struct {
+		name           string
+		ns             *models.Namespace
+		crr            *request.CreateRunRequest
+		expectedResult *request.CreateRunRequest
```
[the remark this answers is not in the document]
suprjinx: done

pkg/api/mlflow/service/run/service.go
suprjinx: okay, moved below the validation step

pkg/api/mlflow/service/run/service.go
suprjinx: additionally, in this function there is a reliance on finding the correct experiment to provide as a parameter to `ConvertCreateRunRequestToDBModel` so I needed to make the adjustment before entering the converter. If we move this adjustment inside the Converter, and it lets us do this adjustment in a uniform way across functions (others do not have ConvertToDBModel).

pkg/api/mlflow/service/run/service.go
suprjinx: done

suprjinx:
> if you try `mlflow.search_runs()` it will search for runs in experiment_id `"0"`

Added support for experiment_id rewriting in SearchRuns and GetMetricHistories -- I think now covering all cases where client can provide an experiment ID.

tests/integration/golang/mlflow/run/create_with_namespace_test.go
suprjinx: keeping for now

pkg/api/mlflow/service/run/converters.go
dsuhinin: yes, yes -> `NormaliseSearchRunsRequestForNamespace`

tests/integration/golang/mlflow/run/create_test.go
```diff
@@ -105,17 +199,31 @@ func (s *CreateRunTestSuite) Test_Ok() {
...
+		NamespaceID:    namespace.ID,
+		LifecycleStage: models.LifecycleStageActive,
+	})
+	assert.Nil(s.T(), err)
+
+	// set namespace default experiment.
+	namespace.DefaultExperimentID = experiment.ID
+	_, err = s.NamespaceFixtures.UpdateNamespace(context.Background(), namespace)
+	assert.Nil(s.T(), err)
+	nonExistingExperimentID := *experiment.ID + 1
```
dsuhinin, line 220: this could really confuse -> `+1`. I mean why not just put anything into `nonExistingExperimentID` without such a strange calculation? just `-1` or `99999` or anything else?

tests/integration/golang/mlflow/run/create_test.go
```diff
@@ -67,19 +155,25 @@ func (s *CreateRunTestSuite) Test_Ok() {
...
 	resp := response.CreateRunResponse{}
+	client := s.MlflowClient.WithMethod(
+		http.MethodPost,
+	).WithRequest(
+		req,
+	).WithResponse(
+		&resp,
+	)
+	if useNamespaceInRequest {
+		client = client.WithNamespace(
```
dsuhinin, line 170: you don't need to assign to `client`. will be enough only have:
````
client.WithNamespace
```

tests/integration/golang/mlflow/run/create_test.go
```diff
@@ -54,6 +54,94 @@ func (s *CreateRunTestSuite) Test_Ok() {
...
+
+	// update default experiment id.
+	namespace.DefaultExperimentID = experiment.ID
+	_, err = s.NamespaceFixtures.UpdateNamespace(context.Background(), namespace)
+	assert.Nil(s.T(), err)
+
+	s.testCases(namespace, experiment, true, int32(0))
+}
+
+func (s *CreateRunTestSuite) testCases(
```
dsuhinin, line 139: the naming is a bit strange of such function. is hard to say what is going on here and why do we need it.

pkg/api/mlflow/service/run/service.go
dave-gantenbein: This seems like a reasonable need to have a separate function.

tests/integration/golang/mlflow/run/create_test.go
```diff
@@ -128,27 +236,62 @@ func (s *CreateRunTestSuite) Test_Error() {
...
+			name:      "CreateRunWithNotExistingNamespaceAndExistingExperimentID",
+			namespace: "not_existing_namespace",
 			request: request.CreateRunRequest{
-				ExperimentID: "1",
+				ExperimentID: fmt.Sprintf("%d", *experiment.ID),
 			},
 			error: api.NewResourceDoesNotExistError(
-				`unable to find experiment with id '1': error getting experiment by id: 1: record not found`,
+				`unable to find namespace with code: not_existing_namespace`,
+			),
```
dsuhinin, line 246: probably this test tests `namespace` functionality and `middleware` itself. We don't need to have this test here It will be included into my PR related to `namespace` integration tests.

tests/integration/golang/mlflow/run/create_test.go
```diff
@@ -128,27 +236,62 @@ func (s *CreateRunTestSuite) Test_Error() {
...
+				fmt.Sprintf(
+					`unable to find experiment with id '%d': error getting experiment by id: %d: record not found`,
+					nonExistingExperimentID,
+					nonExistingExperimentID,
+				),
+			),
+		},
+		{
+			name:      "CreateRunWithExistingNamespaceAndNotExistingExperiment",
+			namespace: "default",
```
dsuhinin, line 264: you can probably avoid `namespace: default`?

tests/integration/golang/mlflow/run/create_test.go
```diff
@@ -128,27 +236,62 @@ func (s *CreateRunTestSuite) Test_Error() {
...
+				ExperimentID: fmt.Sprintf("%d", *experiment.ID),
 			},
 			error: api.NewResourceDoesNotExistError(
-				`unable to find experiment with id '1': error getting experiment by id: 1: record not found`,
+				`unable to find namespace with code: not_existing_namespace`,
+			),
+		},
+		{
+			name:      "CreateRunWithNotExistingExperiment",
+			namespace: "default",
```
dsuhinin, line 250: you can probably avoid `namespace: default`?

tests/integration/golang/mlflow/run/create_test.go
```diff
@@ -128,27 +236,62 @@ func (s *CreateRunTestSuite) Test_Error() {
...
+			).WithRequest(
+				tt.request,
+			).WithResponse(
+				&resp,
+			)
+			if tt.namespace != "" {
+				client = client.WithNamespace(
+					tt.namespace,
+				)
+			}
```
dsuhinin, line 291: so if you remove test which I mentioned and remove `namespace: default` so in that case you don't need to have such check.

tests/integration/golang/mlflow/run/create_test.go
dsuhinin: also I don't understand what this function tests? it just creates `Run` and that's it. We don't test `error` or `response` or that `run` has been actually created.

pkg/api/mlflow/service/run/converters_test.go
```diff
@@ -0,0 +1,93 @@
...
+			inputRequest: &request.CreateRunRequest{
+				ExperimentID: "456",
+			},
+			resultRequest: &request.CreateRunRequest{
+				ExperimentID: "456",
+			},
+		},
+	}
+
+	for _, tc := range testData {
```
dsuhinin, line 87: in most cases we use `tt`. can you please unify that?

dsuhinin (changes requested):

tests/integration/golang/mlflow/run/create_test.go
suprjinx: well this is ensuring that whatever ID was just assigned for the existing experiment could not possibly collide with our non-existent id. I think PG tests are using sequence for ID which continues to increase on every run even when we truncate tables.

tests/integration/golang/mlflow/run/create_test.go
suprjinx: that seemed to be the idea of "fluent" methods like this, that they return the struct you are to use.

pkg/api/mlflow/service/run/converters.go
suprjinx: I don't think Normalise quite fits this situation

tests/integration/golang/mlflow/run/create_test.go
suprjinx: I was just adapting the existing test for namespace inputs

tests/integration/golang/mlflow/run/create_test.go
suprjinx: i was running into some issues, possibly related to caching, where omitting the namespace did not select the default but rather the previously used namespace.

pkg/api/mlflow/service/run/converters_test.go
suprjinx: sure

tests/integration/golang/mlflow/run/create_test.go
suprjinx: Well the change in this file was to take the test logic from the old Test_OK and make it reusable with different namespace/experiment inputs. I didn't really change the interior of those tests.

tests/integration/golang/mlflow/run/create_test.go
suprjinx: okay "caching" because we are reusing the client across tests, it saved the namespace field. Changed this to always set namespace to test case field value, which clears it when omitted.

tests/integration/golang/mlflow/run/create_test.go
suprjinx: removed

tests/integration/golang/mlflow/run/create_test.go
dave-gantenbein: Just call it "testSuccessCases" and we can always improve it later.

tests/integration/golang/mlflow/run/create_test.go
dave-gantenbein: I'm vetoing the scope creep here.

tests/integration/golang/mlflow/run/create_test.go
suprjinx: named it "successCases"

dave-gantenbein: Hooray

dave-gantenbein (approved):
Everything seems resolved, ship it!

PR #453, opened 2023-10-14
return valid http status(404) code when requested namespace was not found

dsuhinin:


jgiannuzzi (commented):
We used to handle ResourceDoesNotExist errors in the fiber error handlers and set the status code accordingly. I didn't check the entire code base, but how come it has to be set manually here?

dsuhinin:
> We used to handle ResourceDoesNotExist errors in the fiber error handlers and set the status code accordingly. I didn't check the entire code base, but how come it has to be set manually here?

that's true, but in middleware `fiber` works a bit strange with errors. When I tried to return `NotFound` error like we do in many other places I just got always this:
Screenshot 2023-10-14 at 14 39 35
as you can see it is not a valid `json` object and status wrong, should be `404` instead of `500`.

Then I've checked theirs predefined `middlewares` like this -> https://github.com/gofiber/fiber/blob/master/middleware/basicauth/basicauth.go#L34

so they just set everything manually -> https://github.com/gofiber/fiber/blob/master/middleware/basicauth/config.go#L95

so that's why we can't use predefined errors like we use them in our `service` layers -> just return it and it will be parced correctly.

dsuhinin:
> We used to handle ResourceDoesNotExist errors in the fiber error handlers and set the status code accordingly. I didn't check the entire code base, but how come it has to be set manually here?

jgiannuzzi:
Ah I see! This is good enough for this PR, but we’ll need to refactor this further:
- there is duplicated code between getting _a_ namespace and the getting the _default_ namespace
- We should return JSON only when it’s a API call, and return an html page when it’s a UI call, ideally the chooser, but with an error message instead of the selected namespace and the 2 UIs (so we keep header, list of namespaces, and footer)

jgiannuzzi (approved):

suprjinx (approved):

fabiovincenzi (approved):

PR #452, opened 2023-10-13
Change CI release logic

jgiannuzzi:
This PR changes the CI release logic in the following way:
* add a workflow `create.yml` to create release branches and tags in a controlled manner, using a GitHub app that will have exceptions in the rulesets
* change the `release.yml` workflow to be directly called by the main CI workflow
* change `release.yml` to not validate the ref anymore, but instead rely on rulesets and environments
* add support for prereleases
* document workflow in `docs/maintainer.md`

---

The following tasks need to be completed before merging this PR (more information can be found in the documentation):
- [x] create and install a new GitHub app
- [x] add its credentials to a new `create-release` environment that can only be accessed by the `main` branch
- [x] add the required rulesets — this _**has**_ to be done before the next step!
- [x] change the `release` environment to also allow `v*` tag patterns (and keep the existing `main` branch)

---

Examples on a repo that has been setup as described and as "upstream" (so _not_ a fork, but in this case a copy):
> [!NOTE]
> Many of these links may require to be reloaded before they display the actual log line I want to show — this seems to be a GitHub bug at the time of writing
- [Run on `main`](https://github.com/jgiannuzzi-org/fasttrackml/actions/runs/6516940465): notice in particular that the `main` and `edge` tags are [published on Docker Hub](https://github.com/jgiannuzzi-org/fasttrackml/actions/runs/6516940465/job/17701004503#step:6:1)
- [Run to create a `0.3` release branch](https://github.com/jgiannuzzi-org/fasttrackml/actions/runs/6516981234): notice in particular that the app/bot is able to [bypass the branch creation rule](https://github.com/jgiannuzzi-org/fasttrackml/actions/runs/6516981234/job/17701043283#step:5:8) to create the branch
  image
- [Run on the `0.3` release branch](https://github.com/jgiannuzzi-org/fasttrackml/actions/runs/6516983137): notice that the [release workflow is skipped](https://github.com/jgiannuzzi-org/fasttrackml/actions/runs/6516983137/job/17701141666)
- [Run to create a `0.3.5-beta.1` tag](https://github.com/jgiannuzzi-org/fasttrackml/actions/runs/6517063046): the app/bot can [bypass the tag creation rule](https://github.com/jgiannuzzi-org/fasttrackml/actions/runs/6517063046/job/17701193768#step:5:7)
  image
[...]

PR #458, opened 2023-10-16
Improve required checks creation in CI

jgiannuzzi:
This PR improves how we create the required check "All required checks succeeded" in the following ways:
* We now use a GitHub app to create the check instead of using the token provided by GitHub Actions.
  This allows us to create the check run with its own check suite instead of attaching it to the first GHA check suite, which changes when a pull request is closed and reopened, or when a new run attempt is made. See https://github.com/orgs/community/discussions/24616#discussioncomment-6088422 for more details on this very annoying GitHub limitation.
* A check is created and set as `queued` or `in_progress` when the corresponding workflow run's status changes.
  This is needed so that the check is reset when a new run attempt is made or a PR is reopened.
* Logging is improved.
* Start/Completion times are added.

---

The following tasks need to be completed **_before_** merging this PR:
- [x] create and install a new GitHub app with the `check:write` permissions
- [x] add its credentials to a new `create-check` environment that can only be accessed by the `main` branch

---

The following tasks need to be completed **_after_** merging this PR:
- [ ] update branch protection (or ruleset, if #452 has been merged in the meantime) to enforce the required check to originate from the new GitHub app

### DIFF
--- a/.github/workflows/check-required.yml
+++ b/.github/workflows/check-required.yml
@@ -4,11 +4,14 @@ name: Check required jobs
 # It checks if the "All required checks done" job was actually successful
 # (and not just skipped) and creates a check run if that is the case. The
 # check run can be used to protect the main branch from being merged if the
-# CI is not passing.
+# CI is not passing. We need to use a GitHub app token to create the check
+# run because otherwise the check suite will be assigned to the first workflow
+# run for the CI, which might not be the latest one. See 
+# https://github.com/orgs/community/discussions/24616#discussioncomment-6088422
+# for more details.
 
 on:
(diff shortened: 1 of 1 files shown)


suprjinx (approved):
I wish Github CI did not require so much custom code -- seems like never ending maintenance if Armada Project is any guide.

jgiannuzzi:
> I wish Github CI did not require so much custom code -- seems like never ending maintenance if Armada Project is any guide.

After 4 years of using GitHub Actions, I wish so too 😞
However, it does keep improving, and we do have a particularly stringent security posture, so 🤷

PR #472, opened 2023-10-17
Support Google Storage for artifact logging 

dsuhinin:
What's inside the box:
- new `gs` adapter has been added to support Google Storage for artifact logging.


pkg/api/mlflow/service/artifact/storage/storage.go
jgiannuzzi, line 1: General remark about this factory now that we're adding a 3rd storage: I _think_ we should simplify it, as there is a lot of duplicated code. We could do this:
* try to load and return
* if not found, look at the name (switch/case) and create accordingly
* store and return
I don't think there are real benefits in checking whether we got the right kind of storage from the map — something would be incredibly wrong in our code for this to happen, and simplifying it as suggested above should help prevent that

pkg/api/mlflow/service/artifact/storage/gs.go
```diff
@@ -0,0 +1,107 @@
...
+	}
+	return &GS{
+		client: client,
+	}, nil
+}
+
+// List implements ArtifactStorageProvider interface.
+func (s GS) List(ctx context.Context, artifactURI, path string) ([]ArtifactObject, error) {
+	// 1. process input parameters.
+	bucket, rootPrefix, err := ExtractS3BucketAndPrefix(artifactURI)
```
jgiannuzzi, line 41: we should rename this method

dsuhinin: renamed it to be a bit more generic.

pkg/api/mlflow/service/artifact/storage/helpers.go
```diff
@@ -7,11 +7,12 @@ import (
 	"github.com/rotisserie/eris"
 )
 
-// ExtractS3BucketAndPrefix extracts bucket name and prefix path from provided uri.
-// after processing of this uri s3://fasttrackml/2/30357ed2eaac4f2cacdbcd0e06e9e48a/artifacts result will be:
+// ExtractBucketAndPrefix extracts bucket name and prefix path from provided uri.
+// uri could be any S3 compatible storage uri like; AWS S3 or Google Storage.
```
jgiannuzzi, line 11: tiny typo:
```suggestion
// uri could be any S3 compatible storage uri like AWS S3 or Google Storage.
```

pkg/api/mlflow/service/artifact/storage/storage.go
dsuhinin: addressed.

pkg/api/mlflow/service/artifact/storage/storage.go
```diff
@@ -71,47 +71,34 @@ func (s *ArtifactStorageFactory) GetStorage(
...
 		if err != nil {
 			return nil, eris.Wrap(err, "error initializing local artifact storage")
 		}
-		s.storageList.Store(LocalStorageName, storage)
-		return storage, nil
 	default:
 		return nil, eris.Errorf("unsupported schema has been provided: %s", u.Scheme)
 	}
+
+	s.storageList.Store(S3StorageName, storage)
```
jgiannuzzi, line 102: oops! ❌ your're registering everything as S3!!!!

dsuhinin: yep, yep. fixed. my bad.

pkg/api/mlflow/service/artifact/storage/helpers.go
dsuhinin: done.

jgiannuzzi (dismissed):
LGTM 👍
I also tested it locally and can confirm it works as expected.

jgiannuzzi:
@dsuhinin could you please also add `google-cloud-storage` to the `Pipfile`?
```python
pipenv install --dev google-cloud-storage
```

dsuhinin:
> @dsuhinin could you please also add `google-cloud-storage` to the `Pipfile`?
> 
> ```python
> pipenv install --dev google-cloud-storage
> ```

yep. done!

jgiannuzzi:
it looks like the `Pipfile.lock` did not get updated!

dsuhinin:
> it looks like the `Pipfile.lock` did not get updated!

yep, yep. didn't commit it by some reason. now should be good.

jgiannuzzi (approved):

PR #476, opened 2023-10-17
Stop gracefully on SIGTERM

### LINKED ISSUE #445: Kubernetes support: exit cleanly on SIGTERM
When running a FastTrackML server in Kubernetes and the pod gets terminated, we don't exit cleanly, which causes the SQLite WAL file to not be removed.

jgiannuzzi:
Fixes #445.

### DIFF
--- a/pkg/cmd/server.go
+++ b/pkg/cmd/server.go
@@ -6,6 +6,7 @@ import (
 	"os"
 	"os/signal"
 	"strings"
+	"syscall"
 	"time"
 
 	"github.com/gofiber/fiber/v2"
@@ -124,7 +125,7 @@ func serverCmd(cmd *cobra.Command, args []string) error {
 	isRunning := make(chan struct{})
 	go func() {
 		sigint := make(chan os.Signal, 1)
-		signal.Notify(sigint, os.Interrupt)
+		signal.Notify(sigint, os.Interrupt, syscall.SIGTERM)
 		<-sigint
 
 		log.Infof("Shutting down")


pkg/cmd/server.go
```diff
@@ -125,6 +127,9 @@ func serverCmd(cmd *cobra.Command, args []string) error {
 	go func() {
 		sigint := make(chan os.Signal, 1)
 		signal.Notify(sigint, os.Interrupt)
+		if runtime.GOOS != "windows" {
+			signal.Notify(sigint, syscall.SIGTERM)
```
dsuhinin, line 131: why you can't just use:
```
signal.Notify(sigint, os.Interrupt,syscall.SIGTERM )
```
instead? os one of this events will trigger shutdown.

jgiannuzzi: Windows has no concept of signals and so this wouldn't be portable AFAIK

dsuhinin: yes yes, I know, but it won't just work in that case?

dsuhinin: I mean you can just use both of them in one `signal.Notify` and one notification will be triggered based on current system. this is what I meant.

jgiannuzzi: hmm, I thought we would get errors when using `signal.Notify` on signals that don't exist on the runtime platform, but I just tested it on Windows and it works!

dsuhinin (approved):

PR #485, opened 2023-10-18
Implement (again) database reset for SQLite

jgiannuzzi:
The infamous database refactor removed the ability to reset the SQLite database. This PR restores this ability, whilst also finalizing the intended refactoring (based on TODOs). The `--database-pool-max` is also now used in the SQLite database provider.

The SQLite reset is not the most efficient now that it has to be done _after_ opening the connection, as we cannot simply delete the files anymore. We could potentially change that by having `database.Reset(dsn string)` be responsible for calling the provider's implementation, and letting the provider do its logic. Thoughts?


suprjinx (approved):
looks reasonable even if slower than file delete. my practice has always been to set a different db filename in DSN when I want a reset, which is workaround for extra speed.

PR #488, opened 2023-10-18
Implement namespace caching layer.

dsuhinin:
What's inside the box:

- caching layer for namespace logic.
- attach namespace middleware to only mlflow and aim routes instead of global router.


pkg/api/mlflow/dao/listener.go
```diff
@@ -0,0 +1,99 @@
...
+	return &eventListener, nil
+}
+
+// NewNamespaceListener creates new database event listener for Namespace entity.
+func NewNamespaceListener(ctx context.Context, db *gorm.DB) (*EventListener, error) {
+	return NewEventListener(ctx, db, "namespace_update_events")
+}
+
+// Listen listens for incoming database events.
+func (l EventListener) Listen() <-chan string {
```
suprjinx, line 74: i think "l" isn't the best local name since it looks like "1" , etc.

pkg/api/mlflow/dao/repositories/namespace_cached.go
```diff
@@ -0,0 +1,186 @@
...
+)
+
+// NamespaceEvent represents database event.
+type NamespaceEvent struct {
+	Action    NamespaceEventAction `json:"action"`
+	Namespace *models.Namespace    `json:"namespace"`
+}
+
+// NamespaceCachedRepository cached repository to work with `namespace` entity.
+type NamespaceCachedRepository struct {
```
suprjinx, line 35: nice wrapper 👍

pkg/api/mlflow/dao/repositories/namespace_cached.go
```diff
@@ -0,0 +1,186 @@
...
+	}
+
+	namespace, err := r.namespaceRepository.GetByCode(ctx, code)
+	if err != nil {
+		return nil, eris.Wrapf(err, "error getting cached namespace by code: %s", code)
+	}
+
+	// trigger database event to notify current instance and
+	// other instances to add record to theirs local cache.
+	if err := r.sendEvent(NamespaceEventActionFetched, namespace); err != nil {
```
suprjinx, line 120: maybe we don't need this event publish for "Get"? the other instances will update anyway when they try to fetch, or if they have it already it is unchanged.

pkg/api/mlflow/dao/repositories/namespace_cached.go
```diff
@@ -0,0 +1,186 @@
...
+		return nil, eris.Wrap(err, "error sending database event")
+	}
+	return namespace, nil
+}
+
+// GetByID returns namespace by its ID.
+func (r NamespaceCachedRepository) GetByID(ctx context.Context, id uint) (*models.Namespace, error) {
+	return r.namespaceRepository.GetByID(ctx, id)
```
suprjinx, line 128: not checking cache here?

pkg/api/mlflow/dao/repositories/namespace_cached.go
suprjinx: i see later that your are using same notifications for local cache as well as remote -- so probably nevermind :-)

pkg/api/mlflow/dao/repositories/namespace_cached.go
```diff
@@ -0,0 +1,186 @@
...
+	return nil
+}
+
+// List returns all namespaces.
+func (r NamespaceCachedRepository) List(ctx context.Context) ([]models.Namespace, error) {
+	return r.namespaceRepository.List(ctx)
+}
+
+// processEvent process incoming event from database.
+func (r NamespaceCachedRepository) processEvent(data string) error {
```
suprjinx, line 151: the event subscription logic is pretty complex here -- I think it should get unit test.

pkg/api/mlflow/dao/listener.go
dsuhinin: sure. renamed to be `el`

pkg/api/mlflow/dao/repositories/namespace_cached.go
dsuhinin: thank you!

pkg/api/mlflow/dao/repositories/namespace_cached.go
dsuhinin: nope unfortunately, because:
- GetByID used mostly in UI.
- GetByCode used in middleware, so it means that each call to the service will call this method so we have to cache it.
- we store `namespace` in `cache` using `namespace` id.
actually right now I think when we call `GetByID` we can store result inside the cache.

pkg/api/mlflow/dao/repositories/namespace_cached.go
dsuhinin: that's true. unfortunately I don't know how to test it via `unit` tests, because we have to mock this moment:
```
if err := r.db.Exec(
		fmt.Sprintf(`SELECT pg_notify('%s', '%s')`, r.listener.GetChannelName(), data),
	).Error; err != nil {
		return eris.Wrap(err, "error triggering 'pg_notify'")
	}
```
I think it is doable via `integration` tests. added this task #498 as a reminder.

pkg/api/mlflow/dao/listener.go
```diff
@@ -0,0 +1,99 @@
...
+	// if listener not nil, then listen for incoming events from database.
+	// if listener is nil, then just return closed channel to do not do anything further.
+	if el.connection != nil {
+		go func() {
+			defer close(ch)
+			for {
+				notification, err := el.connection.Conn().WaitForNotification(context.Background())
+				if err != nil {
+					log.Errorf("error occurred while listening for the event: %+v", err)
+					return
```
suprjinx, line 85: should we not return here but instead continue? it seems like one connection error will stop the listener currently

dsuhinin: probably if `err` exists, it means that connected broken so we have 2 ways:
- log and stop listening - like we do right now.
- try to reconnect. for example try to re connect N times and then log an error.

dsuhinin: the second item looks like a goo improvement which we can do in separate task.

suprjinx: can you make ticket and link here if it's not being addressed now

dsuhinin: sure. done -> #518

suprjinx (approved):
LGTM

pkg/api/mlflow/dao/repositories/namespace.go
```diff
@@ -17,7 +17,7 @@ type NamespaceRepositoryProvider interface {
 	// Update modifies the existing models.Namespace entity.
 	Update(ctx context.Context, namespace *models.Namespace) error
 	// Delete removes a namespace and it's associated experiments by its ID.
```
fabiovincenzi, line 19: Could you change the comment as well?

dsuhinin: changed all the comments.

pkg/api/mlflow/dao/models/namespace.go
```diff
@@ -8,12 +8,12 @@ import (
...
 type Namespace struct {
-	ID                  uint   `gorm:"primaryKey;autoIncrement"`
-	Code                string `gorm:"unique;index;not null"`
-	Description         string
-	CreatedAt           time.Time
-	UpdatedAt           time.Time
-	DeletedAt           gorm.DeletedAt `gorm:"index"`
-	DefaultExperimentID *int32         `gorm:"not null"`
-	Experiments         []Experiment   `gorm:"constraint:OnDelete:CASCADE"`
+	ID                  uint           `gorm:"primaryKey;autoIncrement" json:"ID"`
```
suprjinx, line 11: why are we using custom json encoding for this instead of the defaults?

dsuhinin: it happened, because of `linter` wants to see them, so that's why I just added them. we serialise this object when send it via `pg_notify`.
Screenshot 2023-11-07 at 16 28 20

suprjinx (approved):
lgtm

fabiovincenzi (approved):
LGTM

PR #490, opened 2023-10-18
Miscellaneous namespace improvements

dsuhinin:
This PR improves our namespace implementation in the following ways:
* Inner joins are used to confine the queries into the selected namespace (instead of left joins)
* More thorough joins to avoid having to issue separate queries
* Better error handling
* The namespace and its ID are not returned via the Aim Apps API anymore


suprjinx (approved):
LGTM

pkg/api/aim/experiments.go
```diff
@@ -173,20 +174,19 @@ func GetExperimentRuns(c *fiber.Ctx) error {
 		run := &database.Run{
 			ID: q.Offset,
 		}
-		if err = database.DB.Select("row_num").First(&run).Error; err != nil && err != gorm.ErrRecordNotFound {
+		if err = database.DB.Select("row_num").First(&run).Error; err != nil && errors.Is(err, gorm.ErrRecordNotFound) {
```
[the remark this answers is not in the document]
dsuhinin: emmmmmmmmmmmm. that's strange:
Screenshot 2023-10-19 at 08 55 18
it seems like but I can't see this changes in github. strange.

pkg/api/aim/dashboards.go
```diff
@@ -151,12 +163,14 @@ func UpdateDashboard(c *fiber.Ctx) error {
 		},
 	}
 	if err := database.DB.
-		Joins("LEFT JOIN apps ON dashboards.app_id = apps.id").
+		InnerJoins(
+			"App",
+			database.DB.Where(&database.App{NamespaceID: ns.ID}, "NamespaceID"),
```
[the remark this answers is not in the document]
dsuhinin: got it!

jgiannuzzi (dismissed):

pkg/api/aim/dashboards.go
```diff
@@ -106,13 +112,19 @@ func GetDashboard(c *fiber.Ctx) error {
...
+			"App",
+			database.DB.Select(
+				"ID", "Type",
+			).Where(
+				&database.App{NamespaceID: ns.ID}, "NamespaceID",
+			),
+		).
 		Where("NOT dashboards.is_archived").
-		Preload("App").
 		Joins("LEFT JOIN apps ON apps.id = dashboards.app_id").
```
jgiannuzzi, line 124: This `Joins` needs to be removed

dsuhinin: nope. this is wasn't a problem unfortunately.

pkg/api/aim/dashboards.go
```diff
@@ -22,10 +23,11 @@ func GetDashboards(c *fiber.Ctx) error {
 
 	var dashboards []database.Dashboard
 	if err := database.DB.
-		Preload("App").
-		Joins("LEFT JOIN apps ON apps.id = dashboards.app_id").
+		Joins(
+			"INNER JOIN apps ON dashboards.app_id = apps.id AND apps.namespace_id = ?",
+			ns.ID,
+		).
```
jgiannuzzi, line 29: the app type is now missing because it's not properly joined

pkg/api/aim/apps.go
jgiannuzzi, line 1: calls to GetApps/GetApp are returning the Namespace in the json response, which it shouldn't:
```json
{
    "id": "d5010047-44ad-4600-8b92-4e7261260408",
    "created_at": "2023-10-24T10:10:52.230560695Z",
    "updated_at": "2023-10-24T10:10:52.230560695Z",
    "type": "metrics",
    "state": {
        "...": {}
    },
    "Namespace": {
        "ID": 0,
        "Apps": null,
        "Code": "",
        "Description": "",
        "CreatedAt": "0001-01-01T00:00:00Z",
        "UpdatedAt": "0001-01-01T00:00:00Z",
        "DeletedAt": null,
        "DefaultExperimentID": null,
        "Experiments": null
    },
    "NamespaceID": 1
}

**Review by jgiannuzzi — CHANGES_REQUESTED** [review]

File: pkg/api/aim/dashboards.go (hunk 4 above)
Comment (dsuhinin) [reply pkg/api/aim/dashboards.go]: fixed.

File: pkg/api/aim/apps.go
Comment (dsuhinin) [reply pkg/api/aim/apps.go:1]: added `json:"-"` to avoid everything related to `namespace`.

File: pkg/api/aim/dashboards.go [hunk 5]
```diff
@@ -197,12 +230,18 @@ func DeleteDashboard(c *fiber.Ctx) error {
 	}
 	if err := database.DB.
 		Select("dashboards.id").
-		Joins("LEFT JOIN apps ON dashboards.app_id = apps.id").
+		InnerJoins(
+			"App",
+			database.DB.Where(
+				&database.App{
```
dsuhinin, line 236: @jgiannuzzi it is a bit strange I know but:
```
Base: database.Base{
 						IsArchived: false,
 					},
```
leads Python tests failing. Can't say right now why.

jgiannuzzi: can you please investigate what the Python tests do that our integration tests don't and add it to integration tests?

dsuhinin: @jgiannuzzi can we do that in scope of other task? Im not sure that it could be super fast, since I never touch the Python tests part?

pkg/api/aim/dashboards.go
```diff
@@ -22,10 +23,18 @@ func GetDashboards(c *fiber.Ctx) error {
 
 	var dashboards []database.Dashboard
 	if err := database.DB.
-		Preload("App").
-		Joins("LEFT JOIN apps ON apps.id = dashboards.app_id").
+		InnerJoins(
+			"App",
+			database.DB.Select(
+				"ID", "Type", "IsArchived",
```
jgiannuzzi, line 29: We don't need to select `IsArchived` here

pkg/api/aim/dashboards.go
```diff
@@ -106,13 +116,21 @@ func GetDashboard(c *fiber.Ctx) error {
 		},
 	}
 	if err := database.DB.
+		InnerJoins(
+			"App",
+			database.DB.Select(
+				"ID", "Type", "IsArchived",
```
jgiannuzzi, line 122: same remark as above

pkg/api/aim/dashboards.go
```diff
@@ -151,18 +169,28 @@ func UpdateDashboard(c *fiber.Ctx) error {
 		},
 	}
 	if err := database.DB.
-		Joins("LEFT JOIN apps ON dashboards.app_id = apps.id").
+		InnerJoins(
+			"App",
+			database.DB.Select(
+				"NamespaceID",
```
jgiannuzzi, line 175: I think we would be better off selecting `ID` and `Type` instead, just so we are extra safe in _not_ updating the app further. We already use `Omit("App")` to avoid it, and with this we'll be even safer.

pkg/api/aim/dashboards.go
```diff
@@ -197,18 +225,28 @@ func DeleteDashboard(c *fiber.Ctx) error {
 	}
 	if err := database.DB.
 		Select("dashboards.id").
-		Joins("LEFT JOIN apps ON dashboards.app_id = apps.id").
+		InnerJoins(
+			"App",
+			database.DB.Select(
+				"NamespaceID",
```
jgiannuzzi, line 231: same remark as above

pkg/api/aim/dashboards.go
dsuhinin: yep, it was just before. I can remove it.

dsuhinin: done.

pkg/api/aim/dashboards.go
dsuhinin: done

pkg/api/aim/dashboards.go
dsuhinin: ook!

pkg/api/aim/dashboards.go
dsuhinin: oook!

jgiannuzzi (approved):

PR #492, opened 2023-10-18
Fix wrong Y axis scale in Aim UI when metric has infinity values

### LINKED ISSUE #489: AIM ignore NaN metric values for Y axis limits?
User Stephen Chestnut reports: "if a metric has some runs with nan and some runs with numeric values then the plot in metrics explorer just looks empty, and with no y-axis ticks, until I specify the y-range using the "Axis-properties" button.  Is possible to make AIM ignore nans when it auto-selects the axes limits?"

jgiannuzzi:
The Aim UI knows how to deal with such values but we were not mapping them correctly in the encoded result, as MLFlow's database storage uses MaxFloat64 instead of infinity. This PR adds a mapping back from MaxFloat64 to proper infinity.

Fixes #489

### DIFF
--- a/pkg/api/aim/runs.go
+++ b/pkg/api/aim/runs.go
@@ -1140,6 +1140,12 @@ func DeleteBatch(c *fiber.Ctx) error {
 func toNumpy(values []float64) fiber.Map {
 	buf := bytes.NewBuffer(make([]byte, 0, len(values)*8))
 	for _, v := range values {
+		switch v {
+		case math.MaxFloat64:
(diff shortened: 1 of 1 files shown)


dave-gantenbein (approved):
I wish I could ask for a unit test here, but this whole thing looks like its up for refactor soon.

jgiannuzzi:
> I wish I could ask for a unit test here, but this whole thing looks like its up for refactor soon.

indeed it is 😄

PR #493, opened 2023-10-18
Restore functionality to update the website when a release is created

jgiannuzzi:
Now that `release.yml` can run directly under the context of a tag (#452), the workflow that updates the website after a release doesn't work anymore — and if it did, it could deploy an older version of the website!
This PR fixes that by trigger a workflow run on the default branch via `workflow_dispatch`, rather than reusing the workflow on the same reference via `workflow_call`. Yes, GitHub Actions can get pretty arcane 😬


dave-gantenbein (approved):
Ship it!

PR #505, opened 2023-10-20
Separate package for each migration

### LINKED ISSUE #486: Migration failure from MLFlow database in `main`
```
ERRO[0000]/workspaces/fasttrackml/pkg/database/logger.go:75 github.com/G-Research/fasttrackml/pkg/database.(*loggerAdaptor).Trace() SQL error                                     app_file="/go/pkg/mod/gorm.io/gorm@v1.25.5/migrator/migrator.go:363" app_func="gorm.io/driver/sqlite.(*Migrator).AddColumn.Migrator.AddColumn.func1()" elapsed=0.008ms error="duplicate column name: namespace_id" rows=0 sql="ALTER TABLE `apps` ADD `namespace_id` integer"
DEBU[0000]/workspaces/fasttrackml/pkg/database/logger.go:81 github.com/G-Research/fasttrackml/pkg/database.(*loggerAdaptor).Trace() SQL trace                                     app_file="/go/pkg/mod/github.com/jgiannuzzi/gorm-sqlite@v1.4.4-0.20221215225833-42389ad31305/migrator.go:27" app_func="gorm.io/driver/sqlite.(*Migrator).RunWithoutForeignKey()" elapsed=0.005ms rows=1 sql="PRAGMA foreign_keys = ON"
FATA[0000]/workspaces/fasttrackml/main.go:12 main.main() error running database migration: error migrating database to FastTrackML schema e0d125c68d9a: duplicate column name: namespace_id
```

suprjinx:
Migration steps need to have a static/unchanging copy of the model from the time the migration was created. This way, as the contemporary model evolves, older schema can still be migrated.

By putting each migration and it's model in a separate package, we keep a historical copy of model and also keep the size of the master database/migrate.go to reasonable size.
Fixes #486


pkg/database/migrate.go
```diff
@@ -195,6 +197,28 @@ func CheckAndMigrateDB(migrate bool, db *gorm.DB) error {
 			case "ac0b8b7c0014":
 				log.Info("Migrating database to FastTrackML schema 8073e7e037e5")
 				if err := db.Transaction(func(tx *gorm.DB) error {
+					// types for migration
+					type Base struct {
+						ID         uuid.UUID `gorm:"type:uuid;primaryKey" json:"id"`
+						CreatedAt  time.Time `json:"created_at"`
```
jgiannuzzi, line 203: We don’t need any of the json stuff

pkg/database/migrate.go
```diff
@@ -195,6 +197,28 @@ func CheckAndMigrateDB(migrate bool, db *gorm.DB) error {
...
+						UpdatedAt  time.Time `json:"updated_at"`
+						IsArchived bool      `json:"-"`
+					}
+					type Dashboard struct {
+						Base
+						Name        string     `json:"name"`
+						Description string     `json:"description"`
+						AppID       *uuid.UUID `gorm:"type:uuid" json:"app_id"`
```
jgiannuzzi, line 211: Any reason for this to be a pointer?

pkg/database/migrate.go
suprjinx: good point :-)

pkg/database/migrate.go
suprjinx: that's just how it was previously defined

jgiannuzzi: ah I see — that's only necessary for the JSON serialization, but let's keep it like this to avoid any confusion with the past model

pkg/database/migrate.go
```diff
@@ -195,6 +197,27 @@ func CheckAndMigrateDB(migrate bool, db *gorm.DB) error {
...
+						Name        string
+						Description string
+						AppID       *uuid.UUID `gorm:"type:uuid"`
+						App         App
+					}
+					type App struct {
+						Base
+						Type  string `gorm:"not null"`
+						State AppState
+					}
```
jgiannuzzi, line 218: to avoid any confusion, could you please give these types new local names that don't shadow the package types?

jgiannuzzi: also you are missing `AppState` — though you could simply give `State` its real historic type of `map[string]any`

suprjinx: I agree with both of these comments, but both require interface implementation methods on the structs in order to work -- and Go doesn't like the definitions to occur here afaict.

suprjinx: well maybe i can do the AppState one without the GormDataType() method. checking...

tests/integration/golang/database/mlflow.db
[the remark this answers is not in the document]
jgiannuzzi: running a python script with a given version of MLFlow (and we can have all the ones we deem relevant) and a DB backend (we can go for all the backends we support)

this is definitely stuff for another PR — please remove the migration test for now

pkg/database/migrate.go
jgiannuzzi: ah right, I forgot this one was custom 🤦 — just go for `string` then!

jgiannuzzi: so I looked at this further and using those local types is super dangerous could bite us in so many places

I would recommend going for a package per migration that just contains the contents of `model.go` as it was at that time. this will be more verbose but it will ensure we have properly self-contained historical types and keep the benefits of automatic SQL generation from gorm

e.g. something like `pkg/database/migrations/e0d125c68d9a/model.go` would currently be a copy of `pkg/database/model.go`, and the migration would look like:

```go
case "5d042539be4f":
	log.Info("Migrating database to FastTrackML schema e0d125c68d9a")
	// We need to run this migration without foreign key constraints to avoid
	// the cascading delete to kick in and delete all the runs.
	if err := runWithoutForeignKeyIfNeeded(func() error {
		if err := db.Transaction(func(tx *gorm.DB) error {
			if err := tx.AutoMigrate(&e0d125c68d9a.Namespace{}); err != nil {
				return err
			}
			if err := tx.Migrator().AddColumn(&e0d125c68d9a.App{}, "NamespaceID"); err != nil {
				return err
			}
			if err := tx.Migrator().CreateConstraint(&e0d125c68d9a.Namespace{}, "Apps"); err != nil {
				return err
			}
			if err := tx.Migrator().AddColumn(&e0d125c68d9a.Experiment{}, "NamespaceID"); err != nil {
				return err
			}
			if err := tx.Migrator().CreateConstraint(&e0d125c68d9a.Namespace{}, "Experiments"); err != nil {
				return err
			}
			if err := tx.Migrator().AlterColumn(&e0d125c68d9a.Experiment{}, "Name"); err != nil {
				return err
			}
			if err := tx.Migrator().CreateIndex(&e0d125c68d9a.Experiment{}, "idx_namespace_name"); err != nil {
				return err
			}
			return tx.Model(&e0d125c68d9a.SchemaVersion{}).
				Where("1 = 1").
				Update("Version", "e0d125c68d9a").
				Error
		}); err != nil {
			return fmt.Errorf("error migrating database to FastTrackML schema e0d125c68d9a: %w", err)
		}
		return nil
	}); err != nil {
		return err
	}
```

dave-gantenbein: The proposed solution here looks really complex. Can we investigate using a technology more suited to doing migrations (ie ActiveRecord) so we aren't maintaining such baroque code?

jgiannuzzi: We should indeed look into that, but doing what I mentioned above took me 5 minutes of research and could be implemented in 30 minutes so I'd advocate to do it now so we can at least avoid the migration issues that we have repeatedly had

jgiannuzzi: What @suprjinx currently proposed will keep creating issues unfortunately (e.g. the local types reference current models and then create foreign keys on them)

suprjinx: okay, I will do this -- I think we can put the migrator code in the `e0d125c68d9a` module as well, so it's more self-contained. I'll just do the two AutoMigrate cases for now

suprjinx: done

tests/integration/golang/database/mlflow.db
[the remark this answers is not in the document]
suprjinx: why not keep this test? it is actually providing a good validation that we don't otherwise have.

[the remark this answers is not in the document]
jgiannuzzi: Because we're both adding binary blobs whose provenance cannot be attested

pkg/database/migrate.go
```diff
@@ -325,33 +316,7 @@ func CheckAndMigrateDB(migrate bool, db *gorm.DB) error {
 				// We need to run this migration without foreign key constraints to avoid
 				// the cascading delete to kick in and delete all the runs.
 				if err := runWithoutForeignKeyIfNeeded(func() error {
```
jgiannuzzi, line 318: this is part of the migration

pkg/database/migrate.go
```diff
@@ -325,33 +316,7 @@ func CheckAndMigrateDB(migrate bool, db *gorm.DB) error {
...
-						}
-						if err := tx.Migrator().CreateIndex(&Experiment{}, "idx_namespace_name"); err != nil {
-							return err
-						}
-						return tx.Model(&SchemaVersion{}).
-							Where("1 = 1").
-							Update("Version", "e0d125c68d9a").
-							Error
-					}); err != nil {
+					if err := db.Transaction(migration_5d042539be4f.Migrate); err != nil {
```
jgiannuzzi, line 319: this is the wrong migration name!

jgiannuzzi (changes requested):
We should do this for all the FML migrations at least. Please don't cut corners by doing it just for these two now that we've established that migration issues could pop up from anywhere.
Also please move these all into a common parent package `migrations` like I originally suggested (`pkg/database/migrations/<hash>/{migrate,model}.go` — so `pkg/database` is not polluted by all the migrations)

pkg/database/migrate.go
jgiannuzzi: better tooling tracked in https://github.com/G-Research/fasttrackml/issues/520

suprjinx:
@jgiannuzzi hash alone does not work as package name for these since they are disallowed when they start with numeral.

> We should do this for all the FML migrations at least. Please don't cut corners by doing it just for these two now that we've established that migration issues could pop up from anywhere. Also please move these all into a common parent package `migrations` like I originally suggested (`pkg/database/migrations/<hash>/{migrate,model}.go` — so `pkg/database` is not polluted by all the migrations)

pkg/database/migrate.go
suprjinx: Sorry, thanks for catching!

pkg/database/migrate.go
suprjinx: @jgiannuzzi will you look at the current implementation and see if that looks good?

tests/integration/golang/database/mlflow.sql
```diff
@@ -0,0 +1,161 @@
+PRAGMA foreign_keys=OFF;
```
jgiannuzzi, line 1: there is no mention of the MLFlow version that this schema comes from. I'm sorry but this is a really poor implementation, and I would much rather have a proper test that can create databases from various versions of MLFlow (and cache them) - we could easily do this with e.g. dagger so that it's cached and parallelised

tests/integration/golang/database/migrate_test.go
```diff
@@ -0,0 +1,40 @@
...
+	"time"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/G-Research/fasttrackml/pkg/database"
+)
+
+func TestMigrate(t *testing.T) {
+	// setup sqlite MLFlow database from the schema
+	mlflowDBPath := fmt.Sprintf("%s/mlflow.db", t.TempDir())
```
jgiannuzzi, line 19: please use proper path handling

pkg/database/migrations/common.go
```diff
@@ -0,0 +1,22 @@
...
+
+import (
+	"gorm.io/driver/sqlite"
+	"gorm.io/gorm"
+)
+
+const (
+	SQLiteDialectorName   = "sqlite"
+	PostgresDialectorName = "postgres"
+)
```
jgiannuzzi, line 11: these constants are now declared in 2 places...

jgiannuzzi: also we should get rid of them and use more robust definitions like `sqlite.Dialector{}.Name()` and `postgres.Dialector{}.Name()` — but this is for another PR

jgiannuzzi (dismissed):

pkg/database/migrate.go
```diff
@@ -44,83 +53,31 @@ func CheckAndMigrateDB(migrate bool, db *gorm.DB) error {
...
-					if err := tx.Migrator().CreateIndex(table, "RunID"); err != nil {
-						return err
-					}
-				}
-				return tx.Model(&AlembicVersion{}).
-					Where("1 = 1").
-					Update("Version", "bd07f7e963c5").
-					Error
-			}); err != nil {
+			if err := v_bd07f7e963c5.Migrate(db); err != nil {
```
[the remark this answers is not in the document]
suprjinx: sorry, I went with the git sha from database/migrate.go for the migration section, but it must have had a later change on the lines.

tests/integration/golang/database/mlflow.sql
suprjinx: would adding a version help this? I don't think the perfect should be the enemy of the good here, and the perfect will take some time to develop.

jgiannuzzi: adding a version would be the bare minimum, yes — and if you go for just 1 version on 1 backend, please at least use the oldest version of MLFlow we support so that the test is relevant

pkg/database/migrations/common.go
suprjinx: this was to avoid an import cyclo, so i'll need to find a new package instead of `database` for a shared location

jgiannuzzi: then just go for the good definitions I mentioned above — we should probably not have introduced these constants in the first place

tests/integration/golang/database/migrate_test.go
suprjinx: done

tests/integration/golang/database/mlflow.sql
suprjinx: now labeled and testing oldest supported and most recent MLflow schema

pkg/database/migrate.go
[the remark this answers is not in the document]
suprjinx: i think we're good for historical models now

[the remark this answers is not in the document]
dave-gantenbein: I reviewed the history w/Geoff and it looks nice and tidy now.

tests/integration/golang/database/mlflow.sql
dave-gantenbein: Reviewed requested change, LGTM

dave-gantenbein (approved):
Reviewed last round of changed, GTG

PR #513, opened 2023-10-24
Fix pagination for SearchRuns endpoint in Aim

dsuhinin:


pkg/api/aim/runs.go
```diff
@@ -412,8 +412,8 @@ func SearchRuns(c *fiber.Ctx) error {
 			ID: q.Offset,
 		}
 		// TODO:DSuhinin -> do we need `namespace` restriction? it seems like yyyyess, but ....
-		if tx = database.DB.Select("row_num").First(&run); tx.Error != nil && tx.Error != gorm.ErrRecordNotFound {
-			return fmt.Errorf("unable to find search runs offset %q: %w", q.Offset, tx.Error)
+		if err = database.DB.Select("row_num").First(&run).Error; err != nil && tx.Error != gorm.ErrRecordNotFound {
```
jgiannuzzi, line 415: you're using tx.Error in the if

jgiannuzzi: also let's make sure we assign to a new `err` to be on the safe side (`if err := ...`)

dsuhinin: Im not finished it yet.

dsuhinin: ok, now it is ready for review.

pkg/api/aim/runs.go
```diff
@@ -412,8 +413,8 @@ func SearchRuns(c *fiber.Ctx) error {
 			ID: q.Offset,
 		}
 		// TODO:DSuhinin -> do we need `namespace` restriction? it seems like yyyyess, but ....
-		if tx = database.DB.Select("row_num").First(&run); tx.Error != nil && tx.Error != gorm.ErrRecordNotFound {
-			return fmt.Errorf("unable to find search runs offset %q: %w", q.Offset, tx.Error)
+		if err := database.DB.Select("row_num").First(&run).Error; err != nil && errors.Is(err, gorm.ErrRecordNotFound) {
```
jgiannuzzi, line 416: are you inverting the comparison on purpose?

dsuhinin: probably by mistake. a bit splitted it this `if` statement to do not confuse.

pkg/api/aim/runs.go
```diff
@@ -413,8 +413,10 @@ func SearchRuns(c *fiber.Ctx) error {
 			ID: q.Offset,
 		}
 		// TODO:DSuhinin -> do we need `namespace` restriction? it seems like yyyyess, but ....
-		if err := database.DB.Select("row_num").First(&run).Error; err != nil && errors.Is(err, gorm.ErrRecordNotFound) {
-			return fmt.Errorf("unable to find search runs offset %q: %w", q.Offset, err)
+		if err := database.DB.Select("row_num").First(&run).Error; err != nil {
+			if !errors.Is(err, gorm.ErrRecordNotFound) {
```
jgiannuzzi, line 417: hmm, why the extra `if` instead of just `&&` like we do everywhere else?

dsuhinin: it is more readable.

tests/integration/golang/aim/run/search_test.go
jgiannuzzi, line 1: could you please also add a test of the limit with+without the offset

tests/integration/golang/aim/run/search_test.go
```diff
@@ -237,6 +241,24 @@ func (s *SearchTestSuite) Test_Ok() {
...
+				Query:  `run.archived == True or run.archived == False`,
+				Offset: run2.ID,
+			},
+			runs: []*models.Run{
+				run3,
+				run4,
+			},
+		},
+		{
+			name: "SearchThirdPage",
```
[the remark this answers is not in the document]
dsuhinin: no. First page -> run 1,2, Second -> 3,4, Third -> nothing.

tests/integration/golang/aim/run/search_test.go
```diff
@@ -237,6 +241,24 @@ func (s *SearchTestSuite) Test_Ok() {
 		request request.SearchRunsRequest
 		runs    []*models.Run
 	}{
+		{
+			name: "SearchSecondPage",
+			request: request.SearchRunsRequest{
+				Query:  `run.archived == True or run.archived == False`,
```
jgiannuzzi, line 247: this could actually be `run.archived or not run.archived` — generally speaking this `run.archived == True` syntax is not very Pythonic

dsuhinin: this is want we have in other tests. so I prefer to keep it as is and changed it later everywhere.

pkg/api/aim/runs.go
jgiannuzzi: honestly, I don't think it is + it's inconsistent

dsuhinin: emmm, no. for example in `GetRunInfo`:
```
if err := tx.First(&r).Error; err != nil {
		if tx.Error == gorm.ErrRecordNotFound {
			return fiber.ErrNotFound
		}
		return fmt.Errorf("error retrieving run %q: %w", p.ID, tx.Error)
	}
```

tests/integration/golang/aim/run/search_test.go
jgiannuzzi: agreed — let's change this in another PR

pkg/api/aim/runs.go
dsuhinin: `GetApp`:
```
if err := database.DB.
		Where("NOT is_archived").
		First(&app).
		Error; err != nil {
		if err == gorm.ErrRecordNotFound {
			return fiber.ErrNotFound
		}
		return fiber.NewError(fiber.StatusInternalServerError, fmt.Sprintf("unable to find app %q: %s", p.ID, err))
	}
```

tests/integration/golang/aim/run/search_test.go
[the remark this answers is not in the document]
jgiannuzzi: ah I see that your row_nums are in weird order! that's what confused me — could you please make it a bit more logical and have the row_nums aligned with the run names?

tests/integration/golang/aim/run/search_test.go
dsuhinin: done.

tests/integration/golang/aim/run/search_test.go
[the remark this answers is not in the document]
dsuhinin: done. changed order and tests.

pkg/api/aim/runs.go
jgiannuzzi: that's exactly my point, all the places where we do that are places where we return an error when `err != nil` and then specialize the return based on the error type — this is _not_ what we're doing here! we're filtering the errors that are simple "ErrRecordNotFound".

`GetExperimentRuns` (same logic as `SearchRuns` for the limit/offset):

```go
	if q.Offset != "" {
		run := &database.Run{
			ID: q.Offset,
		}
		if err = database.DB.Select("row_num").First(&run).Error; err != nil && err != gorm.ErrRecordNotFound {
			return fmt.Errorf("unable to find search runs offset %q: %w", q.Offset, err)
		}

		tx.Where("row_num < ?", run.RowNum)
	}
```

tests/integration/golang/aim/run/search_test.go
```diff
@@ -237,6 +241,35 @@ func (s *SearchTestSuite) Test_Ok() {
...
 	}{
+		{
+			name: "SearchFirstPage",
+			request: request.SearchRunsRequest{
+				Query: `run.archived == True or run.archived == False`,
+				Limit: 2,
+			},
+			runs: []*models.Run{
+				run3,
+				run4,
```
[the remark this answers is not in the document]
jgiannuzzi: could you please change the order to the correct one here and in `SearchSecondPage`?
we can then fix the comparison logic in another PR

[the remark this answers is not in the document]
dsuhinin: ahhh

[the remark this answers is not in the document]
dsuhinin: you mean:
```
run3,
				run4,
```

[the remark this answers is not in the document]
dsuhinin: but should be:
```
run4,
run3
```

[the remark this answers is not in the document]
dsuhinin: ?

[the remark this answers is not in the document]
dsuhinin: if so, then yep, it is just how it works if you check.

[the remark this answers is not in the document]
jgiannuzzi: yep, it should `run4, run3` in the order that they are returned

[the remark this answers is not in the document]
dsuhinin: no, this is how test works.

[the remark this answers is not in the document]
jgiannuzzi: the whole decoder has been messed up since I added it — there is supposed to be an order to the objects being streamed, which we lost. and also the output of `fml decode` has become unreadable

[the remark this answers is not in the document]
jgiannuzzi: I just tested locally and the tests are happy whether I specify `[run4, run3]` or `[run3, run4]` — this means we are _not_ verifying the order of the runs being returned and this is wrong
again, this doesn't have to be addressed in this PR at all — I would just like us to have the right expected order here so when we add order validation we don't have to change it anymore

pkg/api/aim/runs.go
dsuhinin: ok. I disagree but reverted it back.

tests/integration/golang/aim/run/search_test.go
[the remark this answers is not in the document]
dsuhinin: yes, there is a problem in this test. how we fetch data from the decoded response.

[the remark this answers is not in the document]
dsuhinin: changed ordering now.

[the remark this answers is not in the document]
jgiannuzzi: I've create #514 and #515 to address this

jgiannuzzi (approved):

PR #531, opened 2023-10-30
Namespaces selector redesign

fabiovincenzi:


pkg/ui/chooser/embed/index.html
```diff
@@ -65,10 +78,16 @@
         </div>
 
         <ul id="namespaces-list">
-            {{ range .Namespaces }}
-            <li>
-                <a href="#" onclick="window.location = window.location.origin + {{if eq .Code "default"}}'/'{{else}}'/ns/{{.Code}}/'{{end}}">{{.Description}}</a>    
+            <li class="selected-namespace">
+                <i id="current-icon" class="Icon__container icon-long-arrow-right"></i>
+                <b id="current-namespace">{{.CurrentNamespace.Description}}</b>
             </li>
```
suprjinx, line 84: It seems like this could cause the current namespace to "move" after clicking (from middle of list to top) -- maybe this logic should be inside the loop instead, so it stays in the same place?

pkg/ui/chooser/embed/index.html
```diff
@@ -63,12 +76,20 @@
                 <p>This is the modern Aim UI, much faster than MLFlow.</p>
             </article>
         </div>
-
+        <p style="display: inline; margin-right: 5px;">Selected namespace:</p>
```
dave-gantenbein, line 79: Can you move this css into a selector please?

dave-gantenbein (approved):

PR #533, opened 2023-10-31
add ISSUE_TEMPLATE/feature_request.yml

naskio:
Adding an issue template for requesting features, including the following inputs:

- Related problem
- Suggested solution/implementation
- Alternatives (optional)
- Additional context (optional)


.github/ISSUE_TEMPLATE/feature_request.yml
```diff
@@ -0,0 +1,37 @@
...
+labels: [ "feature" ]
+
+body:
+  - type: textarea
+    id: related-problem
+    attributes:
+      label: Is your feature request related to a problem? Please describe.
+      description: A clear and concise description of what the problem is. Ex. I'm always frustrated when [...]
+    validations:
+      required: true
```
suprjinx, line 13: Feature request may not always relate to a "problem" -- "I wish this software had feature X" -- maybe shouldn't be required.

.github/ISSUE_TEMPLATE/feature_request.yml
```diff
@@ -0,0 +1,37 @@
...
+
+  - type: textarea
+    id: describe-solution
+    attributes:
```
suprjinx, line 17: Likewise, the new feature isn't necessary a "solution to a problem" but can just be something new. I'd think this could be more "describe the feature you'd like to see" -- and it should be the first field.

.github/ISSUE_TEMPLATE/feature_request.yml
jescalada: Made this optional and prioritized the feature description instead

PR #547, opened 2023-11-02
Fix health check for minio service in integration tests

dsuhinin:
For the latest `minio/minio:latest` image `health` check just started failing. It happened, because they started using another `base` image without `curl`. Workaround is here -> https://github.com/minio/minio/issues/18373#issuecomment-1790003599

### DIFF
--- a/tests/integration/docker-compose.yml
+++ b/tests/integration/docker-compose.yml
@@ -43,7 +43,7 @@ services:
     image: minio/minio:latest
     command: server /data
     healthcheck:
-      test: [ "CMD", "bash", "-c", ":> /dev/tcp/127.0.0.1/9000 || exit 1" ]
+      test: [ "CMD", "bash", "-c", ":> /dev/tcp/127.0.0.1/9000" ]
       interval: 5s
       timeout: 5s
     environment:
--- a/tests/integration/docker-compose.yml
+++ b/tests/integration/docker-compose.yml
@@ -43,7 +43,7 @@ services:
     image: minio/minio:latest
     command: server /data
     healthcheck:
-      test: ["CMD", "curl", "-f", "http://localhost:9000/minio/health/live"]
+      test: [ "CMD", "bash", "-c", ":> /dev/tcp/127.0.0.1/9000 || exit 1" ]
       interval: 5s
       timeout: 5s
     environment:
--- a/tests/integration/docker-compose.yml
+++ b/tests/integration/docker-compose.yml
@@ -43,7 +43,7 @@ services:
     image: minio/minio:latest
     command: server /data
     healthcheck:
-      test: [ "CMD", "bash", "-c", ":> /dev/tcp/127.0.0.1/9000" ]
+      test: ["CMD", "bash", "-c", ":> /dev/tcp/127.0.0.1/9000"]
       interval: 5s
       timeout: 5s
     environment:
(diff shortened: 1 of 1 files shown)


jgiannuzzi (commented):
Could you please add description to the PR?
Could you also please update the title to make it clear it’s for the integration tests? And also start with a capital and remove the dot so it’s consistent with our other PR titles.

dsuhinin:
> Could you please add a description to the PR?
> Could you also please update the title to make it clear it’s for the integration tests? And also start with a capital and remove the dot so it’s consistent with our other PR titles.

fixed.

tests/integration/docker-compose.yml
```diff
@@ -43,7 +43,7 @@ services:
     image: minio/minio:latest
     command: server /data
     healthcheck:
-      test: ["CMD", "curl", "-f", "http://localhost:9000/minio/health/live"]
+      test: [ "CMD", "bash", "-c", ":> /dev/tcp/127.0.0.1/9000 || exit 1" ]
```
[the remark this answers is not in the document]
dsuhinin: I think we don't need `|| exit 1` in that case. changed and tested -> works well.

tests/integration/docker-compose.yml
```diff
@@ -43,7 +43,7 @@ services:
     image: minio/minio:latest
     command: server /data
     healthcheck:
-      test: ["CMD", "curl", "-f", "http://localhost:9000/minio/health/live"]
+      test: [ "CMD", "bash", "-c", ":> /dev/tcp/127.0.0.1/9000" ]
```
suprjinx, line 46: that's a new one to me? would it maybe be better to lock the tag to the previous release which supported curl?

dsuhinin: Im not sure that we have to lock it. of course we can but why? they started using new `base` images which now have no `curl`. in any case we will need to jump to the fresh `image` so why we can't do that right now?

suprjinx: well if we used a release tag, this kind of thing wouldn't happen in the future, too -- and I think `curl` is more clear than using `/dev` filesystem (but that's me)

dsuhinin: that's true, it wouldn't happened but I think we want to use `latest` images(all our images have :latest prefix) so I think we have to follow this approach. not a big deal really. such cases happen too rarely.

jgiannuzzi, line 46: when formatting the YAML with the extension from our dev container, this is what it looks like:
```suggestion
      test: ["CMD", "bash", "-c", ":> /dev/tcp/127.0.0.1/9000"]
```

dsuhinin: hm, ok! Ill try to find something for Goland

jgiannuzzi: I quite liked the `curl` too, but this is the new recommended healthcheck for the official minio image, so let's go for it
with regards to locking the tag, I would like us to always use the latest available 3rd-party dependencies, and we immediately detected this issue thanks to the [daily scheduled CI run](https://github.com/G-Research/fasttrackml/actions/workflows/ci.yml?query=event%3Aschedule), which is the exact reason why we have it in the first place

jgiannuzzi: actually the same goes for the healthcheck you added a few days ago on line 60 — it should look like this:
```yaml
      test:
        ["CMD", "wget", "-q", "--spider", "http://localhost:4443/storage/v1/b"]
```

dsuhinin: yes. fixed already.

dsuhinin: that's a bit strange but it was auto formatting  in Goland but in just different style.

PR #550, opened 2023-11-04
Permit duplicate param logging

### LINKED ISSUE #523: Fix parameter logging of duplicates
Issue #43 was not completely implemented. For compatibility with MLFlow tracking server, we need:
1) 400 error when batch contains same key/different value parameter (currently returns 500)
2) 200 for same key/same value needs to work across batches, not just within batches.

suprjinx:
Previous iteration on this story only permitted duplicate parameterss within a batch. The changes here rely on the database to detect  permitted duplication.

Additionally, the error status code for parameters conflicts was changed from 500 ISE to 400 Bad Request.
Fixes #523


pkg/api/mlflow/dao/repositories/param.go
```diff
@@ -32,48 +43,65 @@ func NewParamRepository(db *gorm.DB) *ParamRepository {
...
-	var keys []string
-	paramMap := map[string]models.Param{}
-	for _, param := range params {
-		key := fmt.Sprintf("%v-%v-%v", param.RunID, param.Key, param.Value)
-		keys = append(keys, key)
-		paramMap[key] = param
+// findConflictingParams checks if there are conflicting values for the input params. If a key does not
+// yet exist in the db, or if the same key and value already exist for the run, it is not a conflict.
+// If the key already exists for the run but with a different value, it is a conflict. Conflicting keys are returned.
+func (r ParamRepository) findConflictingParams(ctx context.Context, params []models.Param) ([]string, error) {
```
jgiannuzzi, line 72: this whole function could be changed by a single SQL query in this spirit:
```sql
WITH new(key, value) AS (VALUES (?, ?), (?, ?))
SELECT current.key, current.value, new.value
FROM params AS current
INNER JOIN new ON run_uuid = ? AND new.key = current.key
WHERE new.value != current.value;
```

`VALUES` can be built as a list of  `n` `"(?, ?)"` joined by `", "`, where `n` is the amount of params we tried to insert.

The SQL parameters that need to be passed are the list keys and values we tried to insert, and then the run ID.

The DB engine will then return to you all you need to log the error nicely to the user.

suprjinx: I can change to this, but I'd suggest a simple query like we have currently plus Go-layer manipulation is easier to grok

pkg/api/mlflow/dao/repositories/param.go
```diff
@@ -32,48 +43,65 @@ func NewParamRepository(db *gorm.DB) *ParamRepository {
...
 // CreateBatch creates []models.Param entities in batch.
 func (r ParamRepository) CreateBatch(ctx context.Context, batchSize int, params []models.Param) error {
-	// try to create params in batch; error condition requires special handling
-	// to allow certain duplicates
-	if err := r.db.CreateInBatches(params, batchSize).Error; err != nil {
-		// remove duplicate rows and try again
-		dedupedParams, errRemovingMatches := r.removeExactMatches(ctx, params)
-		if errRemovingMatches != nil {
-			return eris.Wrap(errRemovingMatches, "error removing duplicates in batch")
+	tx := r.db.Clauses(clause.OnConflict{
```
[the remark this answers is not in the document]
suprjinx: done

pkg/api/mlflow/dao/repositories/param.go
```diff
@@ -32,48 +43,65 @@ func NewParamRepository(db *gorm.DB) *ParamRepository {
...
-		paramMap[key] = param
+// findConflictingParams checks if there are conflicting values for the input params. If a key does not
+// yet exist in the db, or if the same key and value already exist for the run, it is not a conflict.
+// If the key already exists for the run but with a different value, it is a conflict. Conflicting keys are returned.
+func (r ParamRepository) findConflictingParams(ctx context.Context, params []models.Param) ([]string, error) {
+	dbParams := []models.Param{}
+	paramKeysInError := []string{}
+	if err := r.db.WithContext(ctx).
+		Model(&models.Param{}).
+		Where("run_uuid = ?", params[0].RunID).
```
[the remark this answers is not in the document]
suprjinx: this assumption is removed in rewriting the query to find duplicate

pkg/api/mlflow/dao/repositories/param.go
```diff
@@ -32,48 +43,65 @@ func NewParamRepository(db *gorm.DB) *ParamRepository {
...
-	for _, foundKey := range foundKeys {
-		delete(paramMap, foundKey)
+// collectKeys collects the keys from the params.
+func (r ParamRepository) collectKeys(params []models.Param) []string {
+	keys := make([]string, len(params))
+	for i, param := range params {
+		keys[i] = param.Key
 	}
+	return keys
+}
```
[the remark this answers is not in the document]
suprjinx: removed receiver

pkg/api/mlflow/dao/repositories/param.go
```diff
@@ -43,65 +44,60 @@ func NewParamRepository(db *gorm.DB) *ParamRepository {
...
+	params []models.Param,
+) ([]map[string]string, error) {
+	var paramsInError []map[string]string
 	if err := r.db.WithContext(ctx).
-		Model(&models.Param{}).
-		Where("run_uuid = ?", params[0].RunID).
-		Where("key IN ?", r.collectKeys(params)).
-		Find(&dbParams).Error; err != nil {
+		Raw(fmt.Sprintf(`
+		    WITH new(key, value, run_uuid) AS (VALUES %s)
```
jgiannuzzi, line 84: this is vulnerable to SQL injection
as I explained in https://github.com/G-Research/fasttrackml/pull/550#discussion_r1382611435, please build the values as a list of `n` `"(?, ?)"` strings joined by `", "`, and then pass the actual values as parameters to the query

pkg/api/mlflow/dao/repositories/param.go
```diff
@@ -43,65 +44,60 @@ func NewParamRepository(db *gorm.DB) *ParamRepository {
...
-func (r ParamRepository) findConflictingParams(ctx context.Context, params []models.Param) ([]string, error) {
-	dbParams := []models.Param{}
-	paramKeysInError := []string{}
+// If the key already exists for the run but with a different value, it is a conflict. Conflicting keys
+// are returned.
+func (r ParamRepository) findConflictingParams(
+	ctx context.Context,
+	params []models.Param,
+) ([]map[string]string, error) {
+	var paramsInError []map[string]string
```
jgiannuzzi, line 81: create an actual struct type instead

jgiannuzzi: e.g.
```golang
type errorParam struct {
	RunID string
	Key string
	OldValue string
	NewValue string
}
```

pkg/api/mlflow/dao/repositories/param.go
```diff
@@ -43,65 +44,60 @@ func NewParamRepository(db *gorm.DB) *ParamRepository {
...
 	if err := r.db.WithContext(ctx).
-		Model(&models.Param{}).
-		Where("run_uuid = ?", params[0].RunID).
-		Where("key IN ?", r.collectKeys(params)).
-		Find(&dbParams).Error; err != nil {
+		Raw(fmt.Sprintf(`
+		    WITH new(key, value, run_uuid) AS (VALUES %s)
+		    SELECT current.key as key, current.value as old_value, new.value as new_value
```
jgiannuzzi, line 85: you also need to select the run ID if we want to be run agnostic

jgiannuzzi: alternatively, change the ParamRepositoryProvider interface as discussed before

pkg/api/mlflow/dao/repositories/param.go
```diff
@@ -43,65 +44,60 @@ func NewParamRepository(db *gorm.DB) *ParamRepository {
...
+		    FROM params AS current
+		    INNER JOIN new ON new.run_uuid = current.run_uuid AND new.key = current.key
```
jgiannuzzi, line 87: if we go for being run-agnostic, you can write this as `INNER JOIN new USING(run_uuid, key)`

pkg/api/mlflow/dao/repositories/param.go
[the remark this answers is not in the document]
jgiannuzzi: could you please add a test for this?

pkg/api/mlflow/dao/repositories/param.go
suprjinx: done

pkg/api/mlflow/dao/repositories/param.go
suprjinx: including run id for each param

pkg/api/mlflow/dao/repositories/param.go
suprjinx: will do

pkg/api/mlflow/dao/repositories/param.go
suprjinx: done

pkg/api/mlflow/dao/repositories/param.go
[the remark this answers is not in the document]
suprjinx: added test in integration tests (mlflow/run/log_batch_test.go)

pkg/api/mlflow/dao/repositories/param.go
```diff
@@ -72,32 +85,41 @@ func (r ParamRepository) CreateBatch(ctx context.Context, batchSize int, params
...
-		    INNER JOIN new ON new.run_uuid = current.run_uuid AND new.key = current.key
-		    WHERE new.value != current.value
-                `, prepareSqlValues(params))).
-		Scan(&paramsInError).Error; err != nil {
-		return nil, eris.New("error fetching params from db")
+// If the key already exists for the run but with a different value, it is a conflict. Conflicts are returned.
+func findConflictingParams(tx *gorm.DB, params []models.Param) ([]paramConflict, error) {
+	var conflicts []paramConflict
+	if err := tx.Raw(
+		`WITH new(key, value, run_uuid) AS (VALUES `+makeSqlPlaceholders(params)+`)
```
jgiannuzzi, line 92: could you please use `fmt.Sprintf` here to improve readability?

tests/integration/golang/mlflow/run/log_batch_test.go
jgiannuzzi, line 1: we're missing a test of duplicate key with different value in a _separate batch_

pkg/api/mlflow/dao/repositories/param.go
suprjinx: i started with that but it seemed like this was clearer -- since we use a values array for sql parameter replacement, it seemed weird to have first one value replacement for the template string (sql), then another

suprjinx: looks better w/ separate statement to make the sql template, please take a look

tests/integration/golang/mlflow/run/log_batch_test.go
suprjinx: added a setup to to put "previous batch" param in db

pkg/api/mlflow/dao/repositories/param.go
```diff
@@ -32,48 +57,68 @@ func NewParamRepository(db *gorm.DB) *ParamRepository {
...
 // CreateBatch creates []models.Param entities in batch.
 func (r ParamRepository) CreateBatch(ctx context.Context, batchSize int, params []models.Param) error {
-	// try to create params in batch; error condition requires special handling
-	// to allow certain duplicates
-	if err := r.db.CreateInBatches(params, batchSize).Error; err != nil {
-		// remove duplicate rows and try again
-		dedupedParams, errRemovingMatches := r.removeExactMatches(ctx, params)
-		if errRemovingMatches != nil {
-			return eris.Wrap(errRemovingMatches, "error removing duplicates in batch")
+	if err := r.db.Transaction(func(tx *gorm.DB) error {
```
dave-gantenbein, line 60: Is this just tested via the integration tests? Is it not possible to unit test this method?

pkg/api/mlflow/service/run/service.go
```diff
@@ -520,7 +520,11 @@ func (s Service) LogParam(
 
 	param := convertors.ConvertLogParamRequestToDBModel(run.ID, req)
 	if err := s.paramRepository.CreateBatch(ctx, 1, []models.Param{*param}); err != nil {
-		return api.NewInternalError("unable to insert params for run '%s': %s", run.ID, err)
+		if _, ok := err.(repositories.ParamConflictError); ok {
```
dave-gantenbein, line 523: It seems like it would be nice to have unit tests demonstrating the different error cases.

pkg/api/mlflow/dao/repositories/param.go
suprjinx: I think not very useful to unit test since we are depending on database here

pkg/api/mlflow/service/run/service.go
suprjinx: I think we do cover this in service_test.go

suprjinx: i can add same test for LogBatch..

dsuhinin, line 523: probably we can just use `errors.Is` in that case?

dsuhinin (changes requested):

pkg/api/mlflow/dao/repositories/param.go
```diff
@@ -32,48 +57,68 @@ func NewParamRepository(db *gorm.DB) *ParamRepository {
...
+		if err := tx.WithContext(ctx).Clauses(clause.OnConflict{
+			Columns:   []clause.Column{{Name: "run_uuid"}, {Name: "key"}},
+			DoNothing: true,
+		}).CreateInBatches(params, batchSize).Error; err != nil {
+			return eris.Wrap(tx.Error, "error creating params in batch")
```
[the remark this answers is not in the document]
suprjinx: yep!

pkg/api/mlflow/dao/repositories/param.go
```diff
@@ -32,48 +57,68 @@ func NewParamRepository(db *gorm.DB) *ParamRepository {
...
-	for _, foundKey := range foundKeys {
-		delete(paramMap, foundKey)
+// makeSqlPlaceholders collects a string of (?,?,?), (?,?,?), etc, to the length of the params.
+func makeSqlPlaceholders(params []models.Param) string {
+	valuesArray := make([]string, len(params))
+	for i := range params {
+		valuesArray[i] = "(?,?,?)"
 	}
+	return strings.Join(valuesArray, ",")
+}
```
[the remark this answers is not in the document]
suprjinx: I was trying to balance the "values" function which is highly type-specific, but this could just be `makePlaceholders(numberForEach int, total int) string` or the like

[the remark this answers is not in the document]
suprjinx: or maybe return placeholders and values as a tuple from one function

pkg/api/mlflow/service/run/service.go
```diff
@@ -607,7 +611,11 @@ func (s Service) LogBatch(
 		return api.NewInvalidParameterValueError(err.Error())
 	}
 	if err := s.paramRepository.CreateBatch(ctx, 100, params); err != nil {
-		return api.NewInternalError("unable to insert params for run '%s': %s", run.ID, err)
+		if _, ok := err.(repositories.ParamConflictError); ok {
```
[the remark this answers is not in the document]
suprjinx: i tried and did not seem to work, but I'll look at the `errors` source

pkg/api/mlflow/dao/repositories/param.go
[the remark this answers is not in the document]
jgiannuzzi: I like this second idea!
a generic function that returns both the placeholders string and the values slice sounds like the way to go — once done, please change the other places in our code that use the same pattern (in a separate PR of course)

pkg/api/mlflow/dao/repositories/param.go
```diff
@@ -3,13 +3,38 @@ package repositories
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	"github.com/G-Research/fasttrackml/pkg/api/mlflow/dao/models"
 
 	"github.com/rotisserie/eris"
 	"gorm.io/gorm"
+	"gorm.io/gorm/clause"
```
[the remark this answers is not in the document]
suprjinx: yep

pkg/api/mlflow/service/run/service.go
```diff
@@ -607,7 +611,11 @@ func (s Service) LogBatch(
...
+			return api.NewInvalidParameterValueError("unable to insert params for run '%s': %s", run.ID, err)
+		} else {
```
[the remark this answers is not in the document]
suprjinx: ok

pkg/api/mlflow/service/run/service.go
```diff
@@ -520,7 +520,11 @@ func (s Service) LogParam(
 
 	param := convertors.ConvertLogParamRequestToDBModel(run.ID, req)
 	if err := s.paramRepository.CreateBatch(ctx, 1, []models.Param{*param}); err != nil {
-		return api.NewInternalError("unable to insert params for run '%s': %s", run.ID, err)
+		if _, ok := err.(repositories.ParamConflictError); ok {
+			return api.NewInvalidParameterValueError("unable to insert params for run '%s': %s", run.ID, err)
+		} else {
```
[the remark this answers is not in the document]
suprjinx: ok

tests/integration/golang/mlflow/run/log_parameter_test.go
```diff
@@ -111,9 +150,63 @@ func (s *LogParamTestSuite) Test_Error() {
...
+			http.MethodPost,
+		).WithRequest(
+			req,
+		).WithResponse(
+			&resp,
+		).DoRequest(
+			fmt.Sprintf("%s%s", mlflow.RunsRoutePrefix, mlflow.RunsLogParameterRoute),
+		),
+	)
+	//nolint:lll
```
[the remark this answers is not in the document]
suprjinx: done

pkg/api/mlflow/dao/repositories/param.go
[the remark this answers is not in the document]
suprjinx: i don't think generic works very well here since we need specific fields from the struct, peculiar to a sql statement

pkg/api/mlflow/service/run/service.go
suprjinx: errors.As is working, changed to that

pkg/api/mlflow/service/run/service.go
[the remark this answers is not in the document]
suprjinx: errors.As works

pkg/api/mlflow/dao/repositories/helpers.go
```diff
@@ -0,0 +1,38 @@
...
+	valuesArray := make([]interface{}, len(params)*3)
+	index := 0
+	for _, param := range params {
+		valuesArray[index] = param.Key
+		valuesArray[index+1] = param.Value
+		valuesArray[index+2] = param.RunID
+		index = index + 3
+	}
+	return placeholders, valuesArray
+}
```
dsuhinin, line 38: we can move of course with this solution but my proposal was to have a function based on generics, where you can get the number of properties through the reflection and build `placeholders` for any slice. @jgiannuzzi what do you think?

pkg/api/mlflow/service/run/service.go
```diff
@@ -520,6 +521,9 @@ func (s Service) LogParam(
 
 	param := convertors.ConvertLogParamRequestToDBModel(run.ID, req)
 	if err := s.paramRepository.CreateBatch(ctx, 1, []models.Param{*param}); err != nil {
+		if errors.As(err, &repositories.ParamConflictError{}) {
```
dsuhinin, line 524: nice!

dsuhinin (dismissed):
Looks good for me!

pkg/api/mlflow/service/run/service.go
```diff
@@ -520,11 +521,10 @@ func (s Service) LogParam(
 
 	param := convertors.ConvertLogParamRequestToDBModel(run.ID, req)
 	if err := s.paramRepository.CreateBatch(ctx, 1, []models.Param{*param}); err != nil {
-		if _, ok := err.(repositories.ParamConflictError); ok {
+		if errors.As(err, &repositories.ParamConflictError{}) {
```
jgiannuzzi, line 524: should this not be `errors.Is` instead so we don't use the unwrapped error below and potentially lose parts of the error stack?

pkg/api/mlflow/service/run/service.go
```diff
@@ -611,11 +611,10 @@ func (s Service) LogBatch(
 		return api.NewInvalidParameterValueError(err.Error())
 	}
 	if err := s.paramRepository.CreateBatch(ctx, 100, params); err != nil {
-		if _, ok := err.(repositories.ParamConflictError); ok {
+		if errors.As(err, &repositories.ParamConflictError{}) {
```
jgiannuzzi, line 614: same remark as above

pkg/api/mlflow/dao/repositories/helpers.go
jgiannuzzi: agreed, I would have liked to see something more generic — however this is good enough for this PR

pkg/api/mlflow/dao/repositories/helpers.go
```diff
@@ -0,0 +1,38 @@
+package repositories
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/G-Research/fasttrackml/pkg/api/mlflow/dao/models"
+)
+
+// makeSqlPlaceholders collects a string of (?,?,?), (?,?,?), etc
```
jgiannuzzi, line 10: please provide a better description

pkg/api/mlflow/dao/repositories/helpers.go
```diff
@@ -0,0 +1,38 @@
...
+	placeholderArray := make([]string, numberInEachSet)
+	for i := 0; i < numberInEachSet; i++ {
+		placeholderArray[i] = "?"
+	}
+	setsArray := make([]string, numberOfSets)
+	for i := 0; i < numberOfSets; i++ {
+		setsArray[i] = fmt.Sprintf("(%s)", strings.Join(placeholderArray, ","))
+	}
+	return strings.Join(setsArray, ",")
+}
```
jgiannuzzi, line 21: this could be improved in terms of memory usage by using string builders (especially for the second iteration)

you may ask "is that not premature optimization?" and the answer to that is right up in the project name — we do care _a lot_ about optimizations in this project, and this is our main differentiator compared to the project whose API we implement

pkg/api/mlflow/dao/repositories/param.go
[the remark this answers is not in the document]
jgiannuzzi: the ordered list of fields would be part of the parameters to this generic function — anyway as I commented in another place, this is good enough for this PR

pkg/api/mlflow/dao/repositories/param.go
jgiannuzzi: yes, it looks good now!

pkg/api/mlflow/dao/repositories/helpers.go
suprjinx: Are not preallocated slices of the exact length efficient? The resulting arrays make it easier to use `strings.Join` to get the commas right.

jgiannuzzi: It doubles the allocated size: one for the slice of strings, and one for the joined string. It also puts more load on the GC by allocating many small strings (a slice of strings is just an array of pointers, and all of these pointer references need to be tracked by the GC)

jgiannuzzi: If you want to keep this implementation for readability, then you'd at least want to have the call to `fmt.Sprintf` done only once so that it's the same string reference that gets added n time to a slice

pkg/api/mlflow/service/run/service.go
suprjinx: I think we have to change from Error interface to error instance for `Is` to work.

suprjinx: I'd suggest we keep the error interface and `As` usage so that each ParamConflictError can carry the particular message.

pkg/api/mlflow/dao/repositories/helpers.go
suprjinx: added some description

pkg/api/mlflow/service/run/service.go
jgiannuzzi: I looked into this further, and we could make `Is` work by implementing `func (e ParamConflictError) Is(target error) bool`. However this does not seem necessary (at least at this stage), as indeed `As` does what we want here. I originally thought that `err` would get unwrapped and changed by the call to `As`, but this is not the case. It's the second parameter that gets assigned, but you don't use it so all is good!

pkg/api/mlflow/dao/repositories/helpers.go
```diff
@@ -7,15 +7,17 @@ import (
...
 func makeSqlPlaceholders(numberInEachSet, numberOfSets int) string {
 	placeholderArray := make([]string, numberInEachSet)
 	for i := 0; i < numberInEachSet; i++ {
 		placeholderArray[i] = "?"
 	}
+	placeholderTemplate := strings.Join(placeholderArray, ",")
 	setsArray := make([]string, numberOfSets)
 	for i := 0; i < numberOfSets; i++ {
-		setsArray[i] = fmt.Sprintf("(%s)", strings.Join(placeholderArray, ","))
+		setsArray[i] = fmt.Sprintf("(%s)", placeholderTemplate)
```
jgiannuzzi, line 20: this call to `fmt.Sprintf` should be placed outside the loop

pkg/api/mlflow/dao/repositories/helpers.go
jgiannuzzi: so I ran a quick benchmark of this implementation with the following code:
```go
func Benchmark_makeSqlPlaceholders(b *testing.B) {
	for i := 0; i < b.N; i++ {
		makeSqlPlaceholders(1, 100)
		makeSqlPlaceholders(2, 100)
		makeSqlPlaceholders(3, 100)
	}
}
```

I get the following results:
```
Benchmark_makeSqlPlaceholders-6            51420             22895 ns/op           13892 B/op        611 allocs/op
```

When I switch to using `strings.Builder` instead, I get the following results:
```
Benchmark_makeSqlPlaceholders-6           540442              2178 ns/op            3088 B/op          6 allocs/op
```
In this particular benchmark, it is 10x faster, uses 4x less memory and does 100x less allocations.

Here is the implementation:
```go
func makeSqlPlaceholders(numberInEachSet, numberOfSets int) string {
	setLength := 2*numberInEachSet + 1
	var set strings.Builder
	set.Grow(setLength)
	set.WriteString("(")
	for i := 0; i < numberInEachSet-1; i++ {
		set.WriteString("?")
		set.WriteString(",")
	}
	set.WriteString("?)")
	setString := set.String()

	setsLength := 2*numberOfSets*setLength - 1
	var sets strings.Builder
	sets.Grow(setsLength)
	for i := 0; i < numberOfSets-1; i++ {
		sets.WriteString(setString)
		sets.WriteString(",")
	}
	sets.WriteString(set.String())
	return sets.String()
}
```

I then tried another implementation that uses `strings.Repeat` and got these results:
```
Benchmark_makeSqlPlaceholders-6           927678              1131 ns/op            3985 B/op         15 allocs/op
```
It's another 2x faster and uses ~30% more memory and 2x more allocations, but it's much simpler:
```go
func makeSqlPlaceholders(numberInEachSet, numberOfSets int) string {
	set := fmt.Sprintf("(%s)", strings.Repeat("?,", numberInEachSet-1)+"?")
	return strings.Repeat(set+",", numberOfSets-1) + set
}
```

Again, I know this is not on the critical path, but we should always strive to have great performance anywhere in this project as this is our crucial differentiator!

suprjinx: okay, thanks for digging in -- I'll switch to the strings.Repeat implementation

pkg/api/mlflow/dao/repositories/helpers.go
suprjinx: ok

jgiannuzzi (dismissed):

tests/integration/golang/mlflow/run/log_batch_test.go
```diff
@@ -123,6 +123,14 @@ func (s *LogBatchTestSuite) TestParams_Ok() {
 	})
 	require.Nil(s.T(), err)
 
+	// create preexisting param (other batch) for conflict testing
+	_, err = s.ParamFixtures.CreateParam(context.Background(), &models.Param{
+		RunID: run.ID,
+		Key:   "key1",
+		Value: "value1",
+	})
+	assert.Nil(s.T(), err)
```
jgiannuzzi, line 132: ah shoot, we have merged #584 in the meantime, so you should use `require.Nil` instead of `assert.Nil` to be consistent!
sorry for the extra work 😞

suprjinx: done

tests/integration/golang/mlflow/run/log_parameter_test.go
```diff
@@ -107,9 +146,63 @@ func (s *LogParamTestSuite) Test_Error() {
...
+	assert.Empty(s.T(), resp)
+
+	// error conditions
+
+	// missing run_id
+	req = request.LogParamRequest{
+		Key:   "key1",
+		Value: "value1",
+	}
+	assert.Nil(
```
jgiannuzzi, line 158: another one

tests/integration/golang/mlflow/run/log_parameter_test.go
```diff
@@ -107,9 +146,63 @@ func (s *LogParamTestSuite) Test_Error() {
...
 		resp.Error(),
 	)
+
+	// conflicting param
+	req = request.LogParamRequest{
+		RunID: run.ID,
+		Key:   "key1",
+		Value: "value2",
+	}
+	assert.Nil(
```
jgiannuzzi, line 182: and the last one

tests/integration/golang/mlflow/run/log_parameter_test.go
```diff
@@ -79,18 +79,57 @@ func (s *LogParamTestSuite) Test_Ok() {
...
 	)
 	assert.Empty(s.T(), resp)
+
+	// log duplicate, which is OK
+	req = request.LogParamRequest{
+		RunID: run.ID,
+		Key:   "key1",
+		Value: "value1",
+	}
+	assert.Nil(
```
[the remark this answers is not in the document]
suprjinx: done

jgiannuzzi (approved):

Issue #328, opened 2023-09-18
Improve Go caching in CI


jgiannuzzi:
We currently rely on the default for caching provided by `actions/setup-go` and it's pretty good. However we need to make the cache keys more specific as we can currently clash between e.g. building linux/arm64 and linux/amd64, so one of them never gets properly cached. We should also cache tools like mockery that get installed in some of our workflows

PR #555, opened 2023-11-05
Improve Aim Python integration tests

jgiannuzzi:
This PR is the fifth in a series to address #328.

It makes the Aim Python integration tests setup lighter by ditching the unused Aim fixtures. It also makes them more accurate by fixing the server rather than the tests. Finally, it enables more tests that were previously skipped.

### DIFF
--- a/tests/integration/python/aim/custom.patch
+++ b/tests/integration/python/aim/custom.patch
@@ -1,268 +1,201 @@
 diff --git a/tests/api/test_dashboards_api.py b/tests/api/test_dashboards_api.py
-index 94bc29c..3a8237d 100644
+index 94bc29c..85000fc 100644
 --- a/tests/api/test_dashboards_api.py
 +++ b/tests/api/test_dashboards_api.py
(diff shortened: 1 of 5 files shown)


tests/integration/python/aim/custom.patch
```diff
@@ -1,268 +1,201 @@
 diff --git a/tests/api/test_dashboards_api.py b/tests/api/test_dashboards_api.py
-index 94bc29c..3a8237d 100644
+index 94bc29c..85000fc 100644
```
suprjinx, line 2: This "diff of a diff" isn't very easy to review -- should we think about an approach like the UI modules for these tests?

suprjinx (approved):

tests/integration/python/aim/custom.patch
jgiannuzzi: Agreed that it makes it hard to review. I do wonder if the module approach may be overkill for these tests as they shouldn't change much?

PR #578, opened 2023-11-13
Display and Update experiment description in aim UI

### LINKED ISSUE #522: Add support for displaying and updating experiment description in Aim UI
The experiment description is currently saved as the experiment tag `mlflow.note.content` in the MLFlow UI.
This should be exposed in all the Aim endpoints that return an experiment, and the `UpdateExperiment` endpoint needs to be implemented.

fabiovincenzi:
Fixes #522.
New `UpdateExperiment` endpoint in aim, which allows to edit experiment name, description and archived.
`GetExperiment` and `GetExperiments` now return the experiment description.
Update `GetExperiment` and `GetExperiments` integration tests to actually test experiment desciption.


pkg/api/aim/experiments.go
```diff
@@ -327,3 +335,77 @@ func DeleteExperiment(c *fiber.Ctx) error {
 		"status": "OK",
 	})
 }
+
+func UpdateExperiment(c *fiber.Ctx) error {
```
dave-gantenbein, line 339: I know we're going to refactor this soon, but rather than put this net new method here where it soon will be moved (and without unit tests), can we instead create the first repository object for aim and put it there (with tests)? @jgiannuzzi  - what do you think?

pkg/api/aim/experiments.go
```diff
@@ -327,3 +335,77 @@ func DeleteExperiment(c *fiber.Ctx) error {
...
+	if err = c.ParamsParser(&params); err != nil {
+		return fiber.NewError(fiber.StatusUnprocessableEntity, err.Error())
+	}
+
+	var updateRequest request.UpdateExperimentRequest
+	if err = c.BodyParser(&updateRequest); err != nil {
+		return fiber.NewError(fiber.StatusUnprocessableEntity, err.Error())
+	}
+
+	id, err := strconv.ParseInt(params.ID, 10, 32)
```
suprjinx, line 358: perhaps move up to line 352 so it's adjacent to ID parsing

pkg/api/aim/experiments.go
```diff
@@ -327,3 +335,77 @@ func DeleteExperiment(c *fiber.Ctx) error {
...
+			}
+			return nil
+		}); err != nil {
+			return fiber.NewError(fiber.StatusInternalServerError,
+				fmt.Sprintf("unable to update experiment %q: %s", params.ID, err))
+		}
+	}
+	if updateRequest.Description != nil {
+		description := models.ExperimentTag{
+			Key:          "mlflow.note.content",
```
suprjinx, line 398: i'm wondering if this should be a first-class field on Experiment table instead of using a magic tag. We could then translate experiment.description -> experiment_tag on the mlflow side where needed. That way the mlflow implementation (magic tag) would be isolated on the mlflow side of the house.

tests/integration/golang/aim/experiment/get_experiment_test.go
```diff
@@ -43,7 +44,7 @@ func (s *GetExperimentTestSuite) Test_Ok() {
 		Name: "Test Experiment",
 		Tags: []models.ExperimentTag{
 			{
-				Key:   "key1",
+				Key:   "mlflow.note.content",
```
suprjinx, line 47: If we are keeping magic tag, can we make this key a constant?

tests/integration/golang/helpers/experiment.go
```diff
@@ -0,0 +1,15 @@
+package helpers
+
+import (
+	"github.com/G-Research/fasttrackml/pkg/api/mlflow/dao/models"
+)
+
+// GetDescriptionFromexperiment returns the description of a given experiment.
+func GetDescriptionFromExperiment(experiment models.Experiment) string {
+	for _, tag := range experiment.Tags {
+		if tag.Key == "mlflow.note.content" {
```
suprjinx, line 10: use constant for key if keeping this implementation

pkg/api/aim/response/experiment.go
```diff
@@ -2,7 +2,7 @@ package response
 
 // GetExperiment represents the response json fot the GetExperimnt endpoint.
 type GetExperiment struct {
-	ID           int32   `json:"id"`
+	ID           string  `json:"id"`
```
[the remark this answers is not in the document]
fabiovincenzi: They actually are in aim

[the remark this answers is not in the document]
dsuhinin: if it fixed something in AIM UI, then we unfortunately need it in `string` format I think.

pkg/api/aim/experiments.go
dsuhinin: I think there is no big difference right now. My vote to have it here and keep everything similar(even if it looks bad) and then in scope of separate tasks make it better and well structured.

dsuhinin: so yep, looks a bit ugly and overloaded but I hope we an normalise it very soon, because otherwise in scope of this task we will have even more changes we need which we always try to avoid.

pkg/api/aim/response/experiment.go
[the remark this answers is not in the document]
fabiovincenzi: we were already returning a `string` in `GetExperiments` endpoint for the same reason I believe https://github.com/G-Research/fasttrackml/blob/d07b06e866b644735ac70e1c543c63ca8b8b0ab8/pkg/api/aim/experiments.go#L56

tests/integration/golang/helpers/experiment.go
```diff
@@ -0,0 +1,15 @@
...
+			return tag.Value
+		}
+	}
+	return ""
```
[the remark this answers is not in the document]
fabiovincenzi: If there is no description, the experiment tag for the description doesn't exist, but if it doesn't an empty string is returned

https://github.com/fabiovincenzi/fasttrack/blob/8c223e95ae7e1def46b0aa49d04912e0d711aa60/pkg/api/aim/experiments.go#L110

pkg/api/aim/experiments.go
jgiannuzzi: this is a great idea for a future improvement!
for now I would keep using the magic tag so we can get this gap bridged quickly

pkg/api/aim/experiments.go
dave-gantenbein: okay

suprjinx (dismissed):

pkg/api/aim/experiments.go
```diff
@@ -36,10 +38,13 @@ func GetExperiments(c *fiber.Ctx) error {
 			"experiments.lifecycle_stage",
 			"experiments.creation_time",
 			"COUNT(runs.run_uuid) AS run_count",
+			"COALESCE(MAX(experiment_tags.value), '') AS description",
```
suprjinx, line 41: `MAX` function seems a little weird here, with a string value

pkg/api/aim/experiments.go
```diff
@@ -327,3 +335,77 @@ func DeleteExperiment(c *fiber.Ctx) error {
...
+				fmt.Sprintf("unable to update experiment %q: %s", params.ID, err))
+		}
+	}
+	if updateRequest.Description != nil {
+		description := models.ExperimentTag{
+			Key:          common.DescriptionTagKey,
+			Value:        *updateRequest.Description,
+			ExperimentID: *experiment.ID,
+		}
+		if err := tagRepository.CreateExperimentTag(c.Context(), &description); err != nil {
```
suprjinx, line 402: for update, wouldn't we want to update the description tag if it exists instead of creating a new one?

pkg/api/aim/experiments.go
```diff
@@ -101,22 +107,24 @@ func GetExperiment(c *fiber.Ctx) error {
 			"experiments.lifecycle_stage",
 			"experiments.creation_time",
 			"COUNT(runs.run_uuid) AS run_count",
+			"COALESCE(MAX(experiment_tags.value), '') AS description",
```
suprjinx, line 110: another `MAX` function

pkg/api/aim/experiments.go
fabiovincenzi: Yes it is returning the string that alphabetizes to the top, but since there's only one tag with that key it is always the right description.

pkg/api/aim/experiments.go
fabiovincenzi: this should already update the existing tag if it already exists:
https://github.com/G-Research/fasttrackml/blob/202fa759a4704606bb21f7f6e7558e7920f583a1/pkg/api/mlflow/dao/repositories/tag.go#L43

suprjinx (approved):

Issue #561, opened 2023-11-06
Integration tests: add a new way to run them against a new separate server instance per test


jgiannuzzi:
We currently run our integration tests against a single server instance, and we make database changes directly. This was fine whilst the server was completely stateless, but with #291 this assumption won't be true anymore.
~We should introduce a new way to run the integration tests, where each HTTP client instance spins up its own server instance on a random port and sets up its base URI to it. We should keep the ability to run the integration tests towards a specific URI for debugging purposes, but it shouldn't be the default anymore (configurable via env var).~
We should introduce a new way to run the integration tests, where each HTTP client instance uses an instance of our server (`*fiber.App`) and uses `httptest.NewRequest` + `server.Test()`.

- [x] #584
- [x] #585
- [x] #624
- [x] #627
- [x] #632
- [x] #633
- [x] #634
- [x] #642
- [x] #643
- [x] #644
- [x] #645
- [x] #672
- [x] #673
- [x] #674
- [x] #675
- [x] #676
- [x] #687
- [x] #688
- [x] #689
- [x] #690
- [x] #691
- [x] #695
- [x] #696

PR #584, opened 2023-11-13
Use `require` instead of `assert` for checking errors in tests

jgiannuzzi:
We use `assert.Nil` in our tests to check for errors. This means that if there _was_ an error, the tests will continue and then probably panic further. This PR changes all the places where we used `assert.Nil` to validate that a call was successful to use `require.Nil` instead, which will immediately fail the test and return a clean failure instead of a panic.

There are also places where `assert.Nil` is used to validate that something other than an error is `nil`. Those have not been changed.


dsuhinin (approved):
Looks good!

jgiannuzzi:
Part of the work for #561.

PR #597, opened 2023-11-15
Ensure our various copies of models are in sync

jgiannuzzi:
Our models are currently duplicated in multiple places in the code. This PR ensures that they are all in sync.


dave-gantenbein (approved):
LGTM and OMG how nice to see CI is already green!

PR #599, opened 2023-11-15
add json context to metrics and latest_metrics

### LINKED ISSUE #580: Metric contexts: adapt the database schema
At the very least, a new column with JSON type needs to be added to the `metrics` and `latest_metrics` tables. Ideally we would normalize the metrics IDs by having a new `metrics` table that maps a `key` and `context` to an `id`, and then reference those IDs in `metric_values` and `latest_metric_values` (new table names) instead of the `key`.

suprjinx:
This PR adds database support for json context columns in the `metrics` and `latest_metrics` tables.

Fixes #580


go.mod
```diff
@@ -39,6 +39,11 @@ require (
 	gorm.io/plugin/dbresolver v1.4.7
 )
 
+require (
+	github.com/go-sql-driver/mysql v1.7.0 // indirect
+	gorm.io/driver/mysql v1.4.7 // indirect
+)
+
```
dsuhinin, line 46: was that introduced by `datatypes` package?

suprjinx: yes

pkg/api/mlflow/dao/models/metric.go
```diff
@@ -9,6 +11,8 @@ type Metric struct {
 	Step      int64   `gorm:"default:0;not null;primaryKey"`
 	IsNan     bool    `gorm:"default:false;not null;primaryKey"`
 	Iter      int64   `gorm:"index"`
+	ContextID *uint
```
dsuhinin, line 14: maybe we have to use instead `sql.Null` type?

suprjinx: i think pointer type is implicitly nullable, yes?

dsuhinin: I think we have to follow our existing principles -> https://github.com/G-Research/fasttrackml/blob/main/pkg/database/migrations/v_0006/model.go#L50 for example.

dsuhinin: in most of the cases we use `sql.Null` types.

suprjinx: using pointer for now as with other recent additions

dsuhinin (approved):
looks good!

PR #608, opened 2023-11-16
Refactor SetupSuite/SetupTest/TearDownTest usage in tests

jgiannuzzi:
* Refactor how we use `SetupSuite`, `SetupTest`, and `TearDownTest` in our test suites to avoid some repetition for the artifact tests that use bucket storage
* Unify the naming of the method used to delete buckets
* Split `BaseTestSuite` setup between `SetupSuite` (runs once per test suite) and `SetupTest` (runs once per test)


tests/integration/golang/helpers/gs.go
```diff
@@ -9,10 +9,39 @@ import (
...
+
+func (s *BaseArtifactGSTestSuite) SetupTest() {
+	s.BaseTestSuite.SetupTest()
+	require.Nil(s.T(), CreateGSBuckets(s.GsClient, s.TestBuckets))
+}
+
+func (s *BaseArtifactGSTestSuite) TearDownTest() {
+	require.Nil(s.T(), DeleteGSBuckets(s.GsClient, s.TestBuckets))
+}
+
```
dsuhinin, line 44: let's put it into -> `helpers/test.go` because this is just another variant of `BaseTestSuite` special for some cases.

tests/integration/golang/helpers/s3.go
```diff
@@ -9,8 +9,37 @@ import (
...
+}
+
+func (s *BaseArtifactS3TestSuite) SetupTest() {
+	s.BaseTestSuite.SetupTest()
+	require.Nil(s.T(), CreateS3Buckets(s.S3Client, s.TestBuckets))
+}
+
+func (s *BaseArtifactS3TestSuite) TearDownTest() {
+	require.Nil(s.T(), DeleteS3Buckets(s.S3Client, s.TestBuckets))
+}
```
dsuhinin, line 41: the same is here.

tests/integration/golang/helpers/gs.go
```diff
@@ -9,10 +9,39 @@ import (
...
 	"cloud.google.com/go/storage"
 	"github.com/rotisserie/eris"
+	"github.com/stretchr/testify/require"
 	"google.golang.org/api/iterator"
 	"google.golang.org/api/option"
 )
 
+type BaseArtifactGSTestSuite struct {
+	BaseTestSuite
+	GsClient    *storage.Client
```
dsuhinin, line 19: I think it should be `GCClient` since `GC` is abbreviator

dsuhinin (commented):
I have one general question - do we really need such improvement? It seems like Yes and Not. +82 lines and - 93 lines, just -11 lines and all of this changes just for `artifacts` logic.

jgiannuzzi:
> I have one general question - do we really need such improvement? It seems like Yes and Not. +82 lines and - 93 lines, just -11 lines and all of this changes just for `artifacts` logic.

I made some additional changes to share more logic between the bucket storage providers, which now makes it look worse in terms of "line changes" 😬 (+179 −173). However I think this will help a lot when we introduce other bucket storage providers, and when we implement artifact proxying (which would introduce quite a few additional test suites).

I'm happy to park it for now if we don't feel like it's useful!

jgiannuzzi:
Actually my changes bring more issues than they solve, I'll close this!

PR #619, opened 2023-11-20
Extend LogMetric and LogBatch to accept context

### LINKED ISSUE #206: Metric contexts: extend the MLFlow `LogMetric` and `LogBatch` to accept an extra field `context` for each metric
- New context field is optional for callers
- Type is `map[string]any`

fabiovincenzi:
fixes #206 
Accept new optional `context` field in `LogMetric` and `LogBatch` endpoints.


pkg/api/mlflow/service/run/service.go
```diff
@@ -491,7 +491,16 @@ func (s Service) LogMetric(
 		return api.NewResourceDoesNotExistError("unable to find run '%s'", req.RunID)
 	}
 
-	metric, err := convertors.ConvertMetricParamRequestToDBModel(run.ID, req)
+	context, err := convertors.ConvertLogMetricRequestToContextDBModel(req)
```
suprjinx, line 494: Why not just include this in the existing Convert method?

pkg/api/mlflow/service/run/service.go
```diff
@@ -608,7 +617,21 @@ func (s Service) LogBatch(
 		return api.NewResourceDoesNotExistError("Unable to find active run '%s'", req.RunID)
 	}
 
-	metrics, params, tags, err := convertors.ConvertLogBatchRequestToDBModel(run.ID, req)
+	contexts, err := convertors.ConvertLogBatchRequestToContextDBModel(req)
```
suprjinx, line 620: Why not just include this in the existing Convert method?

pkg/api/mlflow/dao/convertors/log.go
```diff
@@ -72,3 +78,22 @@ func ConvertLogBatchRequestToDBModel(
 	}
 	return metrics, params, tags, nil
 }
+
+// ConvertLogBatchRequestToContextDBModel converts request.LogBatchRequest into []*models.Context model.
+func ConvertLogBatchRequestToContextDBModel(req *request.LogBatchRequest) ([]*models.Context, error) {
```
suprjinx, line 83: I'm a little confused why you separately create a slice of models.Context and then provide the slice to other converters as an argument -- couldn't you just FirstOrCreate the context at the place where the Metric is created?

dsuhinin (changes requested):

dsuhinin (dismissed):

pkg/api/mlflow/dao/repositories/metric.go
```diff
@@ -85,8 +85,17 @@ func (r MetricRepository) CreateBatch(
...
 	latestMetrics := make(map[string]models.LatestMetric)
 	for n, metric := range metrics {
+		if metric.Context != nil {
+			hash := getJsonHash(metric.Context.Json)
+			if contextMap[hash] == nil {
+				contextMap[hash] = metric.Context
+			} else {
+				metrics[n].Context = contextMap[hash]
+			}
+		}
```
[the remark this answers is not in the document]
fabiovincenzi: Because yes the context's content is the same but with a different pointer so in this way every metric with the same context will have the same reference

pkg/api/mlflow/dao/repositories/metric.go
```diff
@@ -85,8 +85,19 @@ func (r MetricRepository) CreateBatch(
 		lastIters[lastMetric.Key] = lastMetric.LastIter
 	}
 
+	uniqueContexts := make([]*models.Context, 0, len(metrics))
```
suprjinx, line 88: this var name is a little confusing to me, since this will likely be a lot of not-unique contexts (the repeats are shared). maybe "relatedContexts"?

pkg/api/mlflow/dao/repositories/metric.go
```diff
@@ -85,8 +85,19 @@ func (r MetricRepository) CreateBatch(
...
+	contextMap := make(map[string]*models.Context)
 	latestMetrics := make(map[string]models.LatestMetric)
 	for n, metric := range metrics {
+		if metric.Context != nil {
+			hash := getJsonHash(metric.Context.Json)
+			if contextMap[hash] == nil {
```
suprjinx, line 94: something like `if foundCtx := contextMap[hash]; foundCtx == nil {` might be slightly more efficient since you can then  use `foundCtx` at line 98

tests/integration/golang/mlflow/run/log_batch_test.go
```diff
@@ -367,6 +402,15 @@ func (s *LogBatchTestSuite) TestMetrics_Ok() {
 				s.Require().Nil(err)
 				s.Equal(iteration, lastMetric.LastIter)
 			}
+			for _, metric := range tt.request.Metrics {
+				if metric.Context != nil {
+					metricContextJson, err := json.Marshal(metric.Context)
+					s.Require().Nil(err)
+					context, err := s.ContextFixtures.GetContext(context.Background(), string(metricContextJson))
```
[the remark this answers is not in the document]
fabiovincenzi: We are testing that the context associated with the metric has been created

pkg/api/mlflow/dao/repositories/metric.go
suprjinx: I guess i thought these were not unique because `len(metrics)` -- ie, that you're expecting one context per metric, but they are normalized

pkg/api/mlflow/dao/repositories/metric.go
```diff
@@ -85,8 +85,19 @@ func (r MetricRepository) CreateBatch(
...
+				contextMap[hash] = metric.Context
+				uniqueContexts = append(uniqueContexts, metric.Context)
+			} else {
```
suprjinx, line 98: maybe this isn't an else but rather all cases here?

pkg/api/mlflow/dao/repositories/metric.go
fabiovincenzi: Okay sounds reasonable

pkg/api/mlflow/dao/repositories/metric.go
fabiovincenzi: actually if the context is not found in the map it would be set to nil and if the context was not found in the map metrics[n].Context already contains the right reference

suprjinx: if not found in the map it will be the next time you check (as in the else clause)

suprjinx (approved):

PR #623, opened 2023-11-20
Show namespace.Code in chooser

### LINKED ISSUE #622: Show namespace code and description in chooser
currently showing just description -- which can result in invisible namespace as description is not required.

suprjinx:
Alter namespace display in Chooser to show the namespace code instead of description. Description is shown in the "current namespace" area where it exists.

Fixes #622


suprjinx:
image
image

pkg/ui/chooser/embed/index.html
```diff
@@ -76,18 +76,25 @@
                 <p>This is the modern Aim UI, much faster than MLFlow.</p>
             </article>
         </div>
-        <p>Selected namespace: <b><u>{{.CurrentNamespace.Description}}</u></b></p>
+        <p>Selected namespace: 
+            <b>
+                <u>{{.CurrentNamespace.Code}}</u>
+                {{ if ne "" .CurrentNamespace.Description }}
+                    - {{ .CurrentNamespace.Description }}
+                {{ end }}
```
jgiannuzzi, line 84: I think we should apply this logic everywhere for consistency, probably by computing this before passing it to the template (as e.g. `DisplayName`)

suprjinx: afaik this is the only place -- it looked strange to do this in the listing because the lengths were so different

jgiannuzzi: But the listing is probably where the description is the most useful if we display it, no?
I think we either don't display the description in the chooser at all, or we display it _everywhere_ in the chooser.

suprjinx: yes left-align will probably look better, i'll try it

suprjinx: image
image

jgiannuzzi: could you please make the _text_ left-aligned and move the arrow to the left of it?
also, I _think_ it might look better with the description in parentheses, and with the default namespace being described as "Default namespace".

e.g.
```
  custom 1
  custom2 (has description)
→ default (Default namespace)
```

suprjinx: how's this?
image

fabiovincenzi (approved):

PR #626, opened 2023-11-20
Limit the amount of runs returned by GetRunsActive.

dsuhinin:
What's inside the box:
- apply default limit - `50` for AIM `GetRunsActive` endpoint to decrease overall pressure to database, service and UI.

### DIFF
--- a/pkg/api/aim/runs.go
+++ b/pkg/api/aim/runs.go
@@ -868,6 +868,12 @@ func SearchMetrics(c *fiber.Ctx) error {
 //
 //nolint:gocyclo
 func SearchAlignedMetrics(c *fiber.Ctx) error {
+	ns, err := namespace.GetNamespaceFromContext(c.Context())
+	if err != nil {
+		return api.NewInternalError("error getting namespace from context")
+	}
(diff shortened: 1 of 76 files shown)


pkg/api/aim/runs.go
```diff
@@ -240,6 +240,7 @@ func GetRunsActive(c *fiber.Ctx) error {
 	log.Debugf("getRunsActive namespace: %s", ns.Code)
 
 	q := struct {
+		Limit          int  `query:"limit"`
```
jgiannuzzi, line 243: This is not supported by the UI at this stage (and probably never will)

dsuhinin: correct. I though that we want to just limit default case.

jgiannuzzi: for now I would suggest to simply hard-code the limit to 100.

dsuhinin: ok.

dsuhinin: hardcoded.

jgiannuzzi (dismissed):

jgiannuzzi (approved):

PR #627, opened 2023-11-20
Fix timestamp types in Aim integration tests

jgiannuzzi:
This PR corrects the types used by timestamps in our Aim integration tests. We were incorrectly using integers whilst the API is returning floats. A couple of disabled assertions are also enabled.


jgiannuzzi:
Part of the work for #561.

PR #632, opened 2023-11-21
Use a consistent test style in migration tests

jgiannuzzi:
This PR improves the migration tests to use the same style as all our other tests: using a test suite and sub tests.
This means that if one migration fails but not the others, we will see it in the results, instead of having the entire test fail and not try the other migration(s).


jgiannuzzi:
Part of the work for #561.

PR #633, opened 2023-11-21
Improve the style of the import tests

jgiannuzzi:
This PR improves our import tests to be more consistent with the style of our other tests (use of test suites).
It also checks for errors directly rather than assigning to `err` first, and it enforces that the calls to the row iterator return the same values.


jgiannuzzi:
Part of the work for #561.

PR #634, opened 2023-11-21
Use assert and require in a more idiomatic way in our tests

jgiannuzzi:
This PR changes the way we use `assert` and `require` to the more idiomatic way of using the test suite directly.

For example, this
```go
assert.Equal(s.T(), expected, actual)
```

becomes
```go
s.Equal(expected, actual)
```


dsuhinin (approved):

jgiannuzzi:
Part of the work for #561.

PR #635, opened 2023-11-21
Adapt the Aim runs endpoints to return the context.

dsuhinin:
What's inside the box:
- adopt `GetRunInfo`, `GetRunMetrics`, `GetRunsActive`, `SearchRuns`, `SearchMetrics`, `SearchAlignedMetrics` to work with `context`.


pkg/api/aim/runs.go
```diff
@@ -202,31 +210,43 @@ func GetRunMetrics(c *fiber.Ctx) error {
...
 		if m.IsNan {
 			pv = nil
 		}
 
-		k.values = append(k.values, pv)
+		key := m.Key
+		if m.ContextID != nil {
+			key = fmt.Sprintf("%s%d", m.Key, *m.ContextID)
+		}
+		k := metrics[key]
```
suprjinx, line 229: what is `k`?

pkg/api/aim/runs.go
```diff
@@ -202,31 +210,43 @@ func GetRunMetrics(c *fiber.Ctx) error {
...
+		k.name = m.Key
 		k.iters = append(k.iters, int(m.Iter))
-		metrics[m.Key] = k
+		k.values = append(k.values, pv)
```
suprjinx, line 232: too many single-character var names to follow here

pkg/api/aim/runs.go
```diff
@@ -977,7 +1030,10 @@ func SearchAlignedMetrics(c *fiber.Ctx) error {
 			}
 
 			for rows.Next() {
-				var metric database.Metric
+				var metric struct {
+					database.Metric
```
suprjinx, line 1034: doesn't database.Metric already include Context into this struct?

tests/integration/golang/aim/run/get_run_info_test.go
```diff
@@ -26,20 +28,53 @@ func TestGetRunInfoTestSuite(t *testing.T) {
...
+	s.Require().Nil(err)
+
+	_, err = s.MetricFixtures.CreateLatestMetric(context.Background(), &models.LatestMetric{
+		Key:       "key",
+		Value:     123,
+		Timestamp: 123456789,
+		Step:      1,
+		IsNan:     false,
+		RunID:     run.ID,
+		ContextID: common.GetPointer(metricContext.ID),
```
suprjinx, line 66: this association will be automatic with Gorm, if the above `metricContext` is just created here as `Context`

suprjinx: well -- maybe -- verifying the same thing in my PR

tests/integration/golang/aim/run/get_run_metrics_test.go
```diff
@@ -27,13 +30,102 @@ func TestGetRunMetricsTestSuite(t *testing.T) {
...
+
+	_, err = s.MetricFixtures.CreateMetric(context.Background(), &models.Metric{
+		Key:       "key1",
+		Value:     123.1,
+		Timestamp: 123456789,
+		Step:      1,
+		IsNan:     false,
+		RunID:     run.ID,
+		Iter:      1,
+		ContextID: common.GetPointer(metricContext.ID),
```
suprjinx, line 70: see comment above -- might be able to simplify

tests/integration/golang/aim/run/get_run_metrics_test.go
```diff
@@ -27,13 +30,102 @@ func TestGetRunMetricsTestSuite(t *testing.T) {
...
+
+	_, err = s.MetricFixtures.CreateMetric(context.Background(), &models.Metric{
+		Key:       "key2",
+		Value:     124.2,
+		Timestamp: 123456789,
+		Step:      2,
+		IsNan:     false,
+		RunID:     run.ID,
+		Iter:      4,
+		ContextID: common.GetPointer(metricContext.ID),
```
suprjinx, line 124: see comment above -- might be able to simplify

pkg/api/aim/runs.go
```diff
@@ -202,31 +210,43 @@ func GetRunMetrics(c *fiber.Ctx) error {
...
+			key = fmt.Sprintf("%s%d", m.Key, *m.ContextID)
+		}
+		k := metrics[key]
+		k.name = m.Key
 		k.iters = append(k.iters, int(m.Iter))
-		metrics[m.Key] = k
+		k.values = append(k.values, pv)
+		if m.Context != nil {
+			k.context = m.Context.Json
+		} else {
```
suprjinx, line 235: why can `k.context` have different types?

pkg/api/aim/runs.go
dsuhinin: this moment was here before. this is how empty k(struct) initialised.

tests/integration/golang/aim/run/get_run_metrics_test.go
dsuhinin: yep.

tests/integration/golang/aim/run/get_run_metrics_test.go
```diff
@@ -27,13 +30,102 @@ func TestGetRunMetricsTestSuite(t *testing.T) {
...
+
+	_, err = s.MetricFixtures.CreateMetric(context.Background(), &models.Metric{
+		Key:       "key2",
+		Value:     124.1,
+		Timestamp: 123456789,
+		Step:      1,
+		IsNan:     false,
+		RunID:     run.ID,
+		Iter:      3,
+		ContextID: common.GetPointer(metricContext.ID),
```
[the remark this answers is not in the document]
dsuhinin: yep.

pkg/api/aim/runs.go
dsuhinin: because of endpoint nature. for normal(no stream) endpoints we can use it as is ->` m.Context.Json` but for the stream endpoint it just doesn't work and we have to re wrap it.

pkg/api/aim/runs.go
```diff
@@ -294,17 +314,26 @@ func GetRunsActive(c *fiber.Ctx) error {
...
 							"last_step":  m.LastIter,
 							"last":       v,
 							"version":    2,
+							"context":    fiber.Map{},
 						},
 					}
+					if m.Context != nil {
+						// to be properly decoded by AIM UI, json should be represented as a key:value object.
+						context := fiber.Map{}
+						if err := json.Unmarshal(m.Context.Json, &context); err != nil {
```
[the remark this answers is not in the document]
dsuhinin: yep, like here we have to re wrap it.

pkg/api/aim/runs.go
dsuhinin: yes, but it won't be populated unfortunately, since it is a nested object and `query` is complex so you can't norally use like `Joins("Context")`.

tests/integration/golang/aim/run/get_run_metrics_test.go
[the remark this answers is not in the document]
dsuhinin: done

pkg/api/aim/runs.go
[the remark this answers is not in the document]
dsuhinin: done.

pkg/api/aim/runs.go
```diff
@@ -1006,6 +1061,12 @@ func SearchAlignedMetrics(c *fiber.Ctx) error {
 				}
 				values = append(values, v)
 				iters = append(iters, float64(metric.Iter))
+				if metric.Context != nil {
```
suprjinx, line 1064: if metric.Context were a `map[string]any` instead of `datatypes.Json`, could we avoid these unmarshalling steps? might be worth considering

suprjinx (dismissed):

PR #638, opened 2023-11-22
Metric context query

### LINKED ISSUE #226: Find proper way to decode data of `/metrics/get-histories` endpoint.
Find proper way to decode data here -> https://github.com/G-Research/fasttrackml/blob/e85de6971c2339c48accd535d8a9e82481336bd0/tests/integration/golang/mlflow/metric/get_histories_test.go#L129 and test response.

### LINKED ISSUE #358: Metric contexts: extend the MLFlow `GetMetricHistory` and `GetMetricHistories` endpoints to filter the query down based on the new `context` field
- request: context field, optional, use for filtering when present and not empty
- filtering is done by asking the DB engine to match metrics where the requested context is a _subset_ of the recorded context
- For MVP only one level of searching is required, deeper structures will follow in new tickets.

suprjinx:
Add metric context querying in the repository layer. Metrics are filtered by a map of values, where the key is a json path, and value is what we expect at that path. For instance, given a context struct like `{ "key1": "value1", "key2": { "key3":  "value3"}}`, the context could be queried with `map[string]string{"key1": "value1", "key2.key3", "value3"}`. This seems to fit with the query inputs we expect from Aim client, such as `metric.context.key1 = value1`. 
Fixes #358 
Fixes #226


dsuhinin:
> Add metric context querying in the repository layer. Metrics are filtered by a map of values, where the key is a json path, and value is what we expect at that path. For instance, given a context struct like `{ "key1": "value1", "key2": { "key3": "value3"}}`, the context could be queried with `map[string]any{"key1": "value1", "key2.key3", "value3"}`. This seems to fit with the query inputs we expect from Aim client, such as `metric.context.key1 = value1`.
> Fixes #358

is it a mistake here -> `{"key1": "value1", "key2.key3": "value3"}` ?

dsuhinin (changes requested):
it will be nice to add several `cases` to the `GetHistoriesTestSuite_Ok` test to test actual search functionality by `context` parameters. I can'd find it.

tests/integration/golang/fixtures/metric.go
```diff
@@ -63,3 +108,21 @@ func (f MetricFixtures) GetLatestMetricByRunID(ctx context.Context, runID string
...
+		for k, v := range jsonPathValueMap {
+			path := strings.ReplaceAll(k, ".", ",")
+			tx.Where(fmt.Sprintf("%s#>>'{%s}' = ?", columnName, path), v)
+		}
+	default:
+		for k, v := range jsonPathValueMap {
+			tx.Where(fmt.Sprintf("%s->>'%s' = ?", columnName, k), v)
+		}
+	}
+}
```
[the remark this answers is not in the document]
suprjinx: well this one was a test fixture, and I thought it shouldn't use the very code under test for setup

pkg/api/mlflow/dao/repositories/helpers.go
```diff
@@ -30,3 +33,22 @@ func makeParamConflictPlaceholdersAndValues(params []models.Param) (string, []in
...
+		for k, v := range jsonPathValueMap {
+			path := strings.ReplaceAll(k, ".", ",")
+			tx.Where(fmt.Sprintf("%s#>>'{%s}' = ?", jsonColumnName, path), v)
+		}
+	default:
+		for k, v := range jsonPathValueMap {
+			tx.Where(fmt.Sprintf("%s->>'%s' = ?", jsonColumnName, k), v)
+		}
+	}
+}
```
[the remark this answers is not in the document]
suprjinx: i think this will be awkward because `addJsonCondition` would have to return tuple and only the second is expanded, yes?

pkg/api/mlflow/service/metric/service.go
```diff
@@ -89,6 +89,7 @@ func (s Service) GetMetricHistories(
 		req.MetricKeys,
 		req.ViewType,
 		req.MaxResults,
+		map[string]any{},
```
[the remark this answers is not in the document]
suprjinx: this is the boundary of this story I think, since that map comes from controller/request (different story)

tests/integration/golang/mlflow/metric/get_histories_test.go
```diff
@@ -45,16 +46,64 @@ func (s *GetHistoriesTestSuite) Test_Ok() {
 	})
 	s.Require().Nil(err)
 
-	_, err = s.MetricFixtures.CreateMetric(context.Background(), &models.Metric{
+	m, err := s.MetricFixtures.CreateMetric(context.Background(), &models.Metric{
```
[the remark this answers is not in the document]
suprjinx: ok

tests/integration/golang/fixtures/metric.go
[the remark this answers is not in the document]
dsuhinin: but we use a lot of code from the service: request, response objects, repositories, models, etc. so it will be just fine.

pkg/api/mlflow/dao/repositories/helpers.go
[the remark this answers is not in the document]
dsuhinin: I thought about this only because of the moment related to the tests.  but why awkward? functions can return `tuple` like that:
```
func asdasd() (query interface{}, args []interface{}){}
```
I don't know for me it looks a bit cleaner and testable.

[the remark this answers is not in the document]
suprjinx: i think the current code is less awkward because the condition needs to be a string and a value for each context param -- so we'd have to make one big string and a values slice. I think for db-related functionality we should lean on the integration tests because that's the only way we know the sql is correct.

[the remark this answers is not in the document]
dsuhinin: no no, Im talking about `unit` tests for this part. that's why I want to get rid of `db` here.

tests/integration/golang/fixtures/metric.go
```diff
@@ -26,18 +30,59 @@ func NewMetricFixtures(db *gorm.DB) (*MetricFixtures, error) {
...
+			return nil, eris.Wrap(err, "error creating metric context")
+		}
+		metric.ContextID = &metric.Context.ID
+	}
 	if err := f.baseFixtures.db.WithContext(ctx).Create(metric).Error; err != nil {
-		return nil, eris.Wrap(err, "error creating test metric")
+		return nil, eris.Wrap(err, "error creating metric")
 	}
 	return metric, nil
 }
```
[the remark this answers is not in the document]
suprjinx: Gorm is not doing the magic we need -- it could perhaps become a `BeforeCreate` hook on the Metric and LatestMetric model?

tests/integration/golang/fixtures/metric.go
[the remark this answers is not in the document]
suprjinx: okay

pkg/api/mlflow/dao/repositories/helpers.go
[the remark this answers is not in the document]
suprjinx: i think the code will be a lot less clean (collecting a large sql clause, collecting a values slice, returning both as a tuple) and then we still need to run it in a database (integration test) to have any confidence it's correct so the unit test is low value.

tests/integration/golang/fixtures/metric.go
```diff
@@ -26,18 +30,59 @@ func NewMetricFixtures(db *gorm.DB) (*MetricFixtures, error) {
...
+	if metric.Context != nil {
+		if err := f.baseFixtures.db.WithContext(ctx).FirstOrCreate(&metric.Context).Error; err != nil {
+			return nil, eris.Wrap(err, "error creating latest metric context")
+		}
+		metric.ContextID = &metric.Context.ID
+	}
 	if err := f.baseFixtures.db.WithContext(ctx).Create(metric).Error; err != nil {
-		return nil, eris.Wrap(err, "error creating test latest metric")
+		return nil, eris.Wrap(err, "error creating latest metric")
 	}
```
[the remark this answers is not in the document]
suprjinx: gorm is not able to do "FirstOrCreate" on the association by magic, but I think we could move to Metric.BeforeCreate hook if that's preferable (?)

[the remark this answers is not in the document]
dsuhinin: that's a bit strange cause they say:
```
When creating some data with associations, if its associations value is not zero-value, those associations will be upserted, and its Hooks methods will be invoked.
``` - upsert. if no, then ok.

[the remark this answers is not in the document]
Comment (suprjinx) [reply tests/integration/golang/fixtures/metric.go:83]: my testing did not show that to work in this case, possibly because our FK relationship is inverted (ie, we'd have to create Context and then it could auto create Context.Metrics if we had that association defined).

File: pkg/api/mlflow/service/metric/service.go (hunk 3 above)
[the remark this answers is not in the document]
Comment (suprjinx) [reply pkg/api/mlflow/service/metric/service.go]: you are right, now passing through from request

File: pkg/api/mlflow/dao/repositories/helpers.go [hunk 7]
```diff
@@ -30,3 +33,22 @@ func makeParamConflictPlaceholdersAndValues(params []models.Param) (string, []in
 	}
 	return placeholders, valuesArray
 }
+
+// AddJsonCondition adds where condition(s) to the query to select items having the specified map of json paths
+// and values in the given json column. Json path is expressed as "key" or "outerkey.nestedKey".
+func AddJsonCondition(tx *gorm.DB, jsonColumnName string, jsonPathValueMap map[string]string) {
```
dave-gantenbein, line 39: Testme.

pkg/api/mlflow/dao/repositories/metric.go
```diff
@@ -219,6 +221,11 @@ func (r MetricRepository) GetMetricHistories(
 		query.Where("metrics.key IN ?", metricKeys)
 	}
 
+	if len(jsonPathValueMap) > 0 {
```
dave-gantenbein, line 224: Testme.

pkg/api/mlflow/dao/repositories/helpers.go
suprjinx: tested in integration `get_histories_test.go`, which validates the sql with dbs.

dave-gantenbein: Thank you!

pkg/api/mlflow/dao/repositories/helpers.go
```diff
@@ -30,3 +32,32 @@ func makeParamConflictPlaceholdersAndValues(params []models.Param) (string, []in
...
+	if len(jsonPathValueMap) == 0 {
+		return
+	}
+	var conditionTemplate string
+	switch dialector {
+	case postgres.Dialector{}.Name():
+		conditionTemplate = "%s#>>? = ?"
+		for k, v := range jsonPathValueMap {
+			path := strings.ReplaceAll(k, ".", ",")
+			args = append(args, "{"+path+"}", v)
```
dsuhinin, line 52: we can make it a bit efficient if we allocate the size of `args` based on `jsonPathValueMap` length.

pkg/api/mlflow/dao/repositories/helpers.go
```diff
@@ -30,3 +32,32 @@ func makeParamConflictPlaceholdersAndValues(params []models.Param) (string, []in
...
+		}
+	default:
+		conditionTemplate = "%s->>? = ?"
+		for k, v := range jsonPathValueMap {
+			args = append(args, k, v)
+		}
+	}
+	conditionTemplate = fmt.Sprintf(conditionTemplate, jsonColumnName)
+	sql = strings.Repeat(conditionTemplate+" AND ", len(jsonPathValueMap)-1) + conditionTemplate
+	return
```
dsuhinin, line 62: ohh, naming parameters always looks strange. can we please just return `sql` and `args` explicitly?

pkg/api/mlflow/dao/repositories/helpers_test.go
```diff
@@ -53,3 +53,53 @@ func Test_makeParamConflictPlaceholdersAndValues(t *testing.T) {
...
+		name             string
+		dialector        string
+		jsonColumnName   string
+		jsonPathValueMap map[string]string
+		expectedSQL      string
+		expectedArgs     []interface{}
+	}{
+		{
+			name:           "Postgres",
+			dialector:      "postgres",
```
dsuhinin, line 68: we have a `SQLiteDialectorName` and `PostgresDialectorName` in `database/constants.go`, can you please re use them?

pkg/api/mlflow/dao/repositories/metric.go
```diff
@@ -219,6 +221,12 @@ func (r MetricRepository) GetMetricHistories(
...
+		query.Joins("LEFT JOIN contexts on metrics.context_id = contexts.id")
+		sql, args := BuildJsonCondition(query.Dialector.Name(), "contexts.json", jsonPathValueMap)
+		query.Where(sql, args...)
```
dsuhinin, line 227: if you pass `BuildJsonCondition` immediately   to `query.Where`, won't it work?

tests/integration/golang/fixtures/metric.go
```diff
@@ -26,18 +27,60 @@ func NewMetricFixtures(db *gorm.DB) (*MetricFixtures, error) {
...
+// GetMetricsByContext returns metric by a context partial match.
+func (f MetricFixtures) GetMetricsByContext(
+	ctx context.Context,
+	metricContext map[string]string,
+) ([]*models.Metric, error) {
+	var metrics []*models.Metric
+	tx := f.db.WithContext(ctx).Model(
+		&database.Metric{},
+	).Joins(
+		"LEFT JOIN contexts on metrics.context_id = contexts.id",
```
dsuhinin, line 62: can you please change `on` -> `ON`

dsuhinin: you can even just `preload` and make it shorter.

tests/integration/golang/fixtures/metric.go
```diff
@@ -26,18 +27,60 @@ func NewMetricFixtures(db *gorm.DB) (*MetricFixtures, error) {
...
+	)
+	sql, args := repositories.BuildJsonCondition(tx.Dialector.Name(), "contexts.json", metricContext)
+	tx.Where(sql, args...)
+	if err := tx.Find(&metrics).Error; err != nil {
+		return nil, eris.Wrapf(err, "error getting metric by context: %v", metricContext)
```
dsuhinin, line 67: can you please change `metric` -> `metrics`.

tests/integration/golang/fixtures/metric.go
```diff
@@ -26,18 +27,60 @@ func NewMetricFixtures(db *gorm.DB) (*MetricFixtures, error) {
...
+) ([]*models.Metric, error) {
+	var metrics []*models.Metric
+	tx := f.db.WithContext(ctx).Model(
+		&database.Metric{},
+	).Joins(
+		"LEFT JOIN contexts on metrics.context_id = contexts.id",
+	)
+	sql, args := repositories.BuildJsonCondition(tx.Dialector.Name(), "contexts.json", metricContext)
+	tx.Where(sql, args...)
+	if err := tx.Find(&metrics).Error; err != nil {
```
dsuhinin, line 66: you can probably use it like:
```
tx.Where(sql, args...).Find()
```

tests/integration/golang/fixtures/metric.go
```diff
@@ -26,18 +27,60 @@ func NewMetricFixtures(db *gorm.DB) (*MetricFixtures, error) {
...
+	ctx context.Context,
+	metricContext map[string]string,
+) ([]*models.Metric, error) {
+	var metrics []*models.Metric
+	tx := f.db.WithContext(ctx).Model(
+		&database.Metric{},
+	).Joins(
+		"LEFT JOIN contexts on metrics.context_id = contexts.id",
+	)
+	sql, args := repositories.BuildJsonCondition(tx.Dialector.Name(), "contexts.json", metricContext)
```
dsuhinin, line 64: can't we pass it right into `Where`?

tests/integration/golang/helpers/arrow.go
```diff
@@ -0,0 +1,49 @@
...
+	"github.com/G-Research/fasttrackml/pkg/api/mlflow/dao/models"
+)
+
+func DecodeArrowMetrics(buf *bytes.Buffer) ([]models.Metric, error) {
+	pool := memory.NewGoAllocator()
+
+	// Create a new reader
+	reader, err := ipc.NewReader(buf, ipc.WithAllocator(pool))
+	if err != nil {
+		return nil, err
```
dsuhinin, line 19: can we please wrap `err`?

tests/integration/golang/helpers/arrow.go
```diff
@@ -0,0 +1,49 @@
...
+			if !metric.IsNan {
+				metric.Value = rec.Column(4).(*array.Float64).Value(i)
+			}
+			metrics = append(metrics, metric)
+		}
+		rec.Release()
+	}
+
+	if reader.Err() != nil {
+		return nil, reader.Err()
```
dsuhinin, line 45: and wrap it here please?

pkg/api/mlflow/dao/repositories/helpers.go
dsuhinin, line 52: since we always use `fmt.Sprintf` pattern, will be nice to use it instead of `"{"+path+"}"`

dave-gantenbein: Seems more verbose to me, is there some benefit performance-wise?

pkg/api/mlflow/dao/repositories/metric.go
dave-gantenbein: Probably, but then you've got a very busy line of code to unpack...

pkg/api/mlflow/dao/repositories/helpers.go
suprjinx: well i thought this was a case where the names helped for the return vals in the signature -- but I will make explicit return here

pkg/api/mlflow/dao/repositories/metric.go
suprjinx: it won't work

suprjinx: i think you have to expand the second arg with `...`, which is only possible in this way

pkg/api/mlflow/dao/repositories/helpers.go
suprjinx: ok

pkg/api/mlflow/dao/repositories/helpers_test.go
suprjinx: ok

tests/integration/golang/fixtures/metric.go
suprjinx: preload does not result in the expected column existence, so changing to "ON"

tests/integration/golang/fixtures/metric.go
suprjinx: also we have to `...` the second return val in this approach

tests/integration/golang/fixtures/metric.go
dsuhinin: hm, thats a bit strange because in most of other places it works until you start using like a custom/raw queries.

tests/integration/golang/fixtures/metric.go
suprjinx: ok

pkg/api/mlflow/dao/repositories/helpers.go
suprjinx: changed it to fmt.Sprintf

dsuhinin (approved):
Looks good!

PR #645, opened 2023-11-23
Refactor test suite setup/teardown

jgiannuzzi:
This PR is a big refactor of the way we setup our integration tests. It aims at reducing some of the boilerplate we have in each test suite. I would recommend reviewing by looking at the 4 commits individually.
* [Refactor BaseTestSuite helper to provide setup/teardown](https://github.com/G-Research/fasttrackml/pull/645/commits/b371762452289ed61d508870803ed35fd3326b13) lays down the foundation for this refactor 🏗️ 
* [Refactor bucket storage provider helpers to provide setup/teardown](https://github.com/G-Research/fasttrackml/pull/645/commits/65b4ebe0f11efb59318944cfad06ded773870516) applies these principles to the GCP and S3 helpers 🔨 
* [Adapt all tests to use refactored setup/teardown](https://github.com/G-Research/fasttrackml/pull/645/commits/c38a5a0610a4b6a21252abafecb4ecccc8a58074) is the one where the boilerplate code gets removed/adapted 🗑️
* [Rename UnloadFixtures to TruncateTables](https://github.com/G-Research/fasttrackml/pull/645/commits/48fac842be1bdb7012f0b215da7c291af4525687) is self-explanatory 😄

### DIFF
--- a/tests/integration/golang/aim/app/delete_app_test.go
+++ b/tests/integration/golang/aim/app/delete_app_test.go
@@ -12,8 +12,6 @@ import (
 	"github.com/stretchr/testify/suite"
 
 	"github.com/G-Research/fasttrackml/pkg/api/aim/response"
-	"github.com/G-Research/fasttrackml/pkg/api/mlflow/common"
-	"github.com/G-Research/fasttrackml/pkg/api/mlflow/dao/models"
 	"github.com/G-Research/fasttrackml/pkg/database"
 	"github.com/G-Research/fasttrackml/tests/integration/golang/helpers"
 )
@@ -27,17 +25,6 @@ func TestDeleteAppTestSuite(t *testing.T) {
 }
 
 func (s *DeleteAppTestSuite) Test_Ok() {
-	defer func() {
-		s.Require().Nil(s.NamespaceFixtures.UnloadFixtures())
-	}()
-
-	namespace, err := s.NamespaceFixtures.CreateNamespace(context.Background(), &models.Namespace{
-		ID:                  1,
-		Code:                "default",
-		DefaultExperimentID: common.GetPointer(int32(0)),
-	})
-	s.Require().Nil(err)
-
 	app, err := s.AppFixtures.CreateApp(context.Background(), &database.App{
 		Base: database.Base{
 			ID:        uuid.New(),
@@ -45,7 +32,7 @@ func (s *DeleteAppTestSuite) Test_Ok() {
 		},
 		Type:        "mpi",
 		State:       database.AppState{},
-		NamespaceID: namespace.ID,
+		NamespaceID: s.DefaultNamespace.ID,
 	})
 	s.Require().Nil(err)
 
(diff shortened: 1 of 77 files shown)


tests/integration/golang/helpers/test.go
```diff
@@ -85,6 +95,80 @@ func (s *BaseTestSuite) SetupTest() {
...
+	for _, hook := range s.setupHooks {
+		hook()
+	}
+}
+
+func (s *BaseTestSuite) SetupTest() {
+	if !s.ResetOnSubTest {
+		s.runSetupHooks()
+	}
+}
```
dsuhinin, line 142: let's maybe group private and public function. now a bit hard to navigate over them. maybe let's have next grouping:
- firstly have private functions.
- then public functions.

tests/integration/golang/helpers/test.go
```diff
@@ -85,6 +95,80 @@ func (s *BaseTestSuite) SetupTest() {
...
+
+func (s *BaseTestSuite) SetupSubTest() {
+	if s.ResetOnSubTest {
+		s.runSetupHooks()
+	}
+}
+
+func (s *BaseTestSuite) AddTearDownHook(hook func()) {
+	s.tearDownHooks = append([]func(){hook}, s.tearDownHooks...)
```
dsuhinin, line 151: why just not -> `s.tearDownHooks = append(s.tearDownHooks, hook)`

tests/integration/golang/aim/app/delete_app_test.go
```diff
@@ -75,25 +62,14 @@ func (s *DeleteAppTestSuite) Test_Ok() {
...
-	}()
-
-	namespace, err := s.NamespaceFixtures.CreateNamespace(context.Background(), &models.Namespace{
-		ID:                  1,
-		Code:                "default",
-		DefaultExperimentID: common.GetPointer(int32(0)),
-	})
-	s.Require().Nil(err)
-
-	_, err = s.AppFixtures.CreateApp(context.Background(), &database.App{
```
dsuhinin, line 89: oooo yes, this is what I wanted to do =)

tests/integration/golang/mlflow/run/log_metric_test.go
```diff
@@ -29,33 +28,14 @@ func TestLogMetricTestSuite(t *testing.T) {
...
-	s.Require().Nil(err)
-
 	run := &models.Run{
 		ID:             strings.ReplaceAll(uuid.New().String(), "-", ""),
-		ExperimentID:   *experiment.ID,
+		ExperimentID:   *s.DefaultExperiment.ID,
 		SourceType:     "JOB",
 		LifecycleStage: models.LifecycleStageActive,
 		Status:         models.StatusRunning,
 	}
```
dsuhinin, line 37: since `CreateRun` returns `run` I think we can send `models.Run` object immediately  to the `CreateRun` function.

tests/integration/golang/mlflow/run/log_metric_test.go
```diff
@@ -167,35 +110,17 @@ func (s *LogMetricTestSuite) Test_Error() {
...
 				run := &models.Run{
 					ID:             strings.ReplaceAll(uuid.New().String(), "-", ""),
-					ExperimentID:   *experiment.ID,
+					ExperimentID:   *s.DefaultExperiment.ID,
 					SourceType:     "JOB",
 					LifecycleStage: models.LifecycleStageActive,
 					Status:         models.StatusRunning,
 				}
-				run, err = s.RunFixtures.CreateRun(context.Background(), run)
+				run, err := s.RunFixtures.CreateRun(context.Background(), run)
```
dsuhinin, line 120: the same here, we can pass `models.Run` object right into `CreateRun` function.

dsuhinin (dismissed):
Amazing! That's huge improvement.

tests/integration/golang/helpers/test.go
jgiannuzzi: because I would like this list to be LIFO, like the way `defer` works

dsuhinin: I mean we have no dependency on ordering of `tearDownHooks` it seems like. at least before.

jgiannuzzi: I think we will once we introduce the server, which is why I implemented it that way

dsuhinin: ok, let's see then!

dsuhinin (approved):
Nice!

PR #675, opened 2023-12-01
Fix flaky Run flow integration test

jgiannuzzi:
This test assumed that the ordering of the tags, params, and metrics was guaranteed by the server, but it isn't. This PR changes the way we check for slice equality to allow for any order.


jgiannuzzi:
> ok! don't we need just order them properly on server side instead?

well the order is not and should not be guaranteed by the API IMHO — I _think_ it may even make things slightly slower if we did, but happy for someone to give it a try if they want to!

PR #676, opened 2023-12-01
Move server init to its own package

jgiannuzzi:
This PR lays down the infrastructure for #561 by moving the server init to its own package and interface.


pkg/server/server.go
```diff
@@ -0,0 +1,254 @@
...
+	chooserController "github.com/G-Research/fasttrackml/pkg/ui/chooser/controller"
+	mlflowUI "github.com/G-Research/fasttrackml/pkg/ui/mlflow"
+	"github.com/G-Research/fasttrackml/pkg/version"
+)
+
+type Server interface {
+	Listen(address string) error
+	ShutdownWithTimeout(timeout time.Duration) error
+	Test(req *http.Request, msTimeout ...int) (*http.Response, error)
+}
```
dsuhinin, line 49: not fully sure that we need such interface. I think we can just share `fiber.App` because this is probably high level component, which we are going to call in 2 places:
- `cmd/server.go`
- tests
so no other dependencies which really needs such interface.

suprjinx (dismissed):

pkg/server/server.go
```diff
@@ -0,0 +1,254 @@
...
+
+type server struct {
+	*fiber.App
+}
+
+func NewServer(ctx context.Context, config *mlflowConfig.ServiceConfig) (Server, error) {
```
dsuhinin, line 55: lets add comments

pkg/server/server.go
```diff
@@ -0,0 +1,254 @@
...
+	// init database connection.
+	db, err := initDB(config)
+	if err != nil {
+		return nil, err
+	}
+
+	var namespaceRepository repositories.NamespaceRepositoryProvider
+	var artifactStorageFactory storage.ArtifactStorageFactoryProvider
+	if err := func() error {
```
dsuhinin, line 64: this `if` guy looks a bit strange. Im understand why but in general can we maybe do a bit differenttly?

dsuhinin: maybe we can create `db` in  `serverCmd` and pass it into `NewServer` function. Then we can analyse `err` from `NewServer` and just close `db` in that case. So in that case everything could look a bit more understandable.

dsuhinin (changes requested):
Left few comments.

pkg/server/server.go
jgiannuzzi: good point - I have tried yet another approach of create the namespace repository and its listener in a separate function, which hopefully makes the whole thing simpler to grasp.

dsuhinin (approved):
It looks better. I still think that `db.Close` should be in other place probably(where the config initiated) but we can move forward with this approach without any problem.

PR #677, opened 2023-12-03
Make hidden `decode` command output streamed again.

dsuhinin:


pkg/api/aim/encoding/decoder.go
```diff
@@ -32,17 +32,174 @@ func (d *reader) readField() ([]byte, error) {
...
 	}
 	return data, nil
 }
 
-// Decode decodes input stream of data into map[string]interface{}.
+// DecoderResult represents result of one piece of data after decoding.
+type DecoderResult struct {
+	Data    map[string]interface{}
+	Error   error
+	Portion int
```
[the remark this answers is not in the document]
dsuhinin: yep, just for debugging purpose and in general we can see and check how many entities we've received.

pkg/api/aim/encoding/decoder.go
```diff
@@ -32,17 +32,174 @@ func (d *reader) readField() ([]byte, error) {
...
+}
+
+// DecoderProvider provides an interface to work with Decoder.
+type DecoderProvider interface {
+	// Decode represents syntactic sugar function which returns decoded stream data at once.
+	// this function is just for back compatibility with integration tests.
+	Decode() (map[string]interface{}, error)
+	// DecodeByChunk decodes stream data by chunks.
+	DecodeByChunk() <-chan DecoderResult
```
jgiannuzzi, line 58: I _think_ that we should have an interface that is closer to an iterator (like in the GCS and S3 APIs). Exposing a raw channel is a bit too low-level IMO. The underlying implementation can totally use a channel if you'd like, though I am not convinced of the actual performance benefits of it. The implementation may actually be simpler and more readable without channels.

dsuhinin: ok, let me changes it then.

dsuhinin: ok, I think now it should be similar with S3 or GCS.

suprjinx:
I'm a little unclear why there's a `decoder.go` at the repo root and a similar in `pkg`. Is that intentional?

pkg/api/aim/encoding/decoder.go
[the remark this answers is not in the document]
dsuhinin: deleted `DecoderResult`, because we don't need implementation based on `channels`.

pkg/api/aim/encoding/decoder.go
```diff
@@ -42,36 +43,70 @@ func (d *reader) readField() ([]byte, error) {
...
+	// Decode decodes all the data stream at once.
+	Decode() (map[string]interface{}, error)
+}
+
+// Decoder represents decoders for streaming data.
+type Decoder struct {
+	path     []string
+	reader   reader
+	cursor   string
+	position int
```
jgiannuzzi, line 59: this does not seem to be used anymore

pkg/api/aim/encoding/decoder.go
```diff
@@ -42,36 +43,70 @@ func (d *reader) readField() ([]byte, error) {
...
+				}
 			}
 		}
 
-		valuebuf, err := d.readField()
+		if len(d.path[0]) > 0 && d.cursor == "" {
+			d.cursor = d.path[0]
+		}
+
+		// if `current` has been changed, then we have to release current object.
```
jgiannuzzi, line 102: we should settle on either `current` or `cursor`

jgiannuzzi (commented):
2 tiny comments, but otherwise LGTM 👍

pkg/api/aim/encoding/decoder.go
dsuhinin: yep, left it for debugging purpose but I can remove it.

pkg/api/aim/encoding/decoder.go
dsuhinin: yep. done.

jgiannuzzi (approved):

PR #682, opened 2023-12-04
add context dict to python client for get-metric-histories

suprjinx:
this will supply context dictionary to the FML GetMetricHistories endpoint


dave-gantenbein:
Geoff says no

PR #685, opened 2023-12-05
WIP querying context.json with aim search

suprjinx:


pkg/api/aim/query/clause.go
```diff
@@ -52,3 +56,146 @@ func (regexp Regexp) writeColumn(builder clause.Builder) {
...
+	}
+}
+
+type JsonEq struct {
+	Left  Json
+	Value any
+}
+
+func (eq JsonEq) Build(builder clause.Builder) {
+	eq.Left.Build(builder)
```
suprjinx, line 118: i had to "override" these methods from the Gorm built-in so we could call `Build` on the left-hand portion, instead of writing quoted

fabiovincenzi (commented):
Screenshot 2023-12-12 at 11 22 23
Tried a bit from the UI, and it seems to retrieve metrics that don't have the context key/value specified.

suprjinx:
> Screenshot 2023-12-12 at 11 22 23 Tried a bit from the UI, and it seems to retrieve metrics that don't have the context key/value specified.

@fabiovincenzi I am not able to reproduce -- the filtering works correctly for me. Screenshots:
image
image

fabiovincenzi (approved):

PR #687, opened 2023-12-06
Create a new server instance per test in the integration tests

jgiannuzzi:
This PR changes our integration tests to have a new unique server instance per test, allowing us to debug end-to-end and simplifying the setup.


dsuhinin (approved):
Nice! What is really important that nothing from high levels were affected. all changes just internal so we are on a correct way. Nice!!

PR #695, opened 2023-12-07
Fix import from/to postgres

jgiannuzzi:
Importing from/to a Postgres database was broken and this PR aims at fixing that by using the correct types where needed. We probably want to rewrite this piece of code to use our DAO instead in the future.


dave-gantenbein (approved):
ship it!

PR #714, opened 2023-12-08
Fix default namespace bookmarks disappearing after migration

jgiannuzzi:
During our database migration to namespaces, we didn't do a full data migration for the new foreign keys added to `experiments` and `apps`. Instead we did a partial data migration during the creation of the default namespace in which we updated the existing experiments. However bookmarks (`apps`) were not not updated.

This PR introduces a new migration `v_0008` that:
1. creates the default namespace if it doesn't exist
2. updates all experiments and apps that do not belong to a namespace to be in the default namespace
3. adds a `NOT NULL` constraint to the namespace foreign keys

The default namespace creation is also changed to _not_ do any data migration anymore, as it doesn't belong in there.

Finally, the "context" migration is renamed to `v_0009`, as the `v_0008` migration will need to be backported to the `release/0.4` branch.


suprjinx (approved):
lgtm

PR #725, opened 2023-12-11
Restore cache check to GetByCode

### LINKED ISSUE #724: restore namespace caching
`NamespaceCachedRepository.GetByCode` no longer checks cache for namespace:

Image

suprjinx:
Fixes #724


jgiannuzzi (approved):

pkg/api/mlflow/dao/repositories/namespace_cached.go
```diff
@@ -107,6 +107,10 @@ func (r NamespaceCachedRepository) Update(ctx context.Context, namespace *models
 func (r NamespaceCachedRepository) GetByCode(
 	ctx context.Context, code string,
 ) (*models.Namespace, error) {
+	result, ok := r.cache.Get(code)
+	if ok {
+		return &result, nil
+	}
```
jgiannuzzi, line 113: Please add a blank line after this (like it was before I mistakenly removed this piece of code)

jgiannuzzi (commented):
are the tests still necessary?
if yes, should we move them to regular (non-table) tests?
and/or do we want to test more than one arbitrary endpoint?

pkg/api/mlflow/dao/repositories/namespace_cached.go
suprjinx: ok

suprjinx:
> are the tests still necessary? if yes, should we move them to regular (non-table) tests? and/or do we want to test more than one arbitrary endpoint?

I think the remaining tests still have some value, since they check non-existent namespace in both UIs (something not covered in the flow tests). As for table-style, I personally think it's worth keeping that format even when there's only one case, since it matches style of multi-case tests and makes it easier to grok.

jgiannuzzi (approved):

PR #739, opened 2023-12-12
Alter contexts schema

suprjinx:
this PR adjust the metric `contexts` schema to be not-null, with an empty context document as the default


suprjinx:
closing this -- with some testing, it appears performance is slightly better in Postgres but considerably worse in Sqlite -- on net, best to keep context as nullable field

PR #744, opened 2023-12-13
Support for Context in LogMetric and LogBatch in Python API

### LINKED ISSUE #209: Metric contexts: add new methods to the Python API for metric logging via MLFlow endpoints
- `log_metric` and `log_batch` need to support context
- allow a new field: `context`, which is not required
- if present, the value of the `context` field cannot be null
- needs to be a `dict` object that can be serialized to proper JSON
- if object cannot be serialized, return an exception which wraps the offending context (`InvalidParameterValue` exception)

fabiovincenzi:
Fixes #209.
Created a new `fasttrackml` python package which wraps and extends the `mlflow` package.
Added support for context in `log_metric` and `log_metrics`, they both now include an optional `context` field as a `dict` object.


python/fasttrackml/_tracking_service/client.py
```diff
@@ -0,0 +1,29 @@
...
+from mlflow.utils.validation import MAX_METRICS_PER_BATCH
+
+
+class FasttrackmlTrackingServiceClient(TrackingServiceClient):
+
+    def __init__(self, tracking_uri):
+        super().__init__(tracking_uri)
+        self.store_with_context = CustomRestStore(lambda: MlflowHostCreds(self.tracking_uri))
+
+    def log_metric(self, run_id, key, value, timestamp=None, step=None, context=None):
```
suprjinx, line 19: should we have type hints for the attributes?

fabiovincenzi: We do have type hints in fluent.py, here I didn't use type hints just because in the original implementation in MLflow there aren't but if you think it's better to have theme also here I think we could just add them.

suprjinx: yes, I think they're worth adding

docs/fasttrackml_python_client.md
```diff
@@ -0,0 +1,59 @@
+
+## FastTrackML Python Client Overview
+
+`fasttrackml` is a Python package that fully leverage the capabilities of MLFlow, FastTrackML extends its functionality by providing additional methods for a more convenient and efficient user experience.
```
suprjinx, line 4: this paragraph needs some general revision for grammar

suprjinx (approved):
LGTM

PR #745, opened 2023-12-13
new test showing perf difference with nullable context

suprjinx:


suprjinx:
findings suggest no change needed at this time

PR #763, opened 2023-12-20
Upgrade MLFlow integration tests to 2.9.2

jgiannuzzi:
This PR upgrades our MLFlow integration tests to 2.9.2, enables a few additional tests (artifacts and path validation), disables the SQLite memory backend, and makes a few changes to our code for increased compatibility with MLFlow:
- the `file://` prefix gets added to the default artifact root when a local path is provided
- a distinction is made between generic SQL errors and a run not being found when trying to delete a run tag
- the error message returned for an invalid path has been changed to "Invalid path" (and the validation logic has been DRYed up)
- the error message when an inexistent run is provided has been changed to "Run '_id_' not found"
- the error message when an inexistent tag is provided to `DeleteRunTag` has been changed to "No tag with name: _key_"
- the validator for `LogBatch` is more exhaustive

Our own tests have also been updated to check for these changes.


tests/integration/python/mlflow.patch
```diff
@@ -1,14 +1,30 @@
...
+-            "-m",
+-            "flask",
+-            "--app",
+-            app,
+-            "run",
+-            "--host",
+-            LOCALHOST,
+-            "--port",
+-            str(server_port),
 +            "fml",
```
suprjinx, line 28: 👍

suprjinx (approved):

PR #776, opened 2023-12-27
Fix Run info badge for running runs

### LINKED ISSUE #655: Continuing with Experiment Run fails to show "active" on the run page.
This is a run that has been restarted with `with mlflow.start_run(run_id=run_id) as run:`

https://photos.app.goo.gl/9K8Riq92T9W4fnfAA it shows active on the main page

But not on the runs page: https://photos.app.goo.gl/HVVgYrTk3H69WGoN6

Also:
- The time active for the run does not start increasing
- The activity log for the experiment does not show any events

The important stuff however, like the metrics, still work so this is not too big a problem. Thank you

fabiovincenzi:
Fixes #655 
This PR addresses the issue related to the display of the active state of runs after a restart.
AIM UI shows the status badge based on the Endtime of a run so we need to set Endtime to null when restarted.


pkg/api/mlflow/dao/convertors/run.go
```diff
@@ -86,9 +86,13 @@ func ConvertCreateRunRequestToDBModel(
...
-	run.EndTime = sql.NullInt64{
-		Int64: req.EndTime,
-		Valid: true,
+	if req.Status == string(models.StatusRunning) {
+		run.EndTime = sql.NullInt64{}
+	} else {
+		run.EndTime = sql.NullInt64{
+			Int64: req.EndTime,
+			Valid: true,
+		}
```
dsuhinin, line 95: can you please add a comment here so we can always know why we need this `if` statement.

pkg/api/mlflow/dao/repositories/run.go
```diff
@@ -292,6 +292,12 @@ func (r RunRepository) UpdateWithTransaction(ctx context.Context, tx *gorm.DB, r
 	if err := tx.WithContext(ctx).Model(&run).Updates(run).Error; err != nil {
 		return eris.Wrapf(err, "error updating existing run with id: %s", run.ID)
 	}
+	// Use UpdateColumn so we can reset EndTime to null
+	if run.EndTime.Int64 == 0 {
```
dsuhinin, line 296: EndTime has `IsValid` so in most cases it is much understandable to use it.

pkg/api/mlflow/dao/repositories/run.go
```diff
@@ -292,6 +292,12 @@ func (r RunRepository) UpdateWithTransaction(ctx context.Context, tx *gorm.DB, r
...
+		if err := tx.Model(&run).UpdateColumn("EndTime", sql.NullInt64{}).Error; err != nil {
```
dsuhinin, line 297: we specially use `UpdateColumn` to just make field nullable, right?

dsuhinin (changes requested):
I can't see a test for this special case. If it possible it would be really nice to have it.

pkg/api/mlflow/dao/repositories/run.go
fabiovincenzi: Exactly, regular `Updates` function skip nil values

dsuhinin: even if you pass like `sql.NullInt64{IsValid: true}`?

fabiovincenzi: It actually works! Thanks!

dsuhinin: ohh, that's nice! thank you!

tests/integration/golang/mlflow/run/update_test.go
```diff
@@ -28,7 +28,7 @@ func TestUpdateRunTestSuite(t *testing.T) {
 }
 
 func (s *UpdateRunTestSuite) Test_Ok() {
-	// create test run.
+	// create test runs.
```
dsuhinin, line 31: but you create one run?

dsuhinin (approved):
Looks good!

tests/integration/golang/mlflow/run/update_test.go
fabiovincenzi: they are two

dsuhinin: aaa, understood.

PR #778, opened 2023-12-28
Prevent deletion of default namespace and default experiment

dsuhinin:
PR 1/2. What's inside the box:
- prevent deletion for `default` namespace.
- prevent deletion for `default` experiment.
- fix typos.
- updated `unit` and `integration` tests according to the new logic.


suprjinx (approved):
looks great

Issue #759, opened 2023-12-18
Metric contexts seem to be joined incorrrectly in Aim metrics view


suprjinx:
![Image](https://github.com/G-Research/fasttrackml/assets/1463452/2331a326-29bd-4d10-afc9-eefb6429772e)

The "context" column is not showing the correct context for the particular metric, maybe the join is incorrect? In the screenshot, the "foo" metric for "enchanting-shrew-747" should only have one context key, but it seems to be picking up the most recent metric which has two.

suprjinx:
okay, possibly this is working as expected -- the context, in this view, is a superset, it seems

PR #784, opened 2024-01-03
Fix needed for migrating existing data

suprjinx:
cannot add not-null column (metric.context_id) without default value
Fixes https://github.com/G-Research/fasttrackml/issues/759


pkg/api/mlflow/dao/repositories/metric.go
```diff
@@ -110,8 +110,8 @@ func (r MetricRepository) CreateBatch(
 	}
 
 	for n := range metrics {
-		metrics[n].Context = *allContexts[n]
 		metrics[n].ContextID = allContexts[n].ID
+		metrics[n].Context = *allContexts[n]
```
dave-gantenbein, line 114: Why are there no tests here?

suprjinx: the `repositories` functions are generally tested by the integration tests

dave-gantenbein: Well, this is changing functionality, but there is no corresponding test.

suprjinx: there's an extensive test for this area in `tests/integration/golang/mlflow/run/log_batch_test.go` -- i'll try to add something for this change

pkg/database/migrations/v_0009/migrate.go
```diff
@@ -12,23 +16,81 @@ const Version = "2c2299e4e061"
...
+					return eris.Wrapf(err, "error dropping %s", index)
+				}
+				if err := tx.Migrator().RenameTable(table, backupName(table)); err != nil {
+					return eris.Wrapf(err, "error renaming %s", table)
+				}
+			}
+			index := "idx_metrics_iter"
+			if err := dropIndex(tx, backupName("metrics"), index); err != nil {
+				return eris.Wrapf(err, "error dropping %s", index)
 			}
```
dsuhinin, line 33: maybe we can make `tables` as a `map` like `map["metrics"] = [metrics_indexs]` and iterate over this map instead? in that everything will be a bit cleaner and more understandable and you don't need another `if` statement
```
if err := dropIndex(tx, backupName("metrics"), index); err != nil {
 				return eris.Wrapf(err, "error dropping %s", index)
 			}
```

tests/integration/golang/mlflow/run/log_batch_test.go
```diff
@@ -397,10 +424,20 @@ func (s *LogBatchTestSuite) TestMetrics_Ok() {
...
+					lastMetric, err := s.MetricFixtures.GetLatestMetricByKey(context.Background(), key)
+					s.Require().Nil(err)
+					s.Equal(iteration, lastMetric.LastIter)
+				}
+			}
+			if len(tt.latestMetricKeyCount) > 0 {
+				for key, count := range tt.latestMetricKeyCount {
+					latestMetrics, err := s.MetricFixtures.GetLatestMetricsByKey(context.Background(), key)
+					s.Require().Nil(err)
+					s.T().Log(latestMetrics)
```
dsuhinin, line 438: probably this is debug info?

tests/integration/golang/mlflow/run/log_batch_test.go
```diff
@@ -397,10 +424,20 @@ func (s *LogBatchTestSuite) TestMetrics_Ok() {
...
-				s.Require().Nil(err)
-				s.Equal(iteration, lastMetric.LastIter)
+			if len(tt.latestMetricIteration) > 0 {
+				for key, iteration := range tt.latestMetricIteration {
+					lastMetric, err := s.MetricFixtures.GetLatestMetricByKey(context.Background(), key)
+					s.Require().Nil(err)
+					s.Equal(iteration, lastMetric.LastIter)
+				}
+			}
+			if len(tt.latestMetricKeyCount) > 0 {
```
dsuhinin, line 434: I think you even don't need such `if` statement. `range` won't even iterate if `latestMetricKeyCount` is missing.

dsuhinin (changes requested):
General comment -> this migration looks a bit dangerous(we copy data of several tables ) so maybe, again maybe it will be nice as a final step to make sure that number of records is the same between `metrics` and `metrics_old`, `latest_metrics` and `latest_metrics_old` and nothing was lost in `context` table?

pkg/database/migrations/v_0009/model.go
```diff
@@ -31,6 +31,8 @@ const (
 	LifecycleStageDeleted LifecycleStage = "deleted"
 )
 
+var DefaultContext = Context{ID: 1, Json: datatypes.JSON("{}")}
```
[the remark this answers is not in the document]
suprjinx: this is "frozen in time" whereas models is living/evolving so I think we need both.

pkg/database/migrations/v_0009/migrate.go
```diff
@@ -12,23 +16,81 @@ const Version = "2c2299e4e061"
...
+			if err := db.Create(&DefaultContext).Error; err != nil {
+				return fmt.Errorf("error creating default context: %s", err)
+			}
+		} else {
+			return fmt.Errorf("unable to find default context: %s", err)
+		}
+	}
+	log.Debugf("default metric context: %v", DefaultContext)
+	return nil
+}
```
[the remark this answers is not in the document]
suprjinx: causes cyclic import problem to go from here to `database`, but it seems weird to use method from `v_0009` in the server init. Not sure what to do other than duplicate.

pkg/database/migrations/v_0009/migrate.go
suprjinx: this case is weird because `idx_metrics_iter` only exists for `metrics`, not `latest_metrics` -- so it can't really be in the loop. but it could `map[string][]string` for table -> indexes I suppose

dsuhinin: yep yep, I talked about map of slices so you can specify own index list for each table and it will look a bit more clean.

pkg/database/migrations/v_0009/migrate.go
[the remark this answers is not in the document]
dsuhinin: ok

pkg/database/migrations/v_0009/model.go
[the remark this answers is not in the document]
dsuhinin: ok

suprjinx:
> General comment -> this migration looks a bit dangerous(we copy data of several tables ) so maybe, again maybe it will be nice as a final step to make sure that number of records is the same between `metrics` and `metrics_old`, `latest_metrics` and `latest_metrics_old` and nothing was lost in `context` table?

added a row-count verification. Context table is created new in this migration.

pkg/database/migrations/v_0009/migrate.go
suprjinx: done

tests/integration/golang/mlflow/run/log_batch_test.go
suprjinx: removed `if`s

tests/integration/golang/mlflow/run/log_batch_test.go
suprjinx: removed

pkg/database/migrations/v_0009/migrate.go
```diff
@@ -12,23 +16,95 @@ const Version = "2c2299e4e061"
...
+				// verify
+				var oldRowCount, newRowCount int64
+				if err := tx.Table(table).Count(&newRowCount).Error; err != nil {
+					return eris.Wrapf(err, "error counting rows for %s", table)
+				}
+				if err := tx.Table(backupName(table)).Count(&oldRowCount).Error; err != nil {
+					return eris.Wrapf(err, "error counting rows for %s", backupName(table))
+				}
+				if oldRowCount != newRowCount {
+					return eris.Errorf("rowcount incorrect for for %s (old: %d, new: %d)",
```
dsuhinin, line 64: 2 `for`.

dsuhinin (approved):
Looks good!

dsuhinin (approved):

PR #787, opened 2024-01-04
Integrate context filtering logic into the Aim GetRunMetrics and SearchAlignedMetrics endpoints.

dsuhinin:
Whats inside the box:
- adopted `GetRunMetrics` and `SearchAlignedMetrics` endpoints to work with `context` filtering.
- adjusted integration tests to test `context` filtering.


pkg/api/aim/runs.go
```diff
@@ -181,39 +182,57 @@ func GetRunMetrics(c *fiber.Ctx) error {
...
+	).Where(
+		"key IN ?", metricKeys,
+	).Order(
+		"iter",
+	).Find(
+		&data,
+	).Error; err != nil {
+		return fmt.Errorf("unable to find run metrics: %w", err)
+	}
+
```
dsuhinin, line 227: I decided to split, because in that case it looks more clear and it is much easy to construct queries.

pkg/api/aim/runs.go
```diff
@@ -181,39 +189,60 @@ func GetRunMetrics(c *fiber.Ctx) error {
...
+	var data []database.Metric
+	if err := database.DB.Where(
+		"run_uuid = ?", p.ID,
+	).InnerJoins(
+		"Context",
+		func() *gorm.DB {
+			query := database.DB
+			for _, context := range contexts {
+				if query.Dialector.Name() == database.SQLiteDialectorName {
+					query = query.Or("json(json) = json(?)", context)
```
suprjinx, line 221: I don't think this will do "subset" context filtering as we've done elsewhere (where database context can be a superset of the input context and still match).

dsuhinin: in of this endpoint it seems like we don't need so search by `subset`. AIM send the full `context` object which belongs to particular metric .

suprjinx (approved):

suprjinx (approved):

PR #802, opened 2024-01-11
Fix default experiment parameters for new namespaces.

dsuhinin:
What's inside the box:
- properly initialise default experiment for newly created namespaces. before fix we missing such fields: `CreationTime`, `LastUpdateTime`, `ArtifactLocation`.


pkg/api/admin/service/namespace/service.go
```diff
@@ -47,25 +59,49 @@ func (s Service) CreateNamespace(ctx context.Context, code, description string)
...
 	}
-	exp := &models.Experiment{
-		Name:           "Default",
-		LifecycleStage: models.LifecycleStageActive,
-	}
+
 	namespace := &models.Namespace{
 		Code:                code,
 		Description:         description,
-		Experiments:         []models.Experiment{*exp},
```
suprjinx, line 57: why didn't this approach work here?

dsuhinin: problem in unit tests which failed because of this approach. mainly because of access to `ID` of experiment, that's why I changed the flow.

suprjinx: but couldn't you just mock the expectation for the NamespaceRespository.Update() call? it seems like this is increasing mainline complexity to support mocking in unit test

dsuhinin: I can try to play with and around but in any case we have to use `experiment` mocks.

dsuhinin: ok, I tried to revert back to what we have before and tried to apply unit tests but without success. the main problem that in unit tests I can't normally populate the id of `default` experiment which leads to null pointer problems further. I can populate `id` in `[]models.Experiment{experiment}` but it doesn't affect to original `experiment` and ID stayed equal `nil` which leads to `nil` pointer problem.

suprjinx (approved):

PR #815, opened 2024-01-19
Improvements for `/metrics/get-history` endpoint.

dsuhinin:
What's inside the box:
 - use `Joins` instead of `Preload` to immediately load all the data and do not call database for each `context` record.
 - when we format `response` object, optimise working process with `context` object.


pkg/api/mlflow/api/response/metric.go
```diff
@@ -39,18 +39,27 @@ func NewMetricHistoryResponse(metrics []models.Metric) (*GetMetricHistoryRespons
 		Metrics: make([]MetricPartialResponse, len(metrics)),
 	}
 
+	mappedContext := map[string]map[string]interface{}{}
```
suprjinx, line 42: `map[string]map[string]any{}` might be a little bit more readable

suprjinx (approved):
LGTM

suprjinx (approved):

PR #827, opened 2024-01-24
Add helper for parsing context from the encoded metric traces

### LINKED ISSUE #777: Add "object" decoding to to `encoding.Decoder`, then update `tests/integration/golang/aim/metric/search_test.go` to use it

suprjinx:
fixes #777


tests/integration/golang/aim/metric/search_test.go
```diff
@@ -162,11 +162,7 @@ func (s *SearchMetricsTestSuite) Test_Ok() {
 		RunID:     run2.ID,
 		Iter:      3,
 		Context: models.Context{
-			Json: datatypes.JSON([]byte(`
-				{
-					"testkey": "testvalue"
-				}`,
-			)),
+			Json: datatypes.JSON([]byte(`{"testkey":"testvalue"}`)),
```
dsuhinin, line 165: we probably don't need to have `datatypes.JSON([]byte`. under the hood `datatypes.JSON` is already `[]byte` so we can use just `datatypes.JSON` without `[]byte`.

tests/integration/golang/aim/metric/search_test.go
```diff
@@ -179,11 +175,7 @@ func (s *SearchMetricsTestSuite) Test_Ok() {
 		RunID:     run2.ID,
 		LastIter:  3,
 		Context: models.Context{
-			Json: datatypes.JSON([]byte(`
-				{
-					"testkey": "testvalue"
-				}`,
-			)),
+			Json: datatypes.JSON([]byte(`{"testkey":"testvalue"}`)),
```
dsuhinin, line 178: the same probably here.

tests/integration/golang/aim/metric/search_test.go
```diff
@@ -5929,6 +5921,16 @@ func (s *SearchMetricsTestSuite) Test_Ok() {
 					timestampsKey := prefix + ".timestamps.blob"
 					valuesKey := prefix + ".values.blob"
 
+					contextPrefix := prefix + ".context"
+					contx, err := helpers.ExtractContextBytes(contextPrefix, decodedData)
+					s.Require().Nil(err)
```
dsuhinin, line 5926: I think it should be a part of our standard  decoder unfortunately and work for any object and not only for `context`.

tests/integration/golang/helpers/context.go
```diff
@@ -0,0 +1,24 @@
...
+// ExtractContextBytes finds the metric context in the decoded data, using the given key prefix.
+func ExtractContext(contextPrefix string, decodedData map[string]any) map[string]any {
+	contx := map[string]any{}
+	for key := range decodedData {
+		if strings.HasPrefix(key, contextPrefix) && len(key) > len(contextPrefix) {
+			contx[key[len(contextPrefix)+1:]] = decodedData[key]
+		}
+	}
+	return contx
+}
```
dsuhinin, line 24: Im not sure that this is what we exactly need. so:
- extraction should be a part of our `decoder`.
- extraction should work for any object and not only for context out of the box.
- `json` object could be complex object with other nested objects and extraction should work for this case as well under the hood of `decoder`.

dsuhinin (changes requested):

dsuhinin (dismissed):

tests/integration/golang/aim/metric/search_test.go
suprjinx: done

tests/integration/golang/aim/metric/search_test.go
suprjinx: done

tests/integration/golang/helpers/context.go
suprjinx: saving this improvement for another day -- possibly conversion to arrow records will happen first.

tests/integration/golang/aim/metric/search_test.go
suprjinx: done

dsuhinin (approved):

PR #845, opened 2024-01-31
Helm chart - fasttrackml

ljubon:
The goal of this PR is to package FastTrackML and allow contributors to also use it inside a Kubernetes cluster with simple installation
The scope of change includes

- [x] using existing Docker Image as a base image for k8s deployment
  - [x] migrating [.env](https://github.com/ljubon/fasttrackml/blob/main/.devcontainer/.env) file to [ConfigMap](https://github.com/ljubon/fasttrackml/blob/3bfd8bb82d37dce56ec3f8ac95bfea76a99c16e5/helm/fasttrackml/values.yaml#L14) with exposing `data` section in default [values.yaml](https://github.com/ljubon/fasttrackml/blob/3bfd8bb82d37dce56ec3f8ac95bfea76a99c16e5/helm/fasttrackml/values.yaml#L14) file
- [x] publishing the helm chart to main G-Research/charts with the rest of the charts
  - [x] __NOTE__: For this action org owner will need to create secrets and follow process to onboard `fasttrackml` to this publishing  mehanism
  - [x] PR for adding `fastrackml` charts to the G-Research/config https://github.com/G-Research/charts/pull/130

#### Testing helm chart

```bash
git checkout oss-350-helm-chart-lvl-2
kind create cluster --name fastrackml
kubectl cluster-info --context kind-fastrackml
helm install fasttrackml ./helm/fasttrackml
kubectl port-forward deployment/fasttrackml 5000:5000

# Cleanup command
kind delete cluster --name fastrackml
```

In your local browser at http://localhost:5000 service will be exposed

Resolves: https://github.com/G-Research/oss-portfolio-maturity/issues/350

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -121,7 +121,7 @@ jobs:
 
   chart-release:
     name: Publish Helm chart
-    if: startsWith(github.ref, 'refs/tags/v') || github.ref == 'refs/heads/main'
+    if: startsWith(github.ref, 'refs/tags/v') 
     uses: G-Research/charts/.github/workflows/invoke-push.yaml@master
     secrets:
       APP_ID: ${{ secrets.APP_ID }}
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -125,4 +125,4 @@ jobs:
     uses: G-Research/charts/.github/workflows/invoke-push.yaml@master
     secrets:
       APP_ID: ${{ secrets.APP_ID }}
-      APP_PRIVATE_KEY: ${{ secrets.GH_APP_PEM }}
+      APP_PRIVATE_KEY: ${{ secrets.APP_PRIVATE_KEY }}
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -118,3 +118,11 @@ jobs:
             $(printf -- "--tag ${{ vars.DOCKER_REPO }}:%s " ${tags[@]}) \
             $(printf "${{ vars.DOCKER_REPO }}@%s " ${digests[@]})
           echo "::endgroup::"
+
+  chart-release:
+    name: Publish Helm chart
+    if: startsWith(github.ref, 'refs/tags/v') || github.ref == 'refs/heads/main'
+    uses: G-Research/charts/.github/workflows/invoke-push.yaml@master
+    secrets:
+      APP_ID: ${{ secrets.APP_ID }}
+      APP_PRIVATE_KEY: ${{ secrets.GH_APP_PEM }}
--- a/helm/fasttrackml/values.yaml
+++ b/helm/fasttrackml/values.yaml
@@ -17,7 +17,8 @@ data:
 
   # common service settings.
   FML_DEV_MODE: "true"
-  FML_DATABASE_URI: "postgresql://admin:admin@localhost:5432" # /data/mydatabase.db
+  FML_DATABASE_URI: "postgresql://admin:admin@localhost:5432" # PostgresSQL
+  # FML_DATABASE_URI: "/data/mydatabase.db # SQLite 
   FML_LOG_LEVEL: "DEBUG"
 
   # database settings.
@@ -81,10 +82,6 @@ autoscaling:
   targetCPUUtilizationPercentage: 80
   # targetMemoryUtilizationPercentage: 80
   
-# For mounting SQLite database, set to `true` and uncoment `volumes` and `volumeMounts`
-sqlite:
-  enabled: false
-
 # Additional volumes on the output Deployment definition.
 volumes: []
   # - name: sqlite-db
--- a/helm/fasttrackml/values.yaml
+++ b/helm/fasttrackml/values.yaml
@@ -1,5 +1,6 @@
 ---
 replicaCount: 1
+namespace: default
 
 image:
   repository: gresearch/fasttrackml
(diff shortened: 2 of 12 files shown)


helm/fasttrackml/values.yaml
```diff
@@ -0,0 +1,101 @@
...
+  # Overrides the image tag whose default is the chart appVersion.
+  tag: "latest"
+
+imagePullSecrets: []
+nameOverride: ""
+fullnameOverride: ""
+
+data:
+  # common service settings.
+  FML_DEV_MODE: true
```
suprjinx, line 16: We should add `FML_DATABASE_URI` as an environment var -- with examples for sqlite and postgres

helm/fasttrackml/values.yaml
```diff
@@ -0,0 +1,101 @@
...
+  FML_LOG_LEVEL: DEBUG
+
+  # database settings.
+  LC_COLLATE: POSIX
+  POSTGRES_USER: postgres
```
suprjinx, line 21: POSTGRES settings are not really part of this service. We put user/pwd/host/db in the FML_DATABASE_URI for postgres.

suprjinx:
I get an error installing the chart: ```gwilson@macbook ~/P/fasttrackml (pr/845)> helm install fasttrackml ./helm/fasttrackml
Error: INSTALLATION FAILED: ConfigMap in version "v1" cannot be handled as a ConfigMap: json: cannot unmarshal bool into Go struct field ConfigMap.data of type string```

helm/fasttrackml/values.yaml
```diff
@@ -0,0 +1,103 @@
...
+
+autoscaling:
+  enabled: false
+  minReplicas: 1
+  maxReplicas: 100
+  targetCPUUtilizationPercentage: 80
+  # targetMemoryUtilizationPercentage: 80
+  
+# For mounting SQLite database, set to `true` and uncoment `volumes` and `volumeMounts`
+sqlite:
```
suprjinx, line 85: I don't think this sqlite.enabled switch is really needed -- in the app, sqlite is enabled by using a sqlite uri in the FML_DATABASE_URI field. Maybe you could just put a commented-out sqlite uri example next to the postgres URI, which points to a file on the volume mount?

suprjinx: (that is, points to a file on this commented-out volume mount)

suprjinx:
> > I get an error installing the chart: `gwilson@macbook ~/P/fasttrackml (pr/845)> helm install fasttrackml ./helm/fasttrackml Error: INSTALLATION FAILED: ConfigMap in version "v1" cannot be handled as a ConfigMap: json: cannot unmarshal bool into Go struct field ConfigMap.data of type string`
> 
> Hi @suprjinx, thank you for your time and review 🙏 The above comments are resolved, let me know if the new changes are as you requested
> 
> Regarding the error you encountered, it was specific to templating boolean values of `FML_DEV_MODE`. The fix is in the latest commit with an additional note to keep in mind to quote all values passed under the `data:` section in `values.yaml`

Thank you @ljubon for the contribution! I had one more suggestion on how to simplify the values.yaml

helm/fasttrackml/values.yaml
```diff
@@ -0,0 +1,100 @@
...
+nameOverride: ""
+fullnameOverride: ""
+
+data:
+  # NOTE: Make sure you qoute values for all keys
+
+  # common service settings.
+  FML_DEV_MODE: "true"
+  FML_DATABASE_URI: "postgresql://admin:admin@localhost:5432" # PostgresSQL
+  # FML_DATABASE_URI: "/data/mydatabase.db # SQLite
```
suprjinx, line 21: this would take uri of `"sqlite:///data/mydatabase.db"`

suprjinx (approved):
This looks great, thanks for making those changes! Do you think we need to add ingress? (can be new ticket)

suprjinx (approved):

jgiannuzzi:
@ljubon Secrets created under the `release` environment. The key fingerprint to add to [`config.json`](https://github.com/G-Research/charts/blob/master/.github/workflows/scripts/config.json) is `Da10MkM0RnFLWCcBbZ8ij+lYntgMDGSBdhljq+Oz5HY=`.

.github/workflows/release.yml
```diff
@@ -118,3 +118,11 @@ jobs:
             $(printf -- "--tag ${{ vars.DOCKER_REPO }}:%s " ${tags[@]}) \
             $(printf "${{ vars.DOCKER_REPO }}@%s " ${digests[@]})
           echo "::endgroup::"
+
+  chart-release:
+    name: Publish Helm chart
+    if: startsWith(github.ref, 'refs/tags/v') || github.ref == 'refs/heads/main'
```
jgiannuzzi, line 124: we only want to publish charts on tag push, not on push to `main`!

.github/workflows/release.yml
```diff
@@ -118,3 +118,11 @@ jobs:
             $(printf -- "--tag ${{ vars.DOCKER_REPO }}:%s " ${tags[@]}) \
             $(printf "${{ vars.DOCKER_REPO }}@%s " ${digests[@]})
           echo "::endgroup::"
+
+  chart-release:
```
jgiannuzzi, line 122: you'll need to add `environment: release` for this to work now that the secrets have been created in this environment.

jgiannuzzi (changes requested):

.github/workflows/release.yml
jgiannuzzi: excellent point, done!

The new key fingerprint is `dU/0iLRxlQu5Apyx8y16eVPqhsClH2BMCQiMdEEcyBU=`.

.github/workflows/release.yml
```diff
@@ -118,3 +118,11 @@ jobs:
...
+    name: Publish Helm chart
+    if: startsWith(github.ref, 'refs/tags/v') 
+    uses: G-Research/charts/.github/workflows/invoke-push.yaml@master
+    secrets:
+      APP_ID: ${{ secrets.APP_ID }}
+      APP_PRIVATE_KEY: ${{ secrets.GH_APP_PEM }}
```
jgiannuzzi, line 128:
```suggestion
      APP_PRIVATE_KEY: ${{ secrets.APP_PRIVATE_KEY }}
```

jgiannuzzi (dismissed):

jgiannuzzi (approved):
Thank you so much for this contribution @ljubon!

PR #846, opened 2024-02-01
Add ability to export runs in CSV format

dsuhinin:


pkg/api/aim/helpers.go
```diff
@@ -0,0 +1,271 @@
...
+	"golang.org/x/exp/slices"
+
+	"github.com/G-Research/fasttrackml/pkg/api/aim/encoding"
+	"github.com/G-Research/fasttrackml/pkg/database"
+)
+
+// FormatRunsSearchResponseAsCSV formats and sends Runs search response as a CSV file.
+//
+//nolint:gocyclo
+func FormatRunsSearchResponseAsCSV(ctx *fiber.Ctx, runs []database.Run, excludeTraces, excludeParams bool) {
```
suprjinx, line 23: maybe these two functions could have shorter names? "RunsAsCsv", "RunsAsStream" for instance?

dsuhinin: maybe these two functions could have shorter names? "RunsAsCsv", "RunsAsStream" for instance? - yyyep, Ill make it a bit shorter.

dsuhinin:
```
also, I wonder if they might be better as methods of a new struct like response.SearchRuns 
```
- Im not sure about that, since `response.SearchRuns` is from `mlflow` but in general it should be a bit different I agree. Let's think about that when we will rewrite `AIM`.

suprjinx: we do have other `response` structs in Aim side already, such as `response.GetRunMetrics`, etc, so I don't think we need to wait for big rewrite

dsuhinin: the strange moment here, that we don't need a `struct`, because there is no real `response` object for such endpoint.

tests/integration/golang/aim/run/search_test.go
```diff
@@ -25,7 +31,303 @@ func TestSearchTestSuite(t *testing.T) {
 	suite.Run(t, new(SearchTestSuite))
 }
 
-func (s *SearchTestSuite) Test_Ok() {
+func (s *SearchTestSuite) TestCSVReport_Ok() {
+	// create test experiments.
+	experiment, err := s.ExperimentFixtures.CreateExperiment(context.Background(), &models.Experiment{
```
suprjinx, line 36: could all this setup be shared with `TestStreamData_Ok` by using setup method?

dsuhinin: ok, should be fine now.

suprjinx (approved):

PR #871, opened 2024-02-07
Use JSONB instead of JSON for context object in SQLite.

dsuhinin:


pkg/database/migrations/v_0010/model.go
```diff
@@ -0,0 +1,315 @@
...
+	Step      int64  `gorm:"not null"`
+	IsNan     bool   `gorm:"not null"`
+	RunID     string `gorm:"column:run_uuid;not null;primaryKey;index"`
+	LastIter  int64
+	ContextID uint `gorm:"not null;primaryKey"`
+	Context   Context
+}
+
+// JSONB defined JSONB data type, need to implements driver.Valuer, sql.Scanner interface
+type JSONB json.RawMessage
```
suprjinx, line 154: can we move all the JSONB declarations to some other location that's not duplicated so much? Eg, `migrations/types.go` or the like?

suprjinx (approved):
Looks good

PR #874, opened 2024-02-07
MlFlow migration documentation

### LINKED ISSUE #288: Migration , limitations and differences
Jose Alatorre from Main Sequence,
Hi, I tried to run FastTrackML against our current DB(running on Postgres) of MLFLOW but it was unsuccessful. A few initialization arguments seem incompatible, like backend-store-uri, default-artifact-root, and so on. I couldn't find in the documentation any indication regarding migrations or compatibility. Will be nice if the documentation is a little more extensive so more users can try this implementation with their current deployment. Bests!

fabiovincenzi:
Fixes #288


suprjinx (approved):
Excellent doc :-)

PR #885, opened 2024-02-08
Extend `/project/params` endpoint to be filtered down to selected experiments.

dsuhinin:
What's inside the box:
- adjust `GET /project/params` endpoint to filter data by provided `experiments`.
- adjust `integration` tests.


pkg/api/aim/projects.go
```diff
@@ -117,8 +118,8 @@ func GetProjectParams(c *fiber.Ctx) error {
 	resp := fiber.Map{}
 
 	if !q.ExcludeParams {
-		var paramKeys []string
-		if tx := database.DB.Distinct().Model(
+		// process params.
```
suprjinx, line 121: "process params" seems like we'd be changing them, this is just fetching, yes?

pkg/api/aim/projects.go
```diff
@@ -140,8 +144,8 @@ func GetProjectParams(c *fiber.Ctx) error {
 			}
 		}
 
-		var tagKeys []string
-		if tx := database.DB.Distinct().Model(
+		// process tags.
```
suprjinx, line 147: as above

suprjinx (approved):

PR #892, opened 2024-02-12
Prevent deletion of default experiment in non-default namespaces.

dsuhinin:
What's inside the box:
- change strategy of `default` experiment checking.


suprjinx (approved):
lgtm

PR #897, opened 2024-02-13
Fix fasttrackml python client

### LINKED ISSUE #898: Fix fasttrackml python client
The `fasttrackml` python client is not created correctly, not all the packages are included in the wheel.

fabiovincenzi:
Fixes #898.
Not all packages were included in the python wheel, this PR adds `__init__.py `files in all the packages.


jgiannuzzi (commented):
could you please also:
- ensure that LICENSE and README are again properly included in the wheel
- remove `batch.py` (after validating that the functionality has been properly included in the wheel)
- remove the duplicated `setup.py`

python/setup.py
```diff
@@ -4,13 +4,16 @@
...
 
-install_requires=[
-    "mlflow",
-],
+install_requires = (
+    [
+        "mlflow",
+    ],
+)
+
```
jgiannuzzi, line 12: I don't think this is necessary. It doesn't seem to be used anywhere.

python/pyproject.toml
jgiannuzzi, line 1: there is now a duplicate `pyproject.toml` — could you please delete the one under `python/fasttrackml`?

jgiannuzzi (commented):
I had never looked at the previous python changes, and I have noticed that they haven't been properly formatted. It seems to be because of the `fasttrackml` pattern in the `.gitignore` file. Could you please remove it from the `.gitignore` file (along with the `!python/fasttrackml/` pattern and re-run `make python-format`?

jgiannuzzi:
one more thing: the python library files should include each other as `.`.

e.g. this import is wrong:
```python
from fasttrackml.entities.metric import Metric
```

this one is correct:
```python
from .entities.metric import Metric
```

jgiannuzzi (approved):

PR #899, opened 2024-02-13
Fix GetRunMetrics endpoint behaviour.

dsuhinin:
What's inside the box:
- fix for `GetRunMetrics` endpoint -> now endpoint should return only metrics which were requested.

### DIFF
--- a/pkg/api/aim/runs.go
+++ b/pkg/api/aim/runs.go
@@ -163,8 +163,8 @@ func GetRunMetrics(c *fiber.Ctx) error {
 	}
 
 	var b []struct {
-		Name    string          `json:"name"`
-		Context json.RawMessage `json:"context"`
+		Name    string    `json:"name"`
+		Context fiber.Map `json:"context"`
 	}
 
 	if err := c.BodyParser(&b); err != nil {
@@ -181,9 +181,13 @@ func GetRunMetrics(c *fiber.Ctx) error {
 	metricKeysMap := make(map[metric]any, len(b))
 	for _, m := range b {
 		if m.Context != nil {
+			serializedContext, err := json.Marshal(m.Context)
+			if err != nil {
+				return fiber.NewError(fiber.StatusUnprocessableEntity, err.Error())
+			}
 			metricKeysMap[metric{
 				name:    m.Name,
-				context: string(m.Context),
+				context: string(serializedContext),
 			}] = nil
 		}
 	}


pkg/api/aim/runs.go
```diff
@@ -171,16 +171,15 @@ func GetRunMetrics(c *fiber.Ctx) error {
...
-	metricKeysMap, contexts := make(fiber.Map, len(b)), make([]types.JSONB, 0, len(b))
+	metricKeysMap := make(map[string]types.JSONB, len(b))
 	for _, m := range b {
 		if m.Context != nil {
 			serializedContext, err := json.Marshal(m.Context)
 			if err != nil {
 				return fiber.NewError(fiber.StatusUnprocessableEntity, err.Error())
 			}
-			contexts = append(contexts, serializedContext)
+			metricKeysMap[m.Name] = serializedContext
```
jgiannuzzi, line 181: it looks to me like requesting 2 metrics with the same key and a different context won't work anymore with this code.

e.g. the following request would result in a `metricKeysMap` with only one entry:
```json
[
    {
        "context": {
            "mode": "validation"
        },
        "last_value": 0.1,
        "name": "metric1"
    },
    {
        "context": {
            "mode": "train"
        },
        "last_value": 0.1,
        "name": "metric1"
    }
]
```

pkg/api/aim/runs.go
```diff
@@ -209,26 +208,22 @@ func GetRunMetrics(c *fiber.Ctx) error {
...
 		"Context",
-		func() *gorm.DB {
-			query := database.DB
-			for _, context := range contexts {
-				query = query.Or("json = ?", context)
-			}
-			return query
-		}(),
-	).Where(
-		"key IN ?", metricKeys,
```
jgiannuzzi, line 226: it looks like `metricKeys` is not used anymore, so there is no point in building it earlier in this function

pkg/api/aim/runs.go
suprjinx: could use `models.Metric.UniqueKey()` if they always belong to the same run

jgiannuzzi (changes requested):
So the whole reason we were using `map[string]interface{}` to track the metric keys was as a "set" aka a list of unique strings. This is because the UI sometimes request the same metric key multiple times.

Now we have to deal with a request for key+context tuples, and make sure they are unique. However this approach with a `map[string][]types.JSONB` will not filter duplicated key+context tuples.

pkg/api/aim/runs.go
```diff
@@ -171,23 +171,20 @@ func GetRunMetrics(c *fiber.Ctx) error {
 		return fiber.NewError(fiber.StatusUnprocessableEntity, err.Error())
 	}
 
-	metricKeysMap, contexts := make(fiber.Map, len(b)), make([]types.JSONB, 0, len(b))
+	metricKeysMap := make(map[string]map[string]types.JSONB, len(b))
```
jgiannuzzi, line 174: This has become a really weird structure, especially with this new GetHash function. We receive a list of key ("name") and context tuples from the client, we can just use that as the key and `nil` as the value so we have a proper unordered unique list of key and context tuples.

We can then build the subquery with these tuples.

Finally we can create another map of these tuples in which we store a list of iters and a list of values **and nothing else**. Duplicating the key/name and context in there does not make sense! Once this map is built, we can return the proper structure to the client.

jgiannuzzi: could we please use actual structs instead of concatenated strings? merging and splitting with underscores could introduce all kinds of bugs!

pkg/api/aim/runs.go
```diff
@@ -171,23 +171,25 @@ func GetRunMetrics(c *fiber.Ctx) error {
...
+	type metric struct {
+		name    string
+		context string
+	}
+
+	// collect unique metrics. uniqueness provides metricKeysMap + metric struct.
+	metricKeysMap := make(map[metric]any, len(b))
 	for _, m := range b {
 		if m.Context != nil {
 			serializedContext, err := json.Marshal(m.Context)
```
jgiannuzzi, line 184: should we maybe make `b.Context` a `json.RawMessage` instead so we don't need to re-serialise it here?

jgiannuzzi (approved):
LGTM, just one optional remark remaining!

jgiannuzzi (approved):

PR #901, opened 2024-02-13
Refactor AIM `apps` endpoints

dsuhinin:
What's inside the box:
- refactored `apps` entity according to overal project standarts.

### DIFF
--- a/pkg/api/aim2/api/response/app.go
+++ b/pkg/api/aim2/api/response/app.go
@@ -24,13 +24,14 @@ type AppState map[string]any
 func NewGetAppsResponse(apps []models.App) []App {
 	resp := make([]App, len(apps))
 	for i, app := range apps {
-		resp[i] = NewCreateAppResponse(app)
+		//nolint:gosec
+		resp[i] = NewCreateAppResponse(&app)
 	}
 	return resp
 }
 
 // NewCreateAppResponse creates new response object for `POST /apps` endpoint.
-func NewCreateAppResponse(app models.App) App {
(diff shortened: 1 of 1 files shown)


pkg/api/aim2/service/app/service.go
```diff
@@ -0,0 +1,95 @@
+package app
+
+import (
+	"context"
+
+	"github.com/G-Research/fasttrackml/pkg/api/aim2/api/request"
+	"github.com/G-Research/fasttrackml/pkg/api/aim2/dao/convertors"
+	"github.com/G-Research/fasttrackml/pkg/api/aim2/dao/repositories"
+	"github.com/G-Research/fasttrackml/pkg/api/mlflow/dao/models"
```
suprjinx, line 9: we should use `aim2/dao/models`, yes?

pkg/api/aim2/service/app/service.go
```diff
@@ -0,0 +1,95 @@
...
+	app.State = database.AppState(req.State)
+
+	if err := s.appRepository.Update(ctx, app); err != nil {
+		return nil, api.NewInternalError("unable to update app '%s': %s", app.ID, err)
+	}
+	return app, nil
+}
+
+// GetApps returns the list of active apps.
+func (s Service) GetApps(ctx context.Context, namespace *models.Namespace) ([]database.App, error) {
```
suprjinx, line 73: I don't think we should return `database` package here. I'd think we reserve `database` module for migrations, and use `api2/dao/models` in the app logic.

suprjinx (approved):

PR #922, opened 2024-02-19
AIM `runs` endpoints refactoring.

dsuhinin:
What's inside the box:
 - `runs` endpoints refactoring.
 - improvements and adjustments.


pkg/api/aim2/dao/dto/metric.go
```diff
@@ -0,0 +1,10 @@
+package dto
```
suprjinx, line 1: Can this just be part of models.Metric? I don't think we should add another package type

pkg/api/aim2/service/run/converters.go
```diff
@@ -0,0 +1,30 @@
...
+
+	"github.com/rotisserie/eris"
+
+	"github.com/G-Research/fasttrackml/pkg/api/aim2/api/request"
+	"github.com/G-Research/fasttrackml/pkg/api/aim2/dao/dto"
+)
+
+// ConvertRunMetricsRequestToMetricKeysMapDTO converts request of `GET /runs/:id/metric/get-batch`
+// endpoint to internal DTO object.
+func ConvertRunMetricsRequestToMetricKeysMapDTO(req *request.GetRunMetricsRequest) (dto.MetricKeysMapDTO, error) {
```
suprjinx, line 14: this function name is kind a mouthful -- i think the param and return types don't really need to be in the name since they are in the signature

pkg/api/aim2/dao/dto/metric.go
suprjinx: it seems like this is unique key rather than a transfer object, so i think the package name is not helping to explain what this is. why not have models.MetricKey?

suprjinx (approved):

PR #963, opened 2024-03-01
Cover `migrations` with integration tests.

dsuhinin:
What's inside the box:
- `migrations` now covered with integration tests and it means if we do something wrong, we will see anomaly, that migration took more time then this allowed.


tests/integration/golang/migrations/migrations_test.go
```diff
@@ -0,0 +1,207 @@
...
+	"github.com/G-Research/fasttrackml/pkg/database"
+	"github.com/G-Research/fasttrackml/tests/integration/golang/helpers"
+)
+
+type MigrationsTestSuite struct {
+	suite.Suite
+}
+
+func TestMigrationsTestSuite(t *testing.T) {
+	t.Skip()
```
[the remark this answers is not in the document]
suprjinx: perhaps use build tag so we can run with argument but otherwise won't?

pkg/database/migrations/v_0006/migrate.go
```diff
@@ -41,3 +54,56 @@ func Migrate(db *gorm.DB) error {
...
+	}
+
+	if err := tx.Exec(`ALTER TABLE experiments ALTER COLUMN namespace_id DROP DEFAULT`).Error; err != nil {
+		return eris.Wrapf(err, "error dropping default value for `namespace_id` field in `experiments` table")
+	}
+
+	return nil
+}
+
+// createDefaultMetricContext creates the default metric context if it doesn't exist.
```
suprjinx, line 90: needs updated doc string

tests/integration/golang/migrations/migrations_test.go
```diff
@@ -0,0 +1,210 @@
...
+					Status:         "RUNNING",
+					ID:             uuid.New().String(),
+					SourceType:     "JOB",
+					LifecycleStage: "active",
+					ExperimentID:   experiment.ExperimentID,
+				}
+				s.Require().Nil(db.Create(&run).Error)
+
+				metrics := make([]Metric, 1_000_000)
+				latestMetrics := make([]LatestMetric, 1_000_000)
```
suprjinx, line 108: i think in real-world LatestMetrics is only a fraction of the size of Metrics

tests/integration/golang/migrations/migrations_test.go
```diff
@@ -0,0 +1,209 @@
...
+	t.Skip()
+	suite.Run(t, new(MigrationsTestSuite))
+}
+
+func (s *MigrationsTestSuite) TestMigrations() {
+	type Experiment struct {
+		Name           string `gorm:"type:varchar(256);not null;unique"`
+		ExperimentID   int32  `gorm:"column:experiment_id;not null;primaryKey"`
+		LifecycleStage string `gorm:"type:varchar(32);check:lifecycle_stage IN ('active', 'deleted')"`
+	}
```
suprjinx, line 34: why not use existing model definitions?

suprjinx: (could use `Omit` or `Select` to limit the fields, so current models works with older tables -- I think)

tests/integration/golang/migrations/migrations_test.go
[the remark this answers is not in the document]
suprjinx: I think this requires changing the source code to run it, yes? and then also it will run with all the others in that case, as w/ build tag. But with build tag, you can activate with just `go test` arguments, for CI.

Makefile
```diff
@@ -133,17 +133,23 @@ python-lint: python-env ## check python code formatting.
...
 	@echo ">>> Running unit tests."
 	@go test -tags="$(GO_BUILDTAGS)" ./pkg/...
 
-.PHONY: test-go-integration
-test-go-integration: ## run go integration tests.
+.PHONY: test-pipeline-go-integration
+test-pipeline-go-integration: ## run pipeline go integration tests.
 	@echo ">>> Running integration tests."
-	@go test -tags="$(GO_BUILDTAGS)" ./tests/integration/golang/...
+	@go test -tags="$(GO_BUILDTAGS),pipeline" ./tests/integration/golang/...
```
suprjinx, line 146: we should probably add both tags to the vscode launch config

suprjinx (approved):

suprjinx (approved):

PR #968, opened 2024-03-04
Fix issue in -> https://github.com/G-Research/fasttrackml/pull/963 PR.

dsuhinin:
What's inside the box:
- fix the issue in #965


fabiovincenzi (approved):
Looks good!

PR #1013, opened 2024-03-12
Adapt async approach from mlflow client

### LINKED ISSUE #976: Add async capability to python client
Ben Davidson:
Also would it be possible to add the ability to do things asnychronously, in the vanilla mlflow api there is the synchronous arg, which I would like to have for logging (batched)metrics

suprjinx:
Fixes #976


fabiovincenzi (approved):
Looks great!

Issue #1047, opened 2024-03-25
Error selecting metrics with context in 0.6 candidate


suprjinx:
```
time="2024-03-25T16:44:48Z" level=error msg="SQL error" func="github.com/G-Research/fasttrackml/pkg/database.(*loggerAdaptor).Trace" file="/workspaces/fasttrackml/pkg/database/logger.go:75" app_file="/workspaces/fasttrackml/pkg/api/aim2/dao/repositories/metric.go:249" app_func="github.com/G-Research/fasttrackml/pkg/api/aim2/dao/repositories.MetricRepository.GetContextListByContextObjects()" elapsed=0.512ms error="ERROR: invalid input syntax for type json (SQLSTATE 22P02)" rows=0 sql="SELECT * FROM \"contexts\" WHERE contexts.json = '{\"key1\": \"value2\"}, {\"key3\": \"value3\"}'"
time="2024-03-25T16:44:48Z" level=error msg="Error encountered in GET /aim/api/runs/search/metric: INTERNAL_ERROR: error searching runs: error finding context ids: error getting context list: error getting contexts information: ERROR: invalid input syntax for type json (SQLSTATE 22P02)" func=github.com/G-Research/fasttrackml/pkg/api/aim2.ErrorHandler file="/workspaces/fasttrackml/pkg/api/aim2/errors.go:40"
time="2024-03-25T16:44:48Z" level=info msg="500 -   75.439762ms GET /aim/api/runs/search/metric"
```

dave-gantenbein:
Fixed by #1053

PR #1048, opened 2024-03-25
The search-metrics request from ui is decomposing context to multiple objects

suprjinx:
This is a simple backend normalization.

#1047


suprjinx:
not needed

PR #1060, opened 2024-03-28
Handle correctly not found errors in `admin` and `chooser` parts.

dsuhinin:
What's inside the box:
- now we have `not-found` page for `chooser` and `admin` parts, where we can redirect users in case of some errors. This is how `not-found` page looks now:
Screenshot 2024-03-28 at 14 46 10

### DIFF
--- a/pkg/ui/chooser/routes.go
+++ b/pkg/ui/chooser/routes.go
@@ -2,10 +2,11 @@ package chooser
 
 import (
 	"embed"
-	"github.com/gofiber/fiber/v2/middleware/etag"
-	"github.com/gofiber/fiber/v2/middleware/filesystem"
 	"io/fs"
 	"net/http"
+
+	"github.com/gofiber/fiber/v2/middleware/etag"
+	"github.com/gofiber/fiber/v2/middleware/filesystem"
 
 	"github.com/gofiber/fiber/v2"
 	"github.com/gofiber/template/html/v2"


pkg/ui/admin/embed/layouts/main.html
```diff
@@ -6,9 +6,9 @@
   <meta name="viewport" content="width=device-width, initial-scale=1.0">
   <title>FastTrackML</title>
   <link rel="stylesheet" href="/static/aim/assets/icomoon/icomoonIcons.css">
-  <link rel="stylesheet" href="/static/chooser/css/simple.min.css">
+  <link rel="stylesheet" href="/static/chooser/static/css/simple.min.css">
   <link rel="stylesheet" href="/admin/static/css/namespaces.css">
-  <link rel="icon" type="image/x-icon" href="/static/chooser/favicon.ico">
+  <link rel="icon" type="image/x-icon" href="/static/chooser/static/favicon.ico">
```
suprjinx, line 11: `/static/chooser/static` seems like one-too-many `static` directories.

pkg/ui/admin/middleware/authorization.go
```diff
@@ -16,11 +16,7 @@ func NewAdminUserMiddleware(userPermissions *models.UserPermissions) fiber.Handl
 			strings.Replace(ctx.Get(fiber.HeaderAuthorization), "Basic ", "", 1),
 		)
 		if authToken == nil || !authToken.HasAdminAccess() {
-			return ctx.Status(
-				http.StatusNotFound,
-			).SendString(
-				"unable to find requested resource",
-			)
+			return ctx.Redirect("/errors/not-found", http.StatusMovedPermanently)
```
suprjinx, line 19: 👍

pkg/ui/chooser/embed/layouts/main.html
```diff
@@ -0,0 +1,82 @@
...
+            <picture>
+                <source media="(prefers-color-scheme: dark)" srcset="/static/chooser/static/media/logo-dark.svg">
+                <img class="logo" src="/static/chooser/static/media/logo-light.svg" alt="FastTrackML logo">
+            </picture>
+            <p>A <i>very fast</i> experiment tracker</p>
+        </header>
+
+        <!-- We load the UI with JS to avoid issues when an URL with embedded credentials is used -->
+        <main>
+            {{ embed }}
```
suprjinx, line 73: possibly another placeholder name would be better (eg `content`) since `embed` also refers to the filesystem

suprjinx (approved):

pkg/ui/admin/embed/layouts/main.html
suprjinx: Well at least the sources do not have `static/../static` path, but it seems like there'd be a way to avoid this path.

suprjinx (approved):
LGTM

Issue #1053, opened 2024-03-26
Error when too many metrics selected


suprjinx:
Error is "431 Request Header Fields Too Large", which isn't handled gracefully by the UI. Should perhaps convert to POST data.

PR #1075, opened 2024-04-03
Change SearchMetrics endpoint to a POST

fabiovincenzi:
Fixes https://github.com/G-Research/fasttrackml/issues/1053.
Change the `SearchMetrics` endpoint from a `GET` to a `POST`

### DIFF
--- a/pkg/api/aim2/api/request/run.go
+++ b/pkg/api/aim2/api/request/run.go
@@ -58,11 +58,11 @@ type MetricTuple struct {
 // SearchMetricsRequest is a request struct for `GET /runs/search/metric` endpoint.
 type SearchMetricsRequest struct {
 	BaseSearchRequest
-	Metrics    []MetricTuple `query:"m"`
-	Query      string        `query:"q"`
-	Steps      int           `query:"p"`
-	XAxis      string        `query:"x_axis"`
-	SkipSystem bool          `query:"skip_system"`
+	Metrics    []MetricTuple `json:"m"`
+	Query      string        `json:"q"`
+	Steps      int           `json:"p"`
+	XAxis      string        `json:"x_axis"`
+	SkipSystem bool          `json:"skip_system"`
 }
 
 // SearchAlignedMetricsRequest is a request struct for `GET /runs/search/metric/align` endpoint.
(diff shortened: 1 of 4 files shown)


pkg/api/aim2/api/request/run.go
```diff
@@ -49,20 +49,14 @@ type SearchRunsRequest struct {
...
-
 // SearchMetricsRequest is a request struct for `GET /runs/search/metric` endpoint.
 type SearchMetricsRequest struct {
 	BaseSearchRequest
-	Metrics    []MetricTuple `query:"m"`
-	Query      string        `query:"q"`
-	Steps      int           `query:"p"`
-	XAxis      string        `query:"x_axis"`
-	SkipSystem bool          `query:"skip_system"`
+	Metrics    [][]string `json:"metrics"`
```
suprjinx, line 55: Possibly we should keep `MetricTuple` as a synonym for `[]string`, so this struct field becomes `[]MetricTuple`?

pkg/api/aim2/dao/repositories/metric.go
```diff
@@ -256,7 +256,7 @@ func (r MetricRepository) findContextIDs(ctx context.Context, req *request.Searc
 	contextList := []types.JSONB{}
 	contextsMap := map[string]types.JSONB{}
 	for _, r := range req.Metrics {
-		data := types.JSONB(r.Context)
```
suprjinx, line 259: maybe use `tuple` as range var name to match the usage above?

tests/integration/golang/aim/metric/search_test.go
```diff
@@ -315,14 +316,14 @@ func (s *SearchMetricsTestSuite) Test_Ok() {
 		{
 			name: "SearchMetric",
 			request: request.SearchMetricsRequest{
-				Metrics: []request.MetricTuple{
+				Metrics: [][]string{
 					{
-						Key:     "TestMetric1",
-						Context: "{}",
+						"TestMetric1",
```
suprjinx, line 321: This change makes me think the old `Key` and `Context` were clearer than `[]string` for describing a metric -- is it worth retaining that JSON format when the request is generated in the UI?

suprjinx (approved):

PR #1079, opened 2024-04-04
Migration for OIDC v1.

dsuhinin:
Whats inside the box:
 - migration for OIDC v1 implementation.


pkg/database/migrate.go
```diff
@@ -129,7 +129,8 @@ func CheckAndMigrateDB(migrate bool, db *gorm.DB) error {
 			log.Info("Initializing database")
 			tx := db.Begin()
 			if err := tx.AutoMigrate(
-				&Namespace{},
```
jgiannuzzi, line 132: oops

pkg/api/mlflow/dao/models/relation.go
```diff
@@ -0,0 +1,19 @@
...
+
+import (
+	"time"
+
+	"github.com/google/uuid"
+)
+
+// RoleNamespace represents model to work with `role_relations` table.
+// Model holds relations between Role and Namespace models.
+type RoleNamespace struct {
```
jgiannuzzi, line 11: we should probably embed `Base` here to avoid duplicating `ID`, `CreatedAt`, and `UpdatedAt`. Probably `IsArchived` is not desirable in this case and be moved into a `ArchivableBase` struct.

jgiannuzzi (commented):
looks good generally 👍
There is just a typo in the migration (has to be fixed!), and then a nitpick about DRY (take it or leave it)

pkg/api/mlflow/dao/models/relation.go
suprjinx: i know we have the plan to redo the schema at some point -- at that point we can probably arrive at a uniform base model.

pkg/database/model.go
```diff
@@ -266,3 +266,20 @@ func NewUUID() string {
 	hex.Encode(r[:], u[:])
 	return string(r[:])
 }
+
+type Role struct {
+	ID        uuid.UUID `gorm:"type:uuid;primaryKey"`
```
suprjinx, line 271: could this embed `Base` struct here?

pkg/database/model.go
```diff
@@ -266,3 +266,20 @@ func NewUUID() string {
...
+	Role      string    `gorm:"unique;index;not null"`
+	CreatedAt time.Time
+	UpdatedAt time.Time
+}
+
+type RoleNamespace struct {
+	ID          uuid.UUID `gorm:"type:uuid;primaryKey"`
```
suprjinx, line 278: and here?

PR #1081, opened 2024-04-04
Update creds import in python client

fabiovincenzi:


jgiannuzzi:
please test this on older versions of mlflow and/or introduce a qualifier for the version of mlflow we support in `pyproject.toml`

jgiannuzzi:
you may need to have some conditional imports that catch an exception and do something different when `mlflow.utils.credentials` does not exist

suprjinx (approved):
do we need a version reference in pyproject.toml to make sure the latest version of mlflow is in use?

jgiannuzzi (approved):
LGTM 👍

PR #1104, opened 2024-04-10
Add the k6 load test script to the docs

suprjinx:
@jgiannuzzi's k6 load test script, with a few tweaks for configuring the shape.


fabiovincenzi (commented):
Looks good! I would just increase the  unique metrics per run, something around 2k, and it would be nice to also have some context as well for some of those

fabiovincenzi (approved):
Looks good!

PR #1106, opened 2024-04-11
OIDC business logic implementation.

dsuhinin:
What's inside the box:
- OIDC implementation for `aim`, `mlflow`, `admin`, `chooser` parts.
- a bit of refactoring - several packages were moved to `common`.

### DIFF
--- a/pkg/database/migrations/v_0013/migrate.go
+++ b/pkg/database/migrations/v_0013/migrate.go
@@ -4,7 +4,6 @@ import (
 	"gorm.io/gorm"
 
 	"github.com/G-Research/fasttrackml/pkg/database/migrations"
-	"github.com/G-Research/fasttrackml/pkg/database/migrations/v_0012"
 )
 
 const Version = "20240423055442"
@@ -12,7 +11,10 @@ const Version = "20240423055442"
 func Migrate(db *gorm.DB) error {
 	return migrations.RunWithoutForeignKeyIfNeeded(db, func() error {
 		return db.Transaction(func(tx *gorm.DB) error {
-			if err := tx.Migrator().DropColumn(&v_0012.Role{}, "Role"); err != nil {
+			if err := tx.Migrator().DropTable(Role{}); err != nil {
+				return err
+			}
+			if err := tx.Migrator().DropTable(RoleNamespace{}); err != nil {
 				return err
 			}
 			if err := tx.AutoMigrate(&Role{}); err != nil {


pkg/cmd/server.go
```diff
@@ -76,6 +76,9 @@ func init() {
 	ServerCmd.Flags().String("auth-oidc-client-id", "", "OIDC auth client id")
 	ServerCmd.Flags().String("auth-oidc-client-secret", "", "OIDC auth client secret")
 	ServerCmd.Flags().String("auth-oidc-provider-endpoint", "", "OIDC auth provider endpoint")
+	ServerCmd.Flags().String("auth-oidc-scopes", "", "OIDC requested scopes")
+	ServerCmd.Flags().String("auth-oidc-admin-role", "", "OIDC admin role identifier")
+	ServerCmd.Flags().String("auth-oidc-claim-roles", "", "OIDC claim roles")
```
suprjinx, line 81: I'm not really familiar with OIDC, but I wonder if the descriptions could be a little clearer -- is this "OIDC claim that communicates roles"?

pkg/common/client/oidc/models.go
```diff
@@ -0,0 +1,17 @@
+package oidc
```
suprjinx, line 1: maybe this should be `auth` package instead of `oidc`, since it seems more general-purpose

pkg/common/dao/repositories/roles_cached.go
```diff
@@ -0,0 +1,131 @@
...
+		).Where(
+			&models.Namespace{Code: requestedNamespaceCode},
+		),
+	).Find(&data).Error; err != nil {
+		return false, eris.Wrapf(err, "error getting roles for namespace with code: %s", requestedNamespaceCode)
+	}
+
+	namespaceRoles = make([]string, len(data))
+	for i, namespaceRole := range data {
+		namespaceRoles[i] = namespaceRole.Role.Role
```
suprjinx, line 102: duplicate attribute `Role`?

pkg/server/server.go
```diff
@@ -194,7 +170,28 @@ func createApp(
...
 	}
 
-	app.Use(middleware.NewNamespaceMiddleware(namespaceRepository))
+	// create namespace notification listener.
+	namespaceEventListener, err := dao.NewNamespaceListener(ctx, db.GormDB())
+	if err != nil {
+		return nil, eris.Wrap(err, "error creating namespace notification listener")
+	}
+
+	namespaceCachedRepository, err := mlflowRepositories.NewNamespaceCachedRepository(
```
suprjinx, line 179: for another ticket, it seems like `NamespaceCachedRepository` should move to the `common/dao` package

pkg/cmd/server.go
suprjinx: How about description "OIDC claim to inspect for roles" ?

suprjinx (approved):

PR #1109, opened 2024-04-11
Benchmark

Shugaba-Wuta:
#260  Performance benchmark of MLFlow and FastTrackML


suprjinx:
I wouldn't expect any file changes outside of the `/benchmarks` folder -- looks like some other changes got caught up here

.gitignore
Shugaba-Wuta, line 1: @suprjinx should the files generated benchmark_outputs be tracked by Git? I am thinking of leaving `benchmark_outputs**`

suprjinx: good question -- i don't think we need to track right now

suprjinx (approved):
looks good!

PR #1152, opened 2024-04-23
OIDC integration tests

dsuhinin:
What's inside the box:
- integration tests for OIDC flow for `admin`, `aim`, `mlflow`, `chooser` parts.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -259,16 +259,22 @@ migrations-rebuild: ## rebuild the migrations script to detect new migrations.
 	@echo ">>> Running FastTrackML migrations rebuild."
 	@go run main.go migrations rebuild
 
-.PHONY: ui-aim-start
-ui-aim-start: ## start the Aim UI for development.
+.PHONY: ui-aim-sync
+ui-aim-sync: ## copy Aim UI files to docker volume.
 	@echo ">>> Syncing the Aim UI."
 	@rsync -rvu --exclude node_modules --exclude .git ui/fasttrackml-ui-aim/ $(AIM_BUILD_LOCATION)
+
+.PHONY: ui-aim-start
+ui-aim-start: ui-aim-sync ## start the Aim UI for development.	
 	@echo ">>> Starting the Aim UI."
 	@cd $(AIM_BUILD_LOCATION)/src && npm ci --legacy-peer-deps && npm start
 
-.PHONY: ui-mlflow-start
-ui-mlflow-start: ## start the MLflow UI for development.
+.PHONY: ui-mlflow-sync
+ui-mlflow-sync: ## copy MLflow UI files to docker volume.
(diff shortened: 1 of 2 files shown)


docs/authentication.md
```diff
@@ -0,0 +1,70 @@
...
+
+## Auth configuration
+
+FastTrackML supports 2 types of authentication: OIDC based authentication and Basic authentication.
+
+### OIDC Authentication
+
+To enable OIDC authentication, FastTrackML should be run with next command line parameters:
+```
+--auth-oidc-client-id client_id --auth-oidc-client-secret client_secret --auth-oidc-provider-endpoint http://127.0.0.1 --auth-oidc-admin-role admin --auth-oidc-claim-roles groups --auth-oidc-scopes email openid
```
suprjinx, line 14: maybe break line with backslash-newline after each option so it's easier to read?

docs/authentication.md
```diff
@@ -0,0 +1,70 @@
...
+Basic authentication supports 2 different ways:
+```
+--auth-username username --auth-password password
+```
+in that case FastTrackML will be restricted with `username` and `password` and user will have access to all the existing resources: `mlflow`, `aim`, `admin`, `chooser`.
+```
+--auth-username username --auth-password password --auth-users-config /path/to/config.file
+```
+where:
+- `auth-users-config` is a `users` configuration file which should have next format:
```
suprjinx, line 47: "should have the following format:"

docs/authentication.md
```diff
@@ -0,0 +1,70 @@
...
+      - ns:first
+      - ns:second
+      - ns:third
+  - name: user3
+    password: password3
+    roles:
+      - ns:default
+      - ns:third
+```
+so in that case FastTrackML will use `auth-username` and `auth-password` to check that this user exists in
```
suprjinx, line 67: this is confusing since it appears you mean the command-line options `auth-username` when I think this refers to the Basic auth http header

tests/integration/golang/auth/auth_oidc_based_test.go
```diff
@@ -0,0 +1,612 @@
...
+
+func (s *OIDCAuthTestSuite) TestAIMAuth_Ok() {
+	s.SetupTestSuite()
+
+	tests := []struct {
+		name  string
+		check func()
+	}{
+		{
+			name: "TestUser1Access",
```
suprjinx, line 147: Maybe better to name these for different aspects of test, eg "UserCannotAccessUnauthorizedNamespace"

docs/authentication.md
suprjinx: in this case, with an auth config file, we also supply `auth-username` command line option?

suprjinx (approved):

Issue #1145, opened 2024-04-22
Aim /runs/search should support experiment_names


suprjinx:
relates to #1119

PR #1155, opened 2024-04-24
Added experiment_names to runs/search/run 

Shugaba-Wuta:
Issue https://github.com/G-Research/fasttrackml/issues/1145

Added required `experiment_names` to GET `/runs/search/run`


tests/integration/golang/aim/run/search_test.go
```diff
@@ -262,20 +263,33 @@ func (s *SearchTestSuite) SetupTest() {
 	s.Require().Nil(err)
 }
 
+func (s *SearchTestSuite) TestNoRequiredExperimentNames() {
```
suprjinx, line 266: I think we append append `_Error` for the tests which expect error, eg `TestValidator_Error`

pkg/api/aim2/dao/repositories/run.go
```diff
@@ -407,6 +407,10 @@ func (r RunRepository) SearchRuns(
 		).
 		Order("row_num DESC")
 
+	// if len(req.ExperimentNames) != 0 {
```
suprjinx, line 410: seems like this should be uncommented

suprjinx: once uncommented, there are many test failures. trying to figure out the right encantation...

pkg/api/aim2/controller/runs.go
```diff
@@ -105,6 +105,9 @@ func (c Controller) SearchRuns(ctx *fiber.Ctx) error {
 	if err = ctx.QueryParser(&req); err != nil {
 		return fiber.NewError(fiber.StatusUnprocessableEntity, err.Error())
 	}
+	if err := run.ValidateGetSearchRunsRequest(&req); err != nil {
```
suprjinx, line 108: I'm thinking maybe we don't need this validation -- it's not an error, we should just craft the query so there are no results if there's no selected experiments

suprjinx: i think we removed error in other cases where there's no selection, eg Metrics -- you just don't get results until you pick something.

pkg/api/aim/request/run.go
```diff
@@ -21,12 +21,13 @@ type GetRunMetric struct {
...
+	Query           string   `query:"q"`
+	Limit           int      `query:"limit"`
+	Offset          string   `query:"offset"`
+	Action          string   `query:"action"`
+	SkipSystem      bool     `query:"skip_system"`
+	ReportProgress  bool     `query:"report_progress"`
+	ExcludeParams   bool     `query:"exclude_params"`
+	ExcludeTraces   bool     `query:"exclude_traces"`
+	ExperimentNames []string `query:"experiment_names"`
 }
```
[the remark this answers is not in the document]
suprjinx: reverted

pkg/api/aim2/services/run/validators.go
```diff
@@ -28,3 +28,10 @@ func ValidateGetRunInfoRequest(req *request.GetRunInfoRequest) error {
 	}
 	return nil
 }
+
+func ValidateGetSearchRunsRequest(req *request.SearchRunsRequest) error {
+	if len(req.ExperimentNames) == 0 {
+		return api.NewBadRequestError("required `experiment_names` is missing")
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
suprjinx: removed this test

pkg/api/aim/request/run.go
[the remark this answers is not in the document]
suprjinx: actually, this is necessary for the integration tests to run against both, so we'll need to fix this up when we remove aim/aim2 support

Issue #1108, opened 2024-04-11
Large logging batch fails in sqlite


suprjinx:
`time="2024-04-11T17:11:07Z" level=error msg="Error encountered in POST /api/2.0/mlflow/runs/log-batch: INTERNAL_ERROR: unable to insert metrics for run '5bcb0762756e4cc88da1c61b0d87e57f': error updating latest metrics for run: 5bcb0762756e4cc88da1c61b0d87e57f: too many SQL variables" func=github.com/G-Research/fasttrackml/pkg/api/mlflow/services.ErrorHandler file="/workspaces/fasttrackml/pkg/api/mlflow/services/errors.go:53"`

this is with 8k metrics in one batch.

suprjinx:
It looks like the block updating LatestMetrics does not use batch operation `mlflow/dao/repositories/metric.go#165`

PR #1175, opened 2024-05-01
mlflow: fixed log-batch limit reached with sqlite

Shugaba-Wuta:
#1108 Fix failing batch logging on sqlite


suprjinx (approved):
Looks good!

PR #1185, opened 2024-05-06
Fix issues related to context serialisation for `/metrics/search` endpoint.

dsuhinin:
What's inside the box:
- now `/runs/search/metric` endpoint correctly serialise `context` object before search.
- `integration` tests were adjusted.


suprjinx (approved):
LGTM

suprjinx:
we'll also need to bump the UI module tag when change merges -- might want to wait and add here

suprjinx (approved):

PR #1199, opened 2024-05-13
Create integration tests to test FastTrackML over MLFlow native database

dsuhinin:
What's inside the box:
- integration tests to test FastTrackML over original  MLFlow database

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -12,7 +12,7 @@ require (
 	dagger.io/dagger v0.11.4
 	dario.cat/mergo v1.0.0
 	github.com/DATA-DOG/go-sqlmock v1.5.2
-	github.com/G-Research/fasttrackml-ui-aim v0.31705.44
+	github.com/G-Research/fasttrackml-ui-aim v0.31705.46
 	github.com/G-Research/fasttrackml-ui-mlflow v0.20902.7
 	github.com/PuerkitoBio/goquery v1.9.2
 	github.com/apache/arrow/go/v14 v14.0.2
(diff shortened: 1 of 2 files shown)


tests/integration/mlflow-setup.py
```diff
@@ -0,0 +1,16 @@
+from random import randint, random
+
+import mlflow
+from mlflow import log_metric, log_param
+
+mlflow.set_tracking_uri("http://mlflow-server:5000")
```
jgiannuzzi, line 6: this tracking URI should be the database URI instead, and then you can run the script directly, with no intermediate MLflow server

tests/integration/MLflow.Server.Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM python:latest
+
+# Install python packages
+RUN pip install mlflow psycopg2 boto3
+COPY mlflow-setup.py .
+CMD mlflow server --backend-store-uri ${BACKEND_STORE_URI} --host 0.0.0.0 --port 5000
+
```
jgiannuzzi, line 7: we probably want to be able to target different versions of MLflow (and maybe even Python)

jgiannuzzi (commented):
have a look at how we currently run the Python integration tests for inspiration on how we could dynamically create Python containers in which we can run different versions of MLflow: https://github.com/G-Research/fasttrackml/tree/main/tests/integration/python

suprjinx (approved):

PR #1223, opened 2024-05-23
Fix python-integration-tests name

pavlovic-ivan:
This PR fixes the missing part from the name of `python-integration-tests` job names. `api` was being picked up from the matrix, which doesn't exists, which made github print out empty parenthesis in the job name. Matrix defines 'test-targets' instead. 

Ran into this accidentally, saw an opportunity and fixed it (boy scout rule).

Working example: https://github.com/pavlovic-ivan/fasttrackml/actions/runs/9207382745


jgiannuzzi (approved):
LGTM, thanks @pavlovic-ivan!

PR #1225, opened 2024-05-23
Integrate OIDC login/logout functionality.

dsuhinin:
What's inside the box:
- OIDC login functionality.
- OIDC logout functionality.

### DIFF
--- a/website/docusaurus.config.js
+++ b/website/docusaurus.config.js
@@ -16,7 +16,8 @@ module.exports = async function configCreatorAsync() {
     );
     const releaseVersion = data.tag_name.substring(1);
     const releaseUrl = data.html_url;
-    const slackUrl = 'https://join.slack.com/t/mlops-community/shared_invite/zt-258yl6ifq-zZNzBaddwlhifj56yt7ONg';
+    const slackUrl = 'https://mlops-community.slack.com/archives/C05ALNZ1XJN'
+    const slackInviteUrl = 'https://join.slack.com/t/mlops-community/shared_invite/zt-258yl6ifq-zZNzBaddwlhifj56yt7ONg';
 
     /** @type {import('@docusaurus/types').Config} */
     return {
@@ -92,6 +93,11 @@ module.exports = async function configCreatorAsync() {
                         },
                         {
                             href: slackUrl,
+                            label: 'Go to our Slack',
+                            position: 'right',
+                        },
+                        {
+                            href: slackInviteUrl,
                             label: 'Join our Slack',
                             position: 'right',
                         },


pkg/common/config/auth/auth.go
```diff
@@ -28,12 +27,16 @@ type Config struct {
 
 // IsAuthTypeOIDC makes check that current auth is TypeOIDC.
 func (c *Config) IsAuthTypeOIDC() bool {
-	return c.AuthType == TypeOIDC
+	return c.AuthOIDCClientID != "" &&
```
suprjinx, line 30: this seems like it's validating the config before letting us characterize it as a type of auth -- ie, it seems like it could "be" AuthTypeOIDC even if not configured right.

pkg/common/dao/repositories/roles_cached.go
```diff
@@ -93,7 +93,7 @@ func (r RoleCachedRepository) ValidateRolesAccessToNamespace(
 		).Where(
 			&models.Namespace{Code: requestedNamespaceCode},
 		),
-	).Find(&data).Error; err != nil {
+	).Debug().Find(&data).Error; err != nil {
```
suprjinx, line 96: Need this `Debug`?

pkg/common/middleware/oidc.go
```diff
@@ -78,22 +79,14 @@ func (m OIDCMiddleware) handleChooserResourceRequest(ctx *fiber.Ctx) error {
...
-
-		log.Debugf("user has roles: %v accociated", user.GetRoles())
-		ctx.Locals(oidcUserContextKey, user)
+	user, err := m.client.Verify(ctx.Context(), ctx.Cookies("access_token", ""))
+	if err != nil {
+		log.Errorf("error verifying access token: %+v", err)
+		ctx.Response().Header.Add("Cache-Control", "no-store")
+		return ctx.Redirect("/login", http.StatusMovedPermanently)
 	}
+	log.Debugf("user has roles: %v accociated", user.GetRoles())
```
suprjinx, line 88: "associated"

pkg/common/config/auth/auth.go
suprjinx: it's just kind of semantic -- previously we "declared" the `AuthType`, now it must be valid config, which seems like it might be confusing at some point. A person could attempt to configure for OIDC, but some field is missing, so the error might say  "No auth type is configured" instead of "Invalid OIDC config".

suprjinx: why remove the `AuthType` field?

suprjinx: I think it should attempt to use the OIDC middleware if that's intended, even if some field is missing.

suprjinx (approved):

PR #1260, opened 2024-06-05
Run logs

### LINKED ISSUE #1253: Run Logs: Server endpoint for persisting logs

suprjinx:
Fixes #1253


pkg/api/aim/api/response/log.go
```diff
@@ -0,0 +1,64 @@
...
+	"github.com/gofiber/fiber/v2"
+	"github.com/rotisserie/eris"
+	log "github.com/sirupsen/logrus"
+
+	"github.com/G-Research/fasttrackml/pkg/api/aim/dao/models"
+	"github.com/G-Research/fasttrackml/pkg/api/aim/encoding"
+)
+
+// NewGetRunLogsResponse creates a new response object for `GET /runs/:id/logs` endpoint.
+func NewGetRunLogsResponse(
```
suprjinx, line 18: nice!

pkg/api/mlflow/dao/models/log.go
```diff
@@ -0,0 +1,9 @@
+package models
+
+// Log represents a row of the `logs` table.
+type Log struct {
+	ID        uint   `gorm:"primaryKey;autoIncrement"`
+	Value     string `gorm:"not null"`
+	RunID     string `gorm:"column:run_uuid;not null;index"`
```
suprjinx, line 7: doesn't really matter here, but the gorm annotation for index needs to have the shared name `idx_run_timestamp`

pkg/database/model.go
```diff
@@ -165,6 +166,13 @@ type LatestMetric struct {
 	Context   Context
 }
 
+type Log struct {
+	ID        uint   `gorm:"primaryKey;autoIncrement"`
+	Value     string `gorm:"not null"`
+	RunID     string `gorm:"column:run_uuid;not null;index"`
```
suprjinx, line 172: so we don't want a compound index on run and timestamp?

pkg/api/mlflow/dao/models/log.go
suprjinx: previously there was index on runID + timestamp as a compound index (which requires both fields to reference the same index name). Now it is just separate index on Timestamp and RunID -- but the name for the Timestamp one looks like a compound.

pkg/database/migrations/v_0015/model.go
```diff
@@ -0,0 +1,292 @@
...
+	RunID     string `gorm:"column:run_uuid;not null;primaryKey;index"`
+	LastIter  int64
+	ContextID uint `gorm:"not null;primaryKey"`
+	Context   Context
+}
+
+type Log struct {
+	ID        uint   `gorm:"primaryKey;autoIncrement"`
+	Value     string `gorm:"not null"`
+	RunID     string `gorm:"column:run_uuid;not null;index"`
```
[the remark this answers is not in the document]
suprjinx: see comment above

pkg/database/model.go
suprjinx: see above

dave-gantenbein (approved):

PR #1321, opened 2024-06-28
Support artifacts logging in AIM

dsuhinin:
What's inside the box:
- new endpoint to log different artifact for AIM UI.


pkg/api/aim/api/request/run.go
```diff
@@ -22,6 +22,18 @@ type GetRunMetricsRequest []struct {
...
+// CreateRunArtifactRequest is a request object for `POST /runs/:id/artifacts` endpoint.
+type CreateRunArtifactRequest struct {
+	Iter    int64  `json:"iter"`
+	Step    int64  `json:"step"`
+	Caption string `json:"caption"`
+	Index   int64  `json:"index"`
+	Width   int64  `json:"width"`
+	Height  int64  `json:"height"`
+	Format  string `json:"format"`
+	BlobURI string `json:"blob_uri"`
```
suprjinx, line 34: i'm wondering if `BlobURI` might better be the `StorageKey` or something (full path to append to the artifact uri). That way it's relying on the artifact storage configuration for the scheme/host and this would just be the key therein. ?

pkg/api/aim/controller/runs.go
```diff
@@ -62,6 +63,26 @@ func (c Controller) GetRunMetrics(ctx *fiber.Ctx) error {
 	return ctx.JSON(resp)
 }
 
+// CreateRunArtifact handles `POST /runs/:id/artifacts` endpoint.
+func (c Controller) CreateRunArtifact(ctx *fiber.Ctx) error {
```
suprjinx, line 67: Should this be in the `mlflow` api space, along with `log_metric` etc? The `mlflow` client has `log_artifact` for this.

pkg/api/aim/api/request/run.go
suprjinx: though I do see that Aim refers to it as BlobURI `self.assertIn('blob_uri',` img)` so i can be convinced either way

suprjinx (approved):
LGTM

Issue #228, opened 2023-08-09
Make artifacts visible in Aim UI


jgiannuzzi:
Once #6 and #224 are implemented, artifacts will become visible in the MLFlow UI.

This issue is about making them visible in the Aim UI. Aim treats artifacts like metrics and thus expects a few extra properties. Here is how I suggest we map them initially:
| Aim | FML |
| -- | -- |
| name | file name |
| context | `{}` |
| step | `0` |
| epoch | `0` |
| timestamp | file timestamp |

There is also limited support for artifact types. I think we only want to expose images and texts initially. We could detect the file's MIME type to distinguish between images and texts. The search possibilities will be quite limited and tricky unless we implement #225 and use it to store metadata in our database.

Sharpz7:
This would be great to see folks. Thank you :))

suprjinx:
here are the endpoints Aim provides:
https://github.com/aimhubio/aim/blob/main/aim/web/api/tags/views.py

PR #1335, opened 2024-07-05
AIM image artefacts

fabiovincenzi:
Fixes https://github.com/G-Research/fasttrackml/issues/228.


pkg/api/aim/controller/runs.go
```diff
@@ -62,6 +76,75 @@ func (c Controller) GetRunMetrics(ctx *fiber.Ctx) error {
...
+	images, err := c.runService.GetRunImagesBatch(ctx.Context(), &req)
+	if err != nil {
+		return err
+	}
+	imagesMap, err := convertImagesToMap(images, req)
+	if err != nil {
+		return err
+	}
+
+	ctx.Context().Response.SetBodyStreamWriter(func(w *bufio.Writer) {
```
suprjinx, line 122: can this go into the `response` package with other similar?

pkg/api/aim/api/response/run.go
```diff
@@ -19,6 +21,8 @@ import (
 	"github.com/G-Research/fasttrackml/pkg/api/aim/dao/models"
 	"github.com/G-Research/fasttrackml/pkg/api/aim/dao/repositories"
 	"github.com/G-Research/fasttrackml/pkg/api/aim/encoding"
+	mlflowCommon "github.com/G-Research/fasttrackml/pkg/api/mlflow/common"
+	"github.com/G-Research/fasttrackml/pkg/api/mlflow/services/artifact/storage"
```
suprjinx, line 25: I'd say we shouldn't introduce a new `aim` -> `mlflow` dependency, since we've managed to get rid of these all pretty recently. Maybe introduce semi-duplicate stuff on the aim side, or move from `mlflow` to `common`? Duplicates for Aim is what I was doing for image search, since the needs are kind of different here.

pkg/api/aim/dao/models/image.go
```diff
@@ -0,0 +1,25 @@
...
+	Step    int         `json:"step"`
+	Width   interface{} `json:"width"`
+}
+
+type Image struct {
+	RecordRange interface{}     `json:"record_range"`
+	IndexRange  interface{}     `json:"index_range"`
+	Name        string          `json:"name"`
+	Context     interface{}     `json:"context"`
+	Values      [][]ImageValues `json:"values"`
```
suprjinx, line 23: I think this is array of `step`, and second dimension is images per step (like `iter`).

suprjinx: additionally, `record_range` seems to reference steps, and `index_range` represents the images per step.

pkg/api/aim/api/response/run.go
fabiovincenzi: I'll create a ticket for that

pkg/api/aim/dao/models/image.go
```diff
@@ -0,0 +1,25 @@
...
+	Index   interface{} `json:"index"`
+	Key     string      `json:"key"`
+	SeqKey  string      `json:"seqKey"`
+	Name    string      `json:"name"`
+	Run     interface{} `json:"run"`
+	Step    int         `json:"step"`
+	Width   interface{} `json:"width"`
+}
+
+type Image struct {
```
suprjinx, line 18: i like the way you modeled this, but our DB uses table/model `Artifact` for these attributes. Maybe we can merge, could be in a new ticket.

pkg/api/aim/services/run/service.go
```diff
@@ -4,13 +4,19 @@ import (
...
+	"io/fs"
+	"net/url"
 
 	"github.com/rotisserie/eris"
 
 	"github.com/G-Research/fasttrackml/pkg/api/aim/api/request"
 	"github.com/G-Research/fasttrackml/pkg/api/aim/common"
 	"github.com/G-Research/fasttrackml/pkg/api/aim/dao/models"
 	"github.com/G-Research/fasttrackml/pkg/api/aim/dao/repositories"
+	"github.com/G-Research/fasttrackml/pkg/api/mlflow/services/artifact/storage"
```
suprjinx, line 19: can you make new ticket to move this to `common`?

suprjinx (approved):

PR #1347, opened 2024-07-12
Aim images search updates

### LINKED ISSUE #1346: Aim Images: support `images.name` for sequence
The Aim images explorer is expecting a concept of `image.name` which would be like metric.key --- the name for a whole sequence of values (images).

suprjinx:
Fixes #1346


tests/integration/python/config.json
jescalada, line 1: Should we add `test_run_images_api.py` here? I noticed those set of tests are not executing yet

jescalada (approved):
Looks good to me!

Issue #1334, opened 2024-07-04
Bug: Run selector on Run page doesn't show relevant info


jescalada:
The run info is loading correctly on the Overview tab, but not on the Run selector, which displays "Invalid Date" and "0ms" instead of the relevant values.

![image](https://github.com/G-Research/fasttrackml/assets/97265671/0c39aacd-99d9-4abd-bcc7-cd54e82d2299)

jescalada:
Here's another example in the experiment overview page:

![image](https://github.com/user-attachments/assets/8c269cde-14dc-4207-be95-6ecea3182051)

PR #1369, opened 2024-08-02
(bugfix) Make Experiments Overview Page Show Dates & Runs

vinayan3:
# Date fix
The server is sending data in camel case which causes the experiment overview page to show invalid date instead of the actual date. The expected data type is here:  https://github.com/G-Research/fasttrackml-ui-aim/blob/release/v3.17.5/src/src/modules/core/api/experimentsApi/types.ts#L84-L90 

Before the change
Screenshot from 2024-08-01 17-58-25

After the change
Screenshot from 2024-08-01 17-58-51

# Runs fix
When you click the runs tab it make an API call to the AIM end points for `GET /runs/search/run`. The UI's code sets the experiment as a query see:
https://github.com/G-Research/fasttrackml-ui-aim/blob/release/v3.17.5/src/src/pages/Experiment/components/ExperimentRunsTab/ExperimentRunsTable/useExperimentRunsTable.tsx#L224-L228

If `ExperimentNames`is empty it shouldn't be included in the query on the DB.

### DIFF
--- a/pkg/api/aim/api/response/experiment.go
+++ b/pkg/api/aim/api/response/experiment.go
@@ -43,8 +43,8 @@ func NewGetExperimentsResponse(experiments []models.ExperimentExtended) []Experi
 type ExperimentRunPartial struct {
 	ID           string `json:"run_id"`
 	Name         string `json:"name"`
-	CreationTime int64  `json:"creationTime"`
-	EndTime      int64  `json:"endTime"`
+	CreationTime int64  `json:"creation_time"`
+	EndTime      int64  `json:"end_time"`
 	Archived     bool   `json:"archived"`
 }
 
--- a/pkg/api/aim/dao/repositories/run.go
+++ b/pkg/api/aim/dao/repositories/run.go
@@ -398,16 +398,22 @@ func (r RunRepository) SearchRuns(
 
 	log.Debugf("Total runs: %d", total)
 
+	experimentsSelect := database.DB.Select(
+		"ID", "Name",
+	).Where(
+		&models.Experiment{NamespaceID: namespaceID},
+	)
+	// Do not include the filter if there are no ExperimentNames
+	if len(req.ExperimentNames) > 0 {
+		experimentsSelect = experimentsSelect.Where(
+			`"Experiment"."name" IN ?`, req.ExperimentNames,
+		)
+	}
+
 	tx := r.GetDB().WithContext(ctx).
 		InnerJoins(
 			"Experiment",
-			database.DB.Select(
-				"ID", "Name",
-			).Where(
-				&models.Experiment{NamespaceID: namespaceID},
-			).Where(
-				`"Experiment"."name" IN ?`, req.ExperimentNames,
-			),
+			experimentsSelect,
 		).
 		Order("row_num DESC")
 


jescalada (approved):
I tested all the issues we were having in https://github.com/G-Research/fasttrackml/issues/1334 and it looks good to me! The code looks fine as well.

Pinging @suprjinx for feedback on the query just in case.

jescalada:
@vinayan3 It seems however that some of the Go tests related to Run Searching are failing ATM. Let's make sure that those are passing first 😃

vinayan3:
@jescalada so I'll need to change the test `NoExperimentNamesNoResults` . The behavior is now different where if you don't provide an experiment name then it allows all.  Instead it'll be `NoExpeirmentNamesResultsFromQuery`.

pkg/api/aim/api/response/experiment.go
```diff
@@ -43,8 +43,8 @@ func NewGetExperimentsResponse(experiments []models.ExperimentExtended) []Experi
 type ExperimentRunPartial struct {
 	ID           string `json:"run_id"`
 	Name         string `json:"name"`
-	CreationTime int64  `json:"creationTime"`
-	EndTime      int64  `json:"endTime"`
+	CreationTime int64  `json:"creation_time"`
+	EndTime      int64  `json:"end_time"`
```
suprjinx, line 47: good catch 👍

pkg/api/aim/dao/repositories/run.go
```diff
@@ -398,16 +398,22 @@ func (r RunRepository) SearchRuns(
 
 	log.Debugf("Total runs: %d", total)
 
+	experimentsSelect := database.DB.Select(
+		"ID", "Name",
+	).Where(
+		&models.Experiment{NamespaceID: namespaceID},
+	)
+	// Do not include the filter if there are no ExperimentNames
+	if len(req.ExperimentNames) > 0 {
```
suprjinx, line 407: I don't think we can support this change -- experiment names are required for fetching data as a way to limit scope, otherwise it can really bog the UI down before the user has even done anything.

We could theoretically add a new function for "50 most recent runs" or something to show on the overview page?

tests/integration/golang/aim/run/search_test.go
```diff
@@ -469,12 +469,17 @@ func (s *SearchTestSuite) TestStreamData_Ok() {
...
 				Query:           `run.archived == True or run.archived == False`,
 				ExperimentNames: []string{},
 			},
-			runs: []*models.Run{},
+			runs: []*models.Run{
+				s.run1,
+				s.run2,
+				s.run3,
+				s.run4,
+			},
```
suprjinx, line 482: this shows the problem -- there could be 1500 runs per experiment, and this would return them all!

suprjinx (changes requested):
Great idea -- but I think we need to limit runs returned when no experiment names are provided. Additionally, I think `SearchRuns` should always require the experiment selection -- maybe a new repository function for use in the overview which returns most recent runs?

vinayan3:
> Great idea -- but I think we need to limit runs returned when no experiment names are provided. Additionally, I think `SearchRuns` should always require the experiment selection -- maybe a new repository function for use in the overview which returns most recent runs?

Could a solution here to be add a limit field to the API request and make experiment names or limit required?

 Otherwise, I'm fine with adding a new request for the overview call.

suprjinx:
@vinayan3 -- I think maybe we should keep your changes for `creation_time`, `end_time` serialization in this PR, and then open a new PR in the UI repo. The runs tab in the experiment overview is issuing a request like this:
`http://localhost:5000/aim/api/runs/search/run?limit=50&exclude_params=true&q=run.experiment%20==%20%27my-experiment%27`

But we need it to issue one like this:
`http://localhost:5000/aim/api/runs/search/run?q=&experiment_names=my-experiment&limit=45`

So we would just alter the `fetchExperimentRuns` function you highlighted to structure the params differently, and it should work.

pkg/api/aim/dao/repositories/run.go
vinayan3: I've reverted it. I'll have to work through the alternative path of getting the UI to hit a different endpoint. That will be in a sep PR.

vinayan3:
> @vinayan3 -- I think maybe we should keep your changes for `creation_time`, `end_time` serialization in this PR, and then open a new PR in the UI repo. The runs tab in the experiment overview is issuing a request like this: `http://localhost:5000/aim/api/runs/search/run?limit=50&exclude_params=true&q=run.experiment%20==%20%27my-experiment%27`
> 
> But we need it to issue one like this: `http://localhost:5000/aim/api/runs/search/run?q=&experiment_names=my-experiment&limit=45`
> 
> So we would just alter the `fetchExperimentRuns` function you highlighted to structure the params differently, and it should work.

I'll have to work through hitting a new endpoint in the UI later this week.

suprjinx:
> > @vinayan3 -- I think maybe we should keep your changes for `creation_time`, `end_time` serialization in this PR, and then open a new PR in the UI repo. The runs tab in the experiment overview is issuing a request like this: `http://localhost:5000/aim/api/runs/search/run?limit=50&exclude_params=true&q=run.experiment%20==%20%27my-experiment%27`
> > But we need it to issue one like this: `http://localhost:5000/aim/api/runs/search/run?q=&experiment_names=my-experiment&limit=45`
> > So we would just alter the `fetchExperimentRuns` function you highlighted to structure the params differently, and it should work.
> 
> I'll have to work through hitting a new endpoint in the UI later this week.

Cool thank you! I think it's not even a new endpoint, just changing the param structure in the ajax request (from `q` to `experiment_names`).

PR #1386, opened 2024-08-15
S3 Testing and adjustments

jescalada:
PR to add documentation for AWS logging.

After testing with the right values for each environment variable required by both the backend and AWS S3, it turns out we don't need to make any changes. I added a guide with the variables and example values.

### Changelog
- Added documentation on how to get FML to work with AWS S3
- ~Set `UsePathStyle` property to false in order to get AWS S3 to track correctly (otherwise the resource cannot be found)~

### DIFF
--- a/docs/aws_s3_tracking.md
+++ b/docs/aws_s3_tracking.md
@@ -1,0 +1,41 @@
+# FastTrackML with AWS S3 Setup
+
+To use FastTrackML with AWS S3 for artifact storage, you'll need to configure your environment variables and make sure the AWS S3 bucket is set up correctly.
+
+## Environment Variables
+
+`AWS_ACCESS_KEY_ID`
+Your AWS access key ID in the format `AKIAxxxxxxxxxxxx` (20 characters).
+
+`AWS_SECRET_ACCESS_KEY`
+Your AWS secret access key (40 characters).
+
+`AWS_DEFAULT_REGION`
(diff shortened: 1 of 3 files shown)


docs/aws_s3_tracking.md
```diff
@@ -0,0 +1,41 @@
...
+`AWS_SECRET_ACCESS_KEY`
+Your AWS secret access key (40 characters).
+
+`AWS_DEFAULT_REGION`
+The AWS region where your S3 bucket is located. Example: `us-east-1`, `us-west-2`, etc.
+
+`FML_S3_ENDPOINT_URI`
+The endpoint URI for your S3 bucket. This will change depending on the path-style access type. If path-style access is disabled, exclude the bucket name subdomain. Example: `https://s3.your-region.amazonaws.com`. Otherwise, the bucket subdomain can be included: `https://your-bucket-name.s3.your-region.amazonaws.com`.
+
+<!-- FML_S3_USE_PATH_STYLE
```
suprjinx, line 19: should we delete this?

docs/aws_s3_tracking.md
suprjinx, line 1: Nice doc 👍

jescalada:
After testing once again the whole flow with proper values for the URIs and AWS params, everything seems to work fine even when omitting the path style. I am a bit confused as to why it wasn't quite working before. 🤔

suprjinx (approved):
Doc only :-)

PR #1490, opened 2025-01-14
Add Attestations

nojaf:
See https://docs.github.com/en/actions/security-for-github-actions/using-artifact-attestations/using-artifact-attestations-to-establish-provenance-for-builds

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -129,16 +129,21 @@ jobs:
           echo "::endgroup::"
           # Output digests as a JSON array for use in a matrix
           echo "::set-output name=digests::$(printf '%s\n' "${digests[@]}" | jq -R . | jq -s .)"
+      outputs:
+        digests: ${{ steps.set-matrix.outputs.matrix }}
 
-      - name: Attest Each Docker Image
-        uses: actions/attest-build-provenance@v2
-        with:
-          subject-name: index.docker.io/${{ vars.DOCKER_REPO }}
-          subject-digest: ${{ matrix.digest }}
-          push-to-registry: true
-        strategy:
-          matrix:
-            digest: ${{ fromJson(steps.push.outputs.digests) }}
+    attest-docker-images:
+      name: Publish Docker images attestations
+      strategy:
+        matrix:
+          digest: ${{ fromJson(needs.docker-release.outputs.digests) }}
+      steps:
+        - name: Attest Each Docker Image
+          uses: actions/attest-build-provenance@v2
+          with:
+            subject-name: index.docker.io/${{ vars.DOCKER_REPO }}
+            subject-digest: ${{ matrix.digest }}
+            push-to-registry: true
 
   chart-release:
     name: Publish Helm chart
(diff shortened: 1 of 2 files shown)


nojaf:
I'm trying to run the attestation action for each digest.
See https://tomasvotruba.com/blog/2020/11/16/how-to-make-dynamic-matrix-in-github-actions/

nojaf:
Dear @jgiannuzzi, could you skim through this as well please?
Many thanks!

.github/workflows/release.yml
```diff
@@ -60,6 +66,9 @@ jobs:
     if: startsWith(github.ref, 'refs/tags/v') || github.ref == 'refs/heads/main'
     runs-on: ubuntu-latest
     environment: release
+    permissions:
+      id-token: write
+      attestations: write
```
jgiannuzzi, line 71: I don't think these permissions are necessary

.github/workflows/release.yml
```diff
@@ -118,6 +128,23 @@ jobs:
             $(printf -- "--tag ${{ vars.DOCKER_REPO }}:%s " ${tags[@]}) \
             $(printf "${{ vars.DOCKER_REPO }}@%s " ${digests[@]})
           echo "::endgroup::"
+          # Output digests as a JSON array for use in a matrix
+          echo "::set-output name=digests::$(printf '%s\n' "${digests[@]}" | jq -R . | jq -s .)"
+      outputs:
+        digests: ${{ steps.publish-docker-hub.outputs.digests }}
+
+    attest-docker-images:
```
jgiannuzzi, line 136: I think you're missing the `needs` and `permissions` keys for this job

jgiannuzzi (commented):
I left a couple of comment but the general approach LGTM

.github/workflows/release.yml
```diff
@@ -118,6 +125,29 @@ jobs:
...
             $(printf "${{ vars.DOCKER_REPO }}@%s " ${digests[@]})
           echo "::endgroup::"
+          # Output digests as a JSON array for use in a matrix
+          echo "digests=$(printf '%s\n' "${digests[@]}" | jq -R . | jq -s .)" >> $GITHUB_OUTPUT
+    outputs:
+      digests: ${{ steps.publish-docker-hub.outputs.digests }}
+
+  attest-docker-images:
+    name: Publish Docker images attestations
+    if: startsWith(github.ref, 'refs/tags/v') || github.ref == 'refs/heads/main'
```
jgiannuzzi, line 135: This condition is not necessary because the job depends on `docker-release`

.github/workflows/release.yml
```diff
@@ -118,6 +125,29 @@ jobs:
...
+    runs-on: ubuntu-latest
+    needs: docker-release
+    permissions:
+      id-token: write
```
jgiannuzzi, line 139: I think this specific permission isn't necessary

nojaf: But the docs say so: https://github.com/actions/attest?tab=readme-ov-file#usage

jgiannuzzi: My bad!

jgiannuzzi (approved):

pavlovic-ivan (approved):
Looks good to me. Thx @nojaf

PR #1491, opened 2025-01-17
Fix attestation generation

nojaf:


nojaf:
image

nojaf:
And in the dynamic matrix:

image


You might be wondering why there is only a single digest.
Well, because I'm setting the value from within a matrix, it override itself I believe:

```
outputs:
      digests: ${{ steps.publish-docker-hub.outputs.digests }}
```

Anyway, this is good to go and I'm going to revert the experimentation.

PR #1494, opened 2025-01-20
Add multiarch image digest for attestation

pavlovic-ivan:
This PR adds the multiarch image digest to the array of digests, later used for attestations. 

Note: grabbed the first tag (`${tags[0]}`) from the `tags` array because `inspect` will return the same manifest. I can't use multiple args/tags for `inspect` the way they are used in `create`.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -125,6 +125,8 @@ jobs:
             $(printf -- "--tag ${{ vars.DOCKER_REPO }}:%s " ${tags[@]}) \
             $(printf "${{ vars.DOCKER_REPO }}@%s " ${digests[@]})
           echo "::endgroup::"
+          multiarch_digest=$(docker buildx imagetools inspect ${{ vars.DOCKER_REPO }}:${tags[0]} --format '{{json .}}' | jq -r '.manifest.digest')
+          digests+=($multiarch_digest)
           # Output digests as a JSON array for use in a matrix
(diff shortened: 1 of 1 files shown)


nojaf (commented):
This looks good to me from afar. I'll leave it to @jgiannuzzi to decide, as I'm stepping out of my comfort zone here.

suprjinx (approved):
lgtm

jgiannuzzi:
I believe this is vulnerable to a race condition where the manifest gets uploaded during the call to `docker buildx imagetools create`, gets overwritten by a malicious actor, and then gets downloaded during `docker buildx imagetools inspect`, resulting in an attestation for something that we didn't actually build in GitHub Actions.

Here is an idea to solve this (it's probably not the best option available but that's the first thing I'm thinking of):
- use the `--dry-run` option of `docker buildx imagetools create` to create the manifest and store it on disk
- calculate the digest of the stored manifest (with `sha256sum` or `skopeo manifest-digest`)
- upload the manifest to the registry (for each given tag) - we could use `oras manifest push` for that

jgiannuzzi:
Here is another, simpler, idea - though it may be flimsier:
- the reason for the steps above was that `docker buildx imagetools create --dry-run` generates a slightly different output from the manifest that actually gets uploaded - it adds a `\n`
- so we could just get rid of it and compute the digest this way
  ```
  docker buildx imagetools create [all the arguments we normally use...] --dry-run | perl -pe 'chomp if eof' | skopeo manifest-digest /dev/stdin
  ```
- ideally we would then also check that the uploaded manifest's digest _matches_ it, so that any change in the output from the `--dry-run` option would be captured

PR #1499, opened 2025-01-31
Disable PyPI attestations as they prevent publishing to PyPI

jgiannuzzi:
We use reusable workflows, which clash with PyPI attestations, so we have to disable them (we never used them and they got introduced in a newer version of the PyPI publishing action we use).


pavlovic-ivan (approved):
LGTM @jgiannuzzi

